PR #215, opened 2024-05-14
Improve shutdown and connection termination

nickpetrovic:
- Gateway: Use timeout on http shutdown
- Gateway: Use gracefulstop on grpc shutdown
- Gateway: Call cancel func at the end
- Gateway: Add termination grace period
- Proxy: Add http server for readiness probe
- Proxy: Stop accepting connections during shutdown (via readiness prob failure), but wait for active connections to finish
- Proxy: Add init container for local dev
- Proxy: Add readiness probe for local dev
- Proxy: Add termination grace period
- Proxy: Fix redis client name
- Worker: Collect all errors during shutdown and return them
- Worker: Call cancel func after remove worker from repo
- Worker: Remove calling cancel func again
- Worker: Close channels when Run() completes

Resolve BE-1356

### DIFF
--- a/internal/gateway/gateway.go
+++ b/internal/gateway/gateway.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/beam-cloud/beta9/internal/abstractions/endpoint"
 	"github.com/beam-cloud/beta9/internal/task"
-
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 	"google.golang.org/grpc"
@@ -304,10 +303,7 @@ func (g *Gateway) registerServices() error {
 
 // Gateway entry point
 func (g *Gateway) Start() error {
-	listener, err := net.Listen("tcp", fmt.Sprintf(":%d", g.Config.GatewayService.GRPCPort))
-	if err != nil {
-		log.Fatalf("Failed to listen: %v", err)
-	}
+	var err error
 
 	err = g.initGrpc()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


internal/gateway/gateway.go
```diff
@@ -348,5 +353,11 @@ func (g *Gateway) Start() error {
 	<-terminationSignal
 	log.Println("Termination signal received. Shutting down...")
 
+	ctx, cancel := context.WithTimeout(g.ctx, g.Config.GatewayService.ShutdownTimeout)
+	defer cancel()
+	g.httpServer.Shutdown(ctx)
+	g.grpcServer.GracefulStop()
```
jsun-m, line 359: https://github.com/grpc/grpc-go/issues/5930

Isn't this bad?

jsun-m: Should we use .Stop instead?

internal/worker/worker.go
```diff
@@ -198,6 +199,10 @@ func (s *Worker) Run() error {
 
 	go s.manageWorkerCapacity()
 	go s.processStopContainerEvents()
+	defer func() {
+		close(s.completedRequests)
```
jsun-m, line 203: Would these be better added in their respective goroutines?

internal/worker/worker.go
```diff
@@ -779,28 +783,32 @@ func (s *Worker) startup() error {
...
+
 	worker, err := s.workerRepo.GetWorkerById(s.workerId)
 	if err != nil {
-		return err
+		errs = errors.Join(errs, err)
 	}
 
-	s.cancel()
-
 	err = s.workerRepo.RemoveWorker(worker)
```
jsun-m, line 796: In the case `worker, err := s.workerRepo.GetWorkerById(s.workerId)` fails, worker can be `nil`. Since we are not immediately returning the error, there is a likelihood the program will panic at `err = s.workerRepo.RemoveWorker(worker)` in this block
`err := r.lock.Acquire(context.TODO(), common.RedisKeys.SchedulerWorkerLock(workerId), common.RedisLockOptions{TtlS: 10, Retries: 3})`

jsun-m (changes requested):

internal/gateway/gateway.go
nickpetrovic: It looks like this was fixed in https://github.com/grpc/grpc-go/releases/tag/v1.53.0. We're on v1.62.0.

internal/worker/worker.go
nickpetrovic: How would that look?

internal/worker/worker.go
nickpetrovic: Updated. Thoughts?

internal/worker/worker.go
jsun-m: Nevermind, I took another look at these 2 functions. This looks good

internal/worker/worker.go
jsun-m: I think this is good

jsun-m (approved):
The go side lgtm

PR #228, opened 2024-05-17
Add initial tests for most apiv1 endpoints

jsun-m:
Tests Added:
-  containers: list
-  deployment: list, retrieve
-  stub: list
- token: list, create, revoke token
- workspace: create

Moved reused mock components to `pkg/repository/testutils.go`


pkg/api/v1/token_test.go
```diff
@@ -0,0 +1,155 @@
...
+			testDetails.server.ServeHTTP(rec, req)
+			if tt.expectedStatus != 0 {
+				assert.Equal(t, tt.expectedStatus, rec.Code)
+			}
+		})
+	}
+}
+
+func TestRevokeToken(t *testing.T) {
+	testDetails := setupTokenEchoGroup()
```
luke-lombardi, line 94: How exactly is the data being inserted w/ the mocks? Is this actually tested the endpoint functionality itself or just testing the mocks themselves?

pkg/api/v1/container_test.go
```diff
@@ -0,0 +1,65 @@
...
+		server:        e,
+		httpDetails:   &md,
+	}
+}
+
+func TestListContainersByWorkspaceId(t *testing.T) {
+	testDetails := setupContainerEchoGroup()
+
+	// 1. Test with valid workspaceId
+	testDetails.httpDetails.AddExpectedDBTokenSelect(testDetails.httpDetails.Mock, *testDetails.httpDetails.TokenForTest.Workspace, testDetails.httpDetails.TokenForTest)
```
luke-lombardi, line 56: What exactly is happening here? This test is a little hard to read - how is this validating that the containers are only containers within your workspace? Or is it just checking that the token is valid?

If it's just checking that the token is valid, we probably don't need a test case for every set of endpoints just to check that

pkg/api/v1/token_test.go
```diff
@@ -0,0 +1,155 @@
...
+				// Second query is for the workspace check inside the token list method
+				addWorkspaceRowSelectQuery(testDetails.httpDetails)
+				// Third query is for the token list backend
+				testDetails.httpDetails.AddExpectedDBTokenSelect(testDetails.httpDetails.Mock, *testDetails.httpDetails.TokenForTest.Workspace, testDetails.httpDetails.TokenForTest)
+			},
+		},
+		{
+			name: "Test error occurs in db",
+			prepares: func() {
+				testDetails.httpDetails.AddExpectedDBTokenSelect(testDetails.httpDetails.Mock, *testDetails.httpDetails.TokenForTest.Workspace, testDetails.httpDetails.TokenForTest)
```
luke-lombardi, line 68: Same comment as above, I don't really follow the structure of these tests - what exactly are we testing here?

luke-lombardi (changes requested):

pkg/api/v1/container_test.go
nickpetrovic: I think `MockContainerTestDetails` and `setupContainerEchoGroup` are analogous to  `ContainerGroup` and `NewContainerGroup`. Maybe we should name them that with the the suffix `ForTest` like we've done before?

In regards to the body of the function, we might want to turn the majority of this into a helper function. I can see this being used in most endpoints (with the exception endpoints that won't require auth).

jsun-m: This `testDetails.httpDetails.AddExpectedDBTokenSelect` is used to make sure the next call to the db returns the token. Unlike pytest, we don't have fixture data so we have expect db calls and return the exact thing we want through some sort of sqlmock queue.

pkg/api/v1/token_test.go
jsun-m: The mocks inserts the call so that the next db call will return the value that you want to mock. These tests are meant to test the entire authentication, url param parsing, filters and query generation before the db call.

Essentially we test everything in between the api call and the db call and what happens when the db returns specific data

pkg/api/v1/base_test.go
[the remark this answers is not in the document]
jsun-m: Yea. The more annoying part is that we have to tell the mock database that we are expecting each individual token call so that it returns the token and verifies it with the http request token. Maybe I'll look into a test harness that uses sqlite like how pytest does.

pkg/api/v1/container_test.go
jsun-m: > I think `MockContainerTestDetails` and `setupContainerEchoGroup` are analogous to `ContainerGroup` and `NewContainerGroup`. Maybe we should name them that with the the suffix `ForTest` like we've done before?
> 
> In regards to the body of the function, we might want to turn the majority of this into a helper function. I can see this being used in most endpoints (with the exception endpoints that won't require auth).

I'll add the `ForTest` suffix

PR #229, opened 2024-05-17
Fix asyncio and map abstraction

nickpetrovic:
- Use patch to allow nesting asyncio loops
- Remove locks on Redis Map service
- Fix MapKeys response
- Exclude sdk code from worker build to avoid busting build cache


docker/Dockerfile.worker
```diff
@@ -1,4 +1,4 @@
-# syntax=docker/dockerfile:1.6
+# syntax=docker/dockerfile:1.7-labs
```
nickpetrovic, line 1: Allows for the --exclude syntax on COPY

docker/Dockerfile.worker
```diff
@@ -66,7 +66,7 @@ WORKDIR /workspace
 
 COPY go.mod go.sum ./
 RUN go mod download
-COPY . .
+COPY --exclude=sdk . .
```
nickpetrovic, line 69: I don't think this is needed here since the SDK is being copied into `/workspace` [here](https://github.com/beam-cloud/beta9/pull/229/files#diff-d0259383c9168a0ceb522864a6877e2ad094e9661fc4b86821763fe8beffeeceR112).

Doing this allows for quicker changes to the SDK in the worker when go code hasn't been modified.

pkg/abstractions/image/base_requirements.txt
```diff
@@ -14,3 +14,4 @@ psutil==5.9.6
 cloudpickle==3.0.0
 rich==13.7.0
 watchdog==4.0.0
+nest-asyncio==1.6.0
```
luke-lombardi, line 17: Can we remove this dep and embed the code directly into sdk/ somewhere?

sdk/src/beta9/runner/endpoint.py
```diff
@@ -6,6 +6,7 @@
 from http import HTTPStatus
 from typing import Any, Dict, Optional, Tuple
 
+import nest_asyncio
```
luke-lombardi, line 9: see above comment about removing the dep. Also, we should test this fix with functions/taskqueues and see what happens

luke-lombardi (changes requested):

pkg/abstractions/image/base_requirements.txt
nickpetrovic: Yep. Added.

sdk/src/beta9/vendor/nest_asyncio.py
```diff
@@ -0,0 +1,247 @@
+# BSD 2-Clause License
```
nickpetrovic, line 1: Added the license since it seems it may need to be mentioned. 

https://github.com/erdewit/nest_asyncio/blob/master/LICENSE

luke-lombardi (approved):

PR #232, opened 2024-05-20
Add new output abstraction

nickpetrovic:
- Add new output abstraction
- Remove unused redis_map vars
- Nest gateway grpc settings to be consistent with http settings

Resolve BE-1366


sdk/src/beta9/abstractions/output.py
```diff
@@ -0,0 +1,123 @@
...
+        stat = res.stat.to_pydict(casing=Casing.SNAKE)  # type:ignore
+        return Stat(**stat)
+
+    def exists(self) -> bool:
+        try:
+            return True if self.stat() else False
+        except (OutputNotSavedError, OutputNotFoundError):
+            return False
+
+    def public_url(self, *a, **kw) -> str:
```
nickpetrovic, line 87: I'm wondering if we should just call this `public_url` instead of `signed_url` because its not really signed. Its obscured with a uuid and eventually expires.

pkg/abstractions/output/output.proto
```diff
@@ -0,0 +1,56 @@
+syntax = "proto3";
+
+option go_package = "github.com/beam-cloud/beta9/proto";
+import "google/protobuf/timestamp.proto";
+
+package output;
+
+service OutputService {
+  rpc OutputSave(OutputSaveRequest) returns (OutputSaveResponse) {}
```
nickpetrovic, line 9: This should probably be a streaming request.

sdk/src/beta9/abstractions/output.py
mernit: +1 for `public_url`

sdk/src/beta9/__init__.py
```diff
@@ -23,4 +24,5 @@
     "env",
     "GpuType",
     "PythonVersion",
+    "Output",
```
luke-lombardi, line 27: Reminder to update beam-client after we release this. Also we should bump the sdk version in this PR.

pkg/abstractions/output/output.proto
nickpetrovic: Updated.

sdk/src/beta9/__init__.py
nickpetrovic: Bumped the SDK version just now. Will build and publish after this is merged.

luke-lombardi (approved):

pkg/abstractions/output/output.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (o *OutputRedisService) getURL(id string) (string, error) {
+	return o.rdb.Get(context.TODO(), Keys.PublicURL(id)).Result()
+}
+
+// Redis keys
+var (
+	Keys             = &keys{}
+	publicURL string = "output:%s"
```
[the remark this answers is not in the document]
nickpetrovic: Updated.

PR #235, opened 2024-05-21
Concurrency Limits

jsun-m:
```
type ConcurrencyLimit struct {
	Id         uint      `db:"id" json:"-" redis:"-"`
	ExternalId string    `db:"external_id" json:"external_id,omitempty" redis:"external_id"`
	GPULimit   uint32    `db:"gpu_limit" json:"gpu_limit" redis:"gpu_limit"`
	CPULimit   uint32    `db:"cpu_limit" json:"cpu_limit" redis:"cpu_limit"`
	CreatedAt  time.Time `db:"created_at" json:"created_at,omitempty" redis:"-"`
	UpdatedAt  time.Time `db:"updated_at" json:"updated_at,omitempty" redis:"-"`
}
```

Currently only limited by gpu (count) and cpu millicores. Can add additional feature to limit by memory as well. 
@luke-lombardi We talked about limiting just by # of containers up but I'm not completely sure if that is a strong enough restriction. Say if we limit it at 10 containers, users can spin up 10 containers with max memory and max cpu which circumvents this feature.


pkg/repository/base.go
```diff
@@ -40,10 +40,15 @@ type ContainerRepository interface {
 	UpdateContainerStatus(string, types.ContainerStatus, time.Duration) error
 	DeleteContainerState(*types.ContainerRequest) error
 	SetWorkerAddress(containerId string, addr string) error
+	SetContainerStateWithConcurrencyLimit(quota *types.ConcurrencyLimit, request *types.ContainerRequest) error
 	GetWorkerAddress(containerId string) (string, error)
 	GetActiveContainersByStubId(stubId string) ([]types.ContainerState, error)
 	GetActiveContainersByWorkspaceId(workspaceId string) ([]types.ContainerState, error)
 	GetFailedContainerCountByStubId(stubId string) (int, error)
+
+	// Caching concurrency limit TODO: discuss if this should be here
```
jsun-m, line 49: @luke-lombardi This is for the caching of the concurrency limit in redis

pkg/repository/container_redis.go
```diff
@@ -306,3 +311,97 @@ func (cr *ContainerRedisRepository) GetFailedContainerCountByStubId(stubId strin
...
+
+	err = c.rdb.Expire(context.Background(), key, time.Duration(cachedConcurrencyLimitTtl)*time.Second).Err()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *ContainerRedisRepository) SetContainerStateWithConcurrencyLimit(quota *types.ConcurrencyLimit, request *types.ContainerRequest) error {
```
luke-lombardi, line 351: why do we need a separate function for setting container state...?

pkg/repository/container_redis.go
```diff
@@ -306,3 +311,97 @@ func (cr *ContainerRedisRepository) GetFailedContainerCountByStubId(stubId strin
 
 	return failedCount, nil
 }
+
+func (c *ContainerRedisRepository) GetConcurrencyLimitByWorkspaceId(workspaceId string) (*types.ConcurrencyLimit, error) {
```
luke-lombardi, line 315: should we have a workspace repo instead of putting this at the container level?

pkg/scheduler/scheduler.go
```diff
@@ -96,20 +96,42 @@ func (s *Scheduler) Run(request *types.ContainerRequest) error {
...
-		Status:      types.ContainerStatusPending,
-		WorkspaceId: request.WorkspaceId,
-		ScheduledAt: time.Now().Unix(),
-	})
+	quota, err := s.getConcurrencyLimit(request)
+	if err != nil {
+		return err
+	}
+
+	err = s.containerRepo.SetContainerStateWithConcurrencyLimit(quota, request)
```
luke-lombardi, line 104: we shouldn't need a separate state

luke-lombardi (changes requested):

pkg/repository/container_redis.go
jsun-m: I actually have the workspace repo in the Temporary auth token PR

pkg/repository/container_redis.go
jsun-m: It's a select for update essentially.
1. I wanted to acquire the concurrency limit lock
2. I then check the current active containers
3. Then I set the container state
4. I release lock

pkg/scheduler/scheduler.go
jsun-m: Not a separate state, it's the same state

luke-lombardi (approved):

pkg/repository/base.go
luke-lombardi: @jsun-m Should we do another PR after this is merged which moves the SetConcurrencyLimitByWorkspaceId over to a Workspace Repo?

PR #250, opened 2024-05-23
Temporary auth token for file downloads

jsun-m:
Add temporary token for authenticating download requests directly from the browser.

Update to AuthMiddleware:
Allow pass through for auth middle if authorization token is not specified. This means auth middleware mainly handles parsing and acquisition of user detail rather than authenticating.
Authentication is moved to the auth wrappers instead (WithWorkspaceAuth and WithClusterAdminAuth).
This was necessary to expose the `DownloadFileWithToken` to an unauthenticated route

### DIFF
--- a/pkg/abstractions/volume/http.go
+++ b/pkg/abstractions/volume/http.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"net/url"
 
-	"github.com/apex/log"
 	"github.com/beam-cloud/beta9/pkg/auth"
 	"github.com/beam-cloud/beta9/pkg/repository"
 	"github.com/labstack/echo/v4"
@@ -91,7 +90,6 @@ func (g *volumeGroup) UploadFile(ctx echo.Context) error {
 
 	go func() {
 		defer close(ch)
-		bytesRead := 0
 
 		for {
 			buf := make([]byte, uploadBufferSize) // 8 Mb
@@ -100,16 +98,11 @@ func (g *volumeGroup) UploadFile(ctx echo.Context) error {
 				break
 			}
 
-			log.Infof("Read %d bytes", n)
-			bytesRead += n
-
 			ch <- CopyPathContent{
 				Path:    decodedVolumePath,
 				Content: buf[:n],
 			}
 		}
-
-		log.Infof("Read %d bytes in total", bytesRead)
 	}()
 
 	if err := g.gvs.copyPathStream(
--- a/pkg/abstractions/volume/http.go
+++ b/pkg/abstractions/volume/http.go
@@ -19,11 +19,10 @@ type volumeGroup struct {
 
 var uploadBufferSize = 1024 * 1000 * 8 // 8 Mb
 
-func registerVolumeRoutes(g *echo.Group, gvs *GlobalVolumeService, workspaceRepo repository.WorkspaceRepository) *volumeGroup {
+func registerVolumeRoutes(g *echo.Group, gvs *GlobalVolumeService) *volumeGroup {
 	group := &volumeGroup{
-		routeGroup:    g,
-		gvs:           gvs,
-		workspaceRepo: workspaceRepo,
+		routeGroup: g,
+		gvs:        gvs,
 	}
 
 	g.GET("/:workspaceId", group.ListVolumes)
@@ -134,7 +133,7 @@ func (g *volumeGroup) DownloadFileWithToken(ctx echo.Context) error {
 		return echo.NewHTTPError(http.StatusBadRequest, "Invalid download token")
 	}
 
-	if err := g.workspaceRepo.ValidateWorkspaceVolumePathDownloadToken(workspaceId, decodedVolumePath, token); err != nil {
+	if err := g.gvs.ValidateWorkspaceVolumePathDownloadToken(workspaceId, decodedVolumePath, token); err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "Invalid download token")
 	}
 
@@ -233,7 +232,7 @@ func (g *volumeGroup) GenerateDownloadToken(ctx echo.Context) error {
 		return echo.NewHTTPError(http.StatusBadRequest, "Invalid volume path")
 	}
 
-	token, err := g.workspaceRepo.GenerateWorkspaceVolumePathDownloadToken(workspaceId, decodedVolumePath)
+	token, err := g.gvs.GenerateWorkspaceVolumePathDownloadToken(workspaceId, decodedVolumePath)
 	if err != nil {
 		return echo.NewHTTPError(http.StatusInternalServerError, "Failed to generate download token")
 	}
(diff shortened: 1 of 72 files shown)


pkg/abstractions/volume/http.go
```diff
@@ -121,14 +162,14 @@ func (g *volumeGroup) DownloadFile(ctx echo.Context) error {
...
-	if f, err := g.gvs.getFileFd(
+	if path, err := g.gvs.getFilePath(
 		ctx.Request().Context(),
 		decodedVolumePath,
 		&workspace,
 	); err != nil {
 		return echo.NewHTTPError(http.StatusInternalServerError, fmt.Sprintf("Failed to download file %v", err))
 	} else {
-		return ctx.Stream(http.StatusOK, "application/octet-stream", f)
+		return ctx.File(path)
```
nickpetrovic, line 172: I think this loads the entire file into memory which we don't really want. Can you double check? 

Also curious why Stream() doesn't work here.

pkg/gateway/services/auth.go
```diff
@@ -2,13 +2,16 @@ package gatewayservices
...
 	"context"
+	"log"
 
 	"github.com/beam-cloud/beta9/pkg/auth"
 	"github.com/beam-cloud/beta9/pkg/types"
 	pb "github.com/beam-cloud/beta9/proto"
 )
 
 func (gws *GatewayService) Authorize(ctx context.Context, in *pb.AuthorizeRequest) (*pb.AuthorizeResponse, error) {
+	log.Printf("Authorize request: %v", in)
```
nickpetrovic, line 13: Can we remove this?

pkg/repository/workspace.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	if data.WorkspaceId != workspaceId || data.VolumePath != volumePath {
+		return errors.New("invalid token")
+	}
+
+	return nil
+}
+
+func generateToken(length int) (string, error) {
```
nickpetrovic, line 51: UUID might be simpler. Its random (uuid4) without having to deal with returning an error.

pkg/repository/workspace.go
nickpetrovic, line 1: For abstractions, we've been putting calls to redis code in the abstraction package. Any reason we're not doing that here? What else would could be placed in this repo?

pkg/abstractions/volume/http.go
jsun-m: Stream doesnt provide file related information like content-type or content-length. The browser won't know how much it needs to download. 
I also think the File response should be optimized and use a reader. I can verify it though

pkg/gateway/services/auth.go
jsun-m: removed

pkg/abstractions/volume/http.go
nickpetrovic: Right, but you can perform a best-effort detection. [This](https://github.com/beam-cloud/beta9/blob/main/pkg/abstractions/output/http.go#L39-L51) might help.

pkg/repository/workspace.go
jsun-m: I was thinking of adding concurrency limits to this repo as well. You're right we have been adding redis requests directly into the abstractions.

Thought it would be cleaner if I did it this way. Let me give it more thought

pkg/abstractions/volume/http.go
jsun-m: I just tested this with a 5.25 gb file. Memory of the pod didnt change. In fact, execing into the pod shows that its not using significant memory

nickpetrovic: I looked through the call stack 

`File() -> FileFS() -> fsFile() -> http.ServeContent() -> serveContent()`

And it doesn't look like it loads the entire file into memory. It does seek through the file to get the size of it. I'm not sure how juicefs processes this e.g. does it download the entire file to Seek or does it just get and use the metadata attributes from juicefs' redis.

nickpetrovic: For what its worth, I asked ChatGPT and it says juicefs just uses metadata in redis when you os.Seek on a file, but I haven't checked juicefs myself.

Also - are you sure Stream doesn't work through the browser? I just downloaded an output (a text file) via the browser just fine. Are you seeing a specific issue with this? I can see maybe content type may not work since in outputs we're not using `mime.TypeByExtension()`.

jsun-m: I think the biggest issue is that the browser won't know the content-length so it won't be able to get time estimates for the download as well as the progress
image
image

vs using this
image
image

nickpetrovic: Hm I see. It still works, but from a UX perspective it's not the best.

Trust in ChatGPT? I'll leave this up to you to investigate or not.

jsun-m: For the metadata, I was under the assumption that the redis stores the os.stat information

nickpetrovic: The underlying function performs an os.Seek, not os.Stat. This is what I'm unsure of. ChatGPT says os.Seek still uses metadata.

pkg/repository/workspace.go
nickpetrovic: Either is fine, just suggesting a simpler way to generate a random token. Feel free to keep this as-is.

pkg/repository/workspace.go
jsun-m: I moved the concurrency limit cache sets/gets into the workspace repo. I believe we might find it useful to have a workspace encompassed set of redis calls for the future

jsun-m:
```
type WorkspaceRepository interface {
	ValidateWorkspaceVolumePathDownloadToken(workspaceId string, volumePath string, token string) error
	GenerateWorkspaceVolumePathDownloadToken(workspaceId string, volumePath string) (string, error)
	GetConcurrencyLimitByWorkspaceId(workspaceId string) (*types.ConcurrencyLimit, error)
	SetConcurrencyLimitByWorkspaceId(workspaceId string, limit *types.ConcurrencyLimit) error
}
```

nickpetrovic: I think concurrency makes sense to be scoped to the workspace or at the workspace level. Could be around the concept of resource management which could include other things like nodes and/or providers. I don't know about volume token though since volume is an abstraction.

For example, I can see volume permissions or maybe even limits for volumes defined at the workspace level. These seem like things you'd want to manage at that level vs at the abstraction level - if you're an admin or something. To me, generating a token is just a function/feature of the abstraction. At least that's where I lean.

luke-lombardi: +1 I think the concurrency stuff should probably stay in workspace repo but volumes token validation should be inside abstractions/volume/

jsun-m: ok fair fair

jsun-m: done

nickpetrovic (approved):

PR #254, opened 2024-05-28
De-asyncio SDK abstractions

nickpetrovic:
Resolve BE-1404

### DIFF
--- a/sdk/src/beta9/channel.py
+++ b/sdk/src/beta9/channel.py
@@ -2,16 +2,15 @@ import functools
 import os
 import sys
 import traceback
+from abc import ABC, abstractmethod
 from contextlib import contextmanager
-from typing import Any, Callable, Optional, Type, cast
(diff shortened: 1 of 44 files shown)


sdk/src/beta9/channel.py
```diff
@@ -26,48 +25,91 @@
 from .exceptions import RunnerException
 
 
-class AuthenticatedChannel(Channel):
-    def __init__(self, *args, token: Optional[str] = None, **kwargs):
-        super().__init__(*args, **kwargs)
+class Channel(InterceptorChannel):
```
luke-lombardi, line 28: Assuming you tested this in the runner as well? Looks good.

luke-lombardi (approved):

sdk/src/beta9/channel.py
nickpetrovic: Yep. Ran the code in runner locally.

PR #255, opened 2024-05-28
Add machine management commands for external machine management

luke-lombardi:
- Add machine create / list commands
- Add pool list command
- Update tokens / grpc endpoints to support manual machine creation


pkg/api/v1/machine.go
```diff
@@ -42,7 +42,10 @@ type RegisterMachineRequest struct {
 }
 
 func (g *MachineGroup) RegisterMachine(ctx echo.Context) error {
-	_, _ = ctx.(*auth.HttpAuthContext)
+	cc, _ := ctx.(*auth.HttpAuthContext)
+	if (cc.AuthInfo.Token.TokenType != types.TokenTypeMachine) && (cc.AuthInfo.Token.TokenType != types.TokenTypeWorker) {
```
jsun-m, line 46: Try replacing this with WithClusterAdmiAuth

luke-lombardi: I can't here because I'm checking against a different token type

mernit (approved):

PR #268, opened 2024-05-29
Hotfix: Don't override config with remote values

luke-lombardi:


pkg/api/v1/machine.go
```diff
@@ -52,10 +53,18 @@ func (g *MachineGroup) RegisterMachine(ctx echo.Context) error {
 		return echo.NewHTTPError(http.StatusBadRequest, "Invalid payload")
 	}
 
+	configByes, err := json.Marshal(g.config)
```
jsun-m, line 56: typo. ConfigBytes

luke-lombardi: thanks, fixed

jsun-m (approved):

PR #276, opened 2024-05-31
Add protobuf codegen verification

jsun-m:


bin/verify_proto.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
jsun-m, line 1: https://github.com/etcd-io/etcd/blob/main/scripts/verify_genproto.sh

luke-lombardi (approved):

PR #280, opened 2024-06-03
Workspace Secrets

jsun-m:
### Purpose
This feature allows users to create and manage secret variables that will be passed into their container runtimes' environment variables. 

### SDK
The secrets are stored globally by the workspace and can be attached to a stub by adding it to the stub definitions.
```
@function(secrets=["test_secret"])
```

### CLI Commands
- list
- create
- modify
- delete
- show


pkg/abstractions/endpoint/instance.go
```diff
@@ -31,21 +31,30 @@ type endpointInstance struct {
...
+		fmt.Sprintf("TIMEOUT=%d", i.StubConfig.TaskPolicy.Timeout),
+	}
+
+	env = append(env, secrets...)
+
 	for c := 0; c < containersToRun; c++ {
 		containerId := i.genContainerId()
 		runRequest := &types.ContainerRequest{
 			ContainerId: containerId,
-			Env: []string{
```
luke-lombardi, line 38: We may want to append the secrets prior to our internal env vars to prevent a user from overriding something critical by mistake

pkg/abstractions/function/task.go
```diff
@@ -124,15 +132,19 @@ func (t *FunctionTask) run(ctx context.Context, stub *types.StubWithRelated) err
...
 
+	env := []string{
+		fmt.Sprintf("TASK_ID=%s", t.msg.TaskId),
+		fmt.Sprintf("HANDLER=%s", stubConfig.Handler),
+		fmt.Sprintf("BETA9_TOKEN=%s", token.Key),
+		fmt.Sprintf("STUB_ID=%s", stub.ExternalId),
+		fmt.Sprintf("CALLBACK_URL=%s", stubConfig.CallbackUrl),
+	}
+
+	env = append(env, secrets...)
```
luke-lombardi, line 143: same comment as above

pkg/abstractions/secret/http.go
```diff
@@ -0,0 +1,136 @@
...
+	ss         *WorkspaceSecretService
+}
+
+func registerSecretRoutes(g *echo.Group, ss *WorkspaceSecretService) *secretGroup {
+	group := &secretGroup{
+		routeGroup: g,
+		ss:         ss,
+	}
+
+	g.POST("/:workspaceId", auth.WithWorkspaceAuth(group.CreateSecret))
```
luke-lombardi, line 23: any reason we're prefixing by workspace ID in the route here? can't we just extract it from the auth token?

pkg/abstractions/secret/http.go
```diff
@@ -0,0 +1,136 @@
...
+	secretInput := types.Secret{}
+	if err := ctx.Bind(&secretInput); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "Invalid secret")
+	}
+
+	// Create the secret
+	secret, err := g.ss.backendRepo.CreateSecret(ctx.Request().Context(), &workspace, cc.AuthInfo.Token.Id, secretInput.Name, secretInput.Value)
+	if err != nil {
+		if err, ok := err.(*pq.Error); ok {
+			if err.Code == "23505" { // Unique violation
```
luke-lombardi, line 53: What is this error code...?

pkg/abstractions/secret/http.go
```diff
@@ -0,0 +1,136 @@
...
+	workspaceId := ctx.Param("workspaceId")
+	workspace, err := g.ss.backendRepo.GetWorkspaceByExternalIdWithSigningKey(ctx.Request().Context(), workspaceId)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "Invalid workspace ID")
+	}
+
+	secretName := ctx.Param("secretName")
+	secret, err := g.ss.backendRepo.GetSecretByNameDecrypted(ctx.Request().Context(), &workspace, secretName)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusNotFound, "Secret not found")
```
luke-lombardi, line 74: we should try to be consistent around the formatting of errors from http endpoints. Are we using lower case or human readable for these elsewhere?

pkg/abstractions/secret/secret.go
```diff
@@ -0,0 +1,164 @@
...
+	registerSecretRoutes(routeGroup.Group(secretRoutePrefix, authMiddleware), ss)
+
+	return ss
+}
+
+func handleErrMsg(err error) string {
+	msg := err.Error()
+
+	if err, ok := err.(*pq.Error); ok {
+		if err.Code == "23505" { // Unique violation
```
luke-lombardi, line 47: need to clarify this a bit, reading this I would have no clue what this magic number is / why it's needed

pkg/abstractions/secret/secret.proto
```diff
@@ -0,0 +1,77 @@
...
+  rpc UpdateSecret(UpdateSecretRequest) returns (UpdateSecretResponse) {}
+  rpc GetSecret(GetSecretRequest) returns (GetSecretResponse) {}
+  rpc ListSecrets(ListSecretsRequest) returns (ListSecretsResponse) {}
+}
+
+message Secret {
+  string id = 1;
+  string name = 2;
+  string value = 3;
+  bool sensitive = 4;
```
luke-lombardi, line 20: remove sensitive field

pkg/abstractions/secret/secret.proto
```diff
@@ -0,0 +1,77 @@
...
+  string last_updated_by = 5;
+  google.protobuf.Timestamp created_at = 6;
+  google.protobuf.Timestamp updated_at = 7;
+}
+
+message CreateSecretRequest {
+  string name = 2;
+  string value = 3;
+  bool sensitive = 4;
```
luke-lombardi, line 29: remove sensitive

pkg/abstractions/taskqueue/instance.go
```diff
@@ -31,20 +31,29 @@ type taskQueueInstance struct {
...
+		fmt.Sprintf("ON_START=%s", i.StubConfig.OnStart),
+		fmt.Sprintf("CALLBACK_URL=%s", i.StubConfig.CallbackUrl),
+		fmt.Sprintf("STUB_ID=%s", i.Stub.ExternalId),
+		fmt.Sprintf("STUB_TYPE=%s", i.Stub.Type),
+		fmt.Sprintf("CONCURRENCY=%d", i.StubConfig.Concurrency),
+		fmt.Sprintf("KEEP_WARM_SECONDS=%d", i.StubConfig.KeepWarmSeconds),
+		fmt.Sprintf("PYTHON_VERSION=%s", i.StubConfig.PythonVersion),
+	}
+
+	env = append(env, secrets...)
```
luke-lombardi, line 51: same comment about moving secrets before env

luke-lombardi (changes requested):

pkg/abstractions/secret/http.go
jsun-m: I'm just thinking we do this so that cluster admins can have access to these endpoints. Maybe it's not a good idea to do it for secrets?

luke-lombardi: Yeah I feel like we should remove the :workspaceId prefix and come up with a different way of "assuming auth" in the cluster admin case

pkg/abstractions/secret/http.go
jsun-m: fixed

pkg/task/dispatch.go
```diff
@@ -47,6 +48,7 @@ func (d *Dispatcher) getTaskMessage() *types.TaskMessage {
 	msg.TaskId = uuid.Must(uuid.NewV4()).String()
 	msg.StubId = ""
 	msg.WorkspaceName = ""
+	msg.WorkspaceId = ""
```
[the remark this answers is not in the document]
jsun-m: removed. This was while i was testing different ways to do things

pkg/abstractions/secret/http.go
jsun-m: would be difficult in future for this particular endpoint. Secret `name` is unique per workspace so workspace would need to be specified in order to access a variable from the cluster admin side. 
A lot of other endpoints use the `/:workspaceId` param

luke-lombardi (approved):

PR #285, opened 2024-06-05
Fix issues with lambda labs machines

luke-lombardi:
- Add optional skip tls verification flag (disabled by default)
- Encapsulate remote config creation into a helper function
- Clean up pool_external and ensure gpu count is set properly on remote machines
- Pass remote config as JSON for simplicity
- Remove secret mount stuff as it's no longer needed if passing config as JSON into an env var
- Add total cpu / total mem / total gpu count fields to track workers maximum capacity


pkg/scheduler/pool_external.go
```diff
@@ -119,15 +120,15 @@ func (wpc *ExternalWorkerPoolController) AddWorker(cpu int64, memory int64, gpuT
 			}
 
 			for _, worker := range workers {
-				machine.State.Cpu -= worker.Cpu
-				machine.State.Memory -= worker.Memory
-				machine.State.GpuCount -= worker.GpuCount
+				machine.State.Cpu -= worker.TotalCpu
+				machine.State.Memory -= worker.TotalMemory
+				machine.State.GpuCount -= uint32(worker.TotalGpuCount)
```
jsun-m, line 125: Do we need this log?

pkg/scheduler/pool_external.go
```diff
@@ -199,10 +204,12 @@ func (wpc *ExternalWorkerPoolController) createWorkerOnMachine(workerId, machine
 		return nil, err
 	}
 
+	log.Printf("Created worked: %+v\n", worker)
+
 	return worker, nil
 }
```
jsun-m, line 210: Remove log

luke-lombardi: Done

pkg/scheduler/pool_external.go
luke-lombardi: Removed

jsun-m (approved):

PR #286, opened 2024-06-06
Add workspace restricting and cleaning functions

jsun-m:
Add http endpoints to 
1. stop deployments
   - Disables deployments and prevents requests from reaching them
   - stops all active containers
2. stop containers
   - stop all containers (catch all for containers not related to a deployment)
3. disable token
   - cluster admin can root-disable a token (users can user-disable their own tokens already)
4.  delete concurrency limits


pkg/auth/grpc.go
```diff
@@ -55,6 +55,10 @@ func (ai *AuthInterceptor) validateToken(md metadata.MD) (*AuthInfo, bool) {
 		return nil, false
 	}
 
+	if !token.Active || token.DisabledByClusterAdmin {
```
jsun-m, line 58: changes to auth to only allow active tokens

pkg/api/v1/container.go
```diff
@@ -42,3 +47,26 @@ func (c *ContainerGroup) ListContainersByWorkspaceId(ctx echo.Context) error {
...
+	workspaceId := ctx.Param("workspaceId")
+	containerStates, err := c.containerRepo.GetActiveContainersByWorkspaceId(workspaceId)
+	if err != nil {
+		return ctx.JSON(http.StatusInternalServerError, map[string]interface{}{
+			"message": "failed to stop containers",
+		})
+	}
+
+	for _, state := range containerStates {
+		err := c.scheduler.Stop(state.ContainerId)
```
luke-lombardi, line 61: should we break on this error? or just continue to the rest of the containers

pkg/api/v1/token.go
```diff
@@ -32,25 +32,58 @@ func (g *TokenGroup) CreateWorkspaceToken(ctx echo.Context) error {
...
 
-type RevokeTokenRequest struct {
-	Token string `json:"token"`
+type ClusterAdminTokensRequestSerializer struct {
+	Disabled bool `json:"enabled"`
 }
 
-func (g *TokenGroup) RevokeWorkspaceToken(ctx echo.Context) error {
-	_, _ = ctx.(*auth.HttpAuthContext)
+func (g *TokenGroup) ClusterAdminUpdateAllWorkspaceTokens(ctx echo.Context) error {
```
luke-lombardi, line 56: this should probably also be renamed - it doesn't just update tokens, it only disables them

luke-lombardi (changes requested):

pkg/api/v1/token.go
jsun-m: It toggles them

pkg/api/v1/container.go
jsun-m: Might be better to continue yea. I'll just log the error message

luke-lombardi (approved):

PR #295, opened 2024-06-13
Add CLI release workflow

nickpetrovic:
- Build standalone Python CLI binary with PyApp
- Upload binary to Action Workflow
- Upload binary to GitHub Release when Workflow is triggered by a Release

Resolve BE-1387

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -22,7 +22,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [macos-latest, ubuntu-latest]
         arch: [x86_64, arm64]
 
     steps:
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -1,0 +1,80 @@
+name: Build and release CLI
+
+on:
+  release:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-cli.yml
```diff
@@ -0,0 +1,112 @@
...
+          path: ${{ steps.build-binary.outputs.bin_file_path }}
+          if-no-files-found: ignore
+          compression-level: "9"
+
+      - name: Upload binary to release
+        if: inputs.app_version == ''
+        run: |
+          gh release upload ${GITHUB_REF_NAME} "${{ steps.build-binary.outputs.tar_file_path }}"
+
+      # - name: Upload binary
```
luke-lombardi, line 100: are we planning on handling this in a separate PR?

luke-lombardi (changes requested):

.github/workflows/release-cli.yml
nickpetrovic: Yeah, I need to clean this part up by removing the commented out yaml and adding a new job that makes an API call to the homebrew repo.

nickpetrovic: Updated.

PR #301, opened 2024-06-17
Fix help and version output in CLI

nickpetrovic:
When using pyapp to build a binary, the --help and --version options output `Usage: python -m beam [OPTIONS] COMMAND [ARGS]... ` and `python -m beam, version 0.2.23` respectively. This fixes that so the `python -m beam` portion is just `beam`.

Resolve BE-1387

### DIFF
--- a/sdk/pyproject.toml
+++ b/sdk/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "beta9"
-version = "0.1.35"
+version = "0.1.37"
 description = ""
 authors = ["beam.cloud <support@beam.cloud>"]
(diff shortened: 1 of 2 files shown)


luke-lombardi (approved):

sdk/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "beta9"
-version = "0.1.35"
+version = "0.1.37"
```
nickpetrovic, line 3: 0.1.36 already exists. Just wasn't committed which is why I'm going straight to 0.1.37.

PR #305, opened 2024-06-18
Add types for H100 and A6000

jsun-m:


luke-lombardi (changes requested):
I think you need to add a version bump to sdk here

luke-lombardi (approved):

PR #312, opened 2024-06-20
Improve hot reloading with serves

nickpetrovic:
- Treat files and directories differently. Files were being created with MkdirAll, then written to, but failed because its not a file
- Add move operation to support moving files, directories, and sub directories
- Remove <Folder/File> <event type> logs
- Update detected changes log to include code file name and highlighted in yellow

Resolve BE-1434


nickpetrovic:
Preview of what the changes look like locally.

image

mernit:
Looks great.

PR #318, opened 2024-06-21
Add directory and PIL.Image support to Outputs

nickpetrovic:
Resolves BE-1474 BE-1482

### DIFF
--- a/sdk/src/beta9/abstractions/output.py
+++ b/sdk/src/beta9/abstractions/output.py
@@ -1,7 +1,10 @@
 import os
+import shutil
+import uuid
+import zipfile
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Dict, Generator, NamedTuple, Optional, Union
+from typing import Any, Dict, Generator, NamedTuple, Optional, Protocol, Union
 
 from betterproto import Casing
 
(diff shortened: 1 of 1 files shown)


sdk/src/beta9/abstractions/output.py
```diff
@@ -21,13 +24,58 @@
...
+        When the output is a directory, it is zipped up before being saved. When it
+        is a file, it's saved as-as (no zipping).
+
+        Args:
+            path: The path to a file or directory.
+
+        Raises:
+            OutputCannotRunLocallyError: _description_
+            OutputTaskIdError: _description_
+            FileNotFoundError: _description_
```
nickpetrovic, line 48: Forgot to update!

sdk/src/beta9/abstractions/output.py
```diff
@@ -21,13 +24,62 @@
...
+            output = Output(path=myfile)
+            output.save()
+            output_url = output.public_url()
+            ```
+
+            Saving a PIL.Image object.
+
+            ```python
+            image = pipe( ... )
+            output = Output.from_pil_image(image)
```
mernit, line 77: Lovely

sdk/src/beta9/abstractions/output.py
```diff
@@ -43,29 +95,101 @@ def __init__(self, *, path: Union[Path, str]) -> None:
         self.id: Optional[str] = None
         self._stub: Optional[OutputServiceStub] = None
 
+    def __del__(self) -> None:
```
luke-lombardi, line 98: Does this get called between task invocations? (like after handler exits and Output obj goes out of scope?)

luke-lombardi (approved):
lgtm, just had a question

sdk/src/beta9/abstractions/output.py
```diff
@@ -21,13 +24,62 @@
 
 class Output(BaseAbstraction):
     """
-    A file that a task has created.
-
-    Use this to save a file you may want to save and share later.
+    A file or directory to persist after a task has ended.
     """
 
+    _tmp_dir = Path("/tmp/outputs")
```
nickpetrovic, line 30: `_tmp_dir` is a class variable used as a base path for temp storage of all `Output`s. 

In this PR, we build temp paths with the `id` of the instance so that when the Output instance is garbage collected, we only delete that instance's temp data. This includes zipped directories and PIL.Image files.

sdk/src/beta9/abstractions/output.py
nickpetrovic: Yes it appears so.

nickpetrovic: Tested with endpoint and task_queue. `__del__` is called after every task invocation.

PR #320, opened 2024-06-22
Fix upload progress when building images

nickpetrovic:
Resolve BE-1453


nickpetrovic:
Preview of what this looks like.



https://github.com/beam-cloud/beta9/assets/4001122/0da9d575-fc89-4a5c-9c18-0cf317dae2ee

PR #330, opened 2024-06-26
Add prometheus (victoria metrics) agent url

luke-lombardi:


pkg/types/config.go
```diff
@@ -284,6 +284,7 @@ type MonitoringConfig struct {
 }
 
 type PrometheusConfig struct {
+	AgentUrl      bool `key:"agentUrl" json:"agent_url"`
```
luke-lombardi, line 287: Is this the right name / location for this?

pkg/types/config.go
```diff
@@ -284,8 +284,11 @@ type MonitoringConfig struct {
 }
 
 type PrometheusConfig struct {
-	ScrapeWorkers bool `key:"scrapeWorkers" json:"scrape_workers"`
-	Port          int  `key:"port" json:"port"`
+	AgentUrl      string `key:"agentUrl" json:"agent_url"`
```
luke-lombardi, line 287: Are these the right names / location for these?

nickpetrovic (approved):

PR #331, opened 2024-06-26
Add timeouts to endpoint and taskqueue serves

jsun-m:
Serves will timeout after 10 minutes of activity.
File changes will refresh this timer.

Users can set their own timeout through the cli command
`beta9 serve file.py:func --timeout 600`
Timeout value of 0 will use the default timeout
Timeout value of -1 will remove the timeout and persist the container for as long as the user is connected


sdk/src/beta9/abstractions/taskqueue.py
```diff
@@ -230,26 +231,35 @@ def _handle_serve_interrupt(self) -> None:
         os._exit(0)  # kills all threads immediately
 
     def _serve(self, *, dir: str, object_id: str):
-        with ThreadPoolExecutor() as pool:
```
jsun-m, line 233: I replaced the pool executor because the thread for `self.parent.sync_dir_to_workspace` never shut down and blocked termination of the application.

sdk/src/beta9/abstractions/taskqueue.py
```diff
@@ -230,26 +231,35 @@ def _handle_serve_interrupt(self) -> None:
...
+            self.parent.taskqueue_stub.task_queue_serve_keep_alive(
+                TaskQueueServeKeepAliveRequest(
+                    stub_id=self.parent.stub_id,
+                )
+            )
+
+        threading.Thread(
+            target=self.parent.sync_dir_to_workspace,
+            kwargs={"dir": dir, "object_id": object_id, "on_event": notify},
+            daemon=True,
```
jsun-m, line 244: This field is required so that the thread automatically shuts down after rest of the process completes

pkg/abstractions/endpoint/serve.go
```diff
@@ -29,19 +29,30 @@ func (es *HttpEndpointService) StartEndpointServe(in *pb.StartEndpointServeReque
...
+		timeoutDuration,
 	)
 
 	container, err := instance.WaitForContainer(ctx, endpointServeContainerTimeout)
 	if err != nil {
 		return err
 	}
 
+	// Remove the container lock and rely on the serve lock to keep container alive
+	instance.Rdb.Del(
```
luke-lombardi, line 51: is this safe?

luke-lombardi (approved):

pkg/abstractions/endpoint/serve.go
jsun-m: yes

PR #333, opened 2024-06-27
Update events logic to allow mapping specific events to specific tags

jsun-m:
HTTP Inputs in Fluent-bit dynamically set `tags` based off the URI of the request.
e.g. a request to `fluentbit.monitoring:9880/tag_name` results in fluent-bit automatically setting a tag of `tag_name` to the event

These tag names are then matched with an output in the example below
```
[OUTPUT]
          Name  http
          Match tag_name
          ....
```

By default, the events will map to the tag `http.1` unless it's override in the input. However, I added a mapping that can be set in `config.yaml` to map events to specific tags by event name as such:
```
monitoring:
  fluentbit:
    events:
      mapping:
      - name: abstraction.trigger
         tag: internal_api
```

### DIFF
--- a/pkg/repository/base.go
+++ b/pkg/repository/base.go
@@ -147,7 +147,9 @@ type EventRepository interface {
 	PushContainerStoppedEvent(containerID string, workerID string)
 	PushWorkerStartedEvent(workerID string)
 	PushWorkerStoppedEvent(workerID string)
-	PushAbstractionTriggeredEvent(workspaceId string, stub *types.Stub)
+	PushDeployStubEvent(workspaceId string, stub *types.Stub)
+	PushServeStubEvent(workspaceId string, stub *types.Stub)
+	PushRunStubEvent(workspaceId string, stub *types.Stub)
 }
 
 type MetricsRepository interface {
--- a/pkg/repository/base.go
+++ b/pkg/repository/base.go
@@ -147,6 +147,7 @@ type EventRepository interface {
 	PushContainerStoppedEvent(containerID string, workerID string)
 	PushWorkerStartedEvent(workerID string)
 	PushWorkerStoppedEvent(workerID string)
+	PushAbstractionTriggeredEvent(workspaceId string, stub *types.Stub)
 }
 
 type MetricsRepository interface {
--- a/pkg/repository/events.go
+++ b/pkg/repository/events.go
@@ -174,11 +174,37 @@ func (t *TCPEventClientRepo) PushWorkerStoppedEvent(workerID string) {
 	)
 }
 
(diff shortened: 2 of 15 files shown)


pkg/types/event.go
```diff
@@ -50,3 +51,12 @@ type EventWorkerLifecycleSchema struct {
 	WorkerID string `json:"worker_id"`
 	Status   string `json:"status"`
 }
+
+var EventAbstractionTriggerSchemaVersion = "1.0"
+
+type EventAbstractionTriggeredSchema struct {
```
luke-lombardi, line 57: Shouldn't this have an event_type field or something? this feels ambiguous

pkg/repository/base.go
```diff
@@ -147,6 +147,7 @@ type EventRepository interface {
 	PushContainerStoppedEvent(containerID string, workerID string)
 	PushWorkerStartedEvent(workerID string)
 	PushWorkerStoppedEvent(workerID string)
+	PushAbstractionTriggeredEvent(workspaceId string, stub *types.Stub)
```
luke-lombardi, line 150: I don't really like this name, it feels confusing. I'd almost argue for more specific functions even at the risk of more code / being more verbose. I./e. `PushDeployedStubEvent()` `PushServedStubEvent()`

pkg/repository/events.go
```diff
@@ -25,9 +26,16 @@ func NewTCPEventClientRepo(config types.FluentBitEventConfig) EventRepository {
 		log.Println("[WARNING] fluentbit host does not appear to be up, events will be dropped")
 	}
 
+	// Parse event mapping
+	eventTagMap := make(map[string]string)
```
luke-lombardi, line 30: you might not need the mapping if you use specific event types right?

luke-lombardi (changes requested):

pkg/types/event.go
jsun-m: The stubtype includes the abstraction and event type

pkg/repository/events.go
jsun-m: The mapping is generic to easily map any event to different outputs

jsun-m: That means users will be able to better direct where they want events to go. E.g. it goes to an API, elastic search, kinesis, or straight to slack through fluentbit

pkg/repository/base.go
jsun-m: I like this better. I renamed the subject from abstraction -> stub

luke-lombardi (approved):
Looks good. I'm still a little iffy on the event tag mapping but let's try it out.

PR #335, opened 2024-06-28
Update agent download urls

luke-lombardi:


nickpetrovic (approved):
lgtm. is this just until use the manifests idea?

luke-lombardi:
> lgtm. is this just until use the manifests idea?

No I'm always updating the https://release.beam.cloud/agent/agent with the latest version, so when you run machine create, you're always by default using the latest version. The manifest thing is only used by the auto-update process.

PR #343, opened 2024-07-01
Add validations to Runner

nickpetrovic:
- Add GPU validation before runner/stub preparation
- Add better type annotations for GpuType and GpuTypeLiteral (strings)
- Add printing of errors when `prepare_runtime` fails on certain validations
- Fix serving a function (ensure not implemented error is printed)

Resolve BE-1500

### DIFF
--- a/sdk/src/beta9/abstractions/base/runner.py
+++ b/sdk/src/beta9/abstractions/base/runner.py
@@ -23,7 +23,7 @@ from ...clients.gateway import (
 from ...config import ConfigContext, SDKSettings, get_config_context, get_settings
 from ...env import called_on_import
 from ...sync import FileSyncer, SyncEventHandler
-from ...type import _AUTOSCALERS, Autoscaler, QueueDepthAutoscaler
+from ...type import _AUTOSCALERS, Autoscaler, GpuType, QueueDepthAutoscaler
 
 CONTAINER_STUB_TYPE = "container"
 FUNCTION_STUB_TYPE = "function"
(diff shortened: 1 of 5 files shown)


sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -267,6 +267,7 @@ def prepare_runtime(
                 self.image_available = True
                 self.image_id = image_build_result.image_id
             else:
+                terminal.error("Image build failed", exit=False)
```
luke-lombardi, line 270: Don't we want to exit in these failure cases?

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -276,14 +277,26 @@ def prepare_runtime(
                 self.files_synced = True
                 self.object_id = sync_result.object_id
             else:
+                terminal.error("File sync failed", exit=False)
```
luke-lombardi, line 280: same comment as above

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/base/runner.py
nickpetrovic: I don't think so. Not for all cases. Take the taskqueue wrapper for example.

```
    def put(self, *args, **kwargs) -> bool:
        if not self.parent.prepare_runtime(
            func=self.func,
            stub_type=TASKQUEUE_STUB_TYPE,
        ):
            return False
```

nickpetrovic: I'm just adding messaging to these validations. I think it'll be fine, but it might show up in the logs if there are errors for things like taskqueue.put.

nickpetrovic:
Added `GpuTypeLiteral`. This allows for validating strings for the types.

image

luke-lombardi (approved):

PR #345, opened 2024-07-02
Add GPU availability endpoint for providers

nickpetrovic:
Resolve BE-1499


nickpetrovic:
@luke-lombardi @jsun-m 
Could use some thoughts/opinions on whether or not machines should be an abstraction. Should we consider a new category?

nickpetrovic:
Example of one can use this in their handler code.
```python
@endpoint()
def func():
    gpus = machines.gpus_available()
    print(f"{gpus=}")
    return {"gpus": gpus}
```

pkg/abstractions/machine/machine.go
```diff
@@ -0,0 +1,41 @@
...
+	providerRepo repository.ProviderRepository
+}
+
+func NewMachineRedisService(config map[string]types.WorkerPoolConfig, providerRepo repository.ProviderRepository) MachineService {
+	return &MachineRedisService{
+		providerRepo: providerRepo,
+	}
+}
+
+func (m *MachineRedisService) GetGPUAvailability(ctx context.Context, in *pb.GetGPUAvailabilityRequest) (*pb.GetGPUAvailabilityResponse, error) {
```
luke-lombardi, line 28: I'm a little unsure if this is an "abstraction" / belongs here. I guess the rationale is that since it's exposed in the SDK, its an abstraction -- not sure though.

sdk/src/beta9/cli/machine.py
```diff
@@ -189,3 +191,27 @@ def delete_machine(service: ServiceClient, machine_id: str, pool: str):
...
     else:
         terminal.error(f"Error: {res.err_msg}")
+
+
+@management.command(
+    name="available-gpus",
+    help="Get available GPU types.",
+)
+@extraclick.pass_service_client
+def available_gpus(service: ServiceClient):
```
luke-lombardi, line 201: @mernit Any thoughts on this interface to machine availability? Specifically on the name `available-gpus`

luke-lombardi (changes requested):
I didn't request any explicit changes here, just wanted to start the conversation with you/eli

sdk/src/beta9/cli/machine.py
mernit: I think it's fine @luke-lombardi @nickpetrovic

pkg/abstractions/machine/machine.go
nickpetrovic: Feels like a core api in a sense, but we really don't have that concept - at least not categorized. I also think if we define a new core api type of category, then we'll need to provide a stronger definition of what an abstraction is vs whatever other category we introduce.

But if we say an abstraction is something exposed to the SDK - like you said - then I think it makes sense to keep machines as an abstraction.

sdk/src/beta9/cli/machine.py
nickpetrovic: Should the CLI command and the SDK function have the same name at least? They're different because I wanted to see if there was a preference of one vs the other.

mernit: Yeah `machine.list()` should ideally list the available GPUs, maybe we should use that so its consistent

pkg/gateway/services/machine.go
```diff
@@ -13,13 +13,21 @@ import (
 
 func (gws *GatewayService) ListMachines(ctx context.Context, in *pb.ListMachinesRequest) (*pb.ListMachinesResponse, error) {
 	authInfo, _ := auth.AuthInfoFromContext(ctx)
+
+	gpus, err := gws.providerRepo.GetGPUAvailability(gws.appConfig.Worker.Pools)
```
nickpetrovic, line 17: @luke-lombardi with this, we will only see external provider GPUs, not local ones e.g. AWS. Is this the ideal behavior?

sdk/src/beta9/terminal.py
```diff
@@ -28,7 +36,19 @@ def header(text: str, subtext: str = "") -> None:
...
+        table.add_row(*("-" * len(table.columns)))
+
+    table.add_section()
+    table.add_row(f"[bold]{row_count} items")
+
+
 def print(*objects: Any, **kwargs: Any) -> None:
+    for obj in objects:
+        if isinstance(obj, Table):
+            _add_footer_to_table(obj)
```
nickpetrovic, line 51: This either needs to be removed or we need to remove all of the duplicate code that adds a count of items at the end of the table.

pkg/gateway/services/machine.go
luke-lombardi: yeah that's fine I think

luke-lombardi (approved):

PR #348, opened 2024-07-02
Soft-delete deployments

jsun-m:


pkg/api/v1/deployment.go
```diff
@@ -84,6 +86,11 @@ func (g *DeploymentGroup) RetrieveDeployment(ctx echo.Context) error {
 
 	deploymentId := ctx.Param("deploymentId")
 	if deployment, err := g.backendRepo.GetDeploymentByExternalId(ctx.Request().Context(), workspace.Id, deploymentId); err != nil {
+		if err == sql.ErrNoRows {
```
nickpetrovic, line 89: This doesn't check for bad input (id not being a uuid) and will surface as a 500 when it should probably be a 400.

Check this out.
https://github.com/beam-cloud/beta9/blob/main/pkg/repository/backend_postgres.go#L419-L427

pkg/api/v1/deployment.go
```diff
@@ -132,6 +144,44 @@ func (g *DeploymentGroup) StopAllActiveDeployments(ctx echo.Context) error {
 	return g.stopDeployments(deployments, ctx)
 }
 
+func (g *DeploymentGroup) DeleteDeployment(ctx echo.Context) error {
+	cc, ok := ctx.(*auth.HttpAuthContext)
+	if !ok {
+		return echo.NewHTTPError(http.StatusInternalServerError, "Failed to get auth info")
```
nickpetrovic, line 150: Is this needed? I thought we took care of this [here](https://github.com/beam-cloud/beta9/blob/main/pkg/auth/http.go#L83-L98).

pkg/api/v1/deployment.go
```diff
@@ -132,6 +144,44 @@ func (g *DeploymentGroup) StopAllActiveDeployments(ctx echo.Context) error {
...
+	}
+
+	deploymentId := ctx.Param("deploymentId")
+
+	// Get deployment
+	deploymentWithRelated, err := g.backendRepo.GetDeploymentByExternalId(ctx.Request().Context(), cc.AuthInfo.Workspace.Id, deploymentId)
+	if err != nil {
+		if err == sql.ErrNoRows {
```
nickpetrovic, line 158: Feels like we're leaking sql into the service. Can we return `nil, nil` when we don't find a deployment and there isn't an error? See my comment about the bad sql input.

pkg/api/v1/deployment.go
```diff
@@ -132,6 +144,44 @@ func (g *DeploymentGroup) StopAllActiveDeployments(ctx echo.Context) error {
...
+	if err := g.stopDeployments([]types.DeploymentWithRelated{*deploymentWithRelated}, ctx); err != nil {
+		return ctx.JSON(http.StatusInternalServerError, map[string]interface{}{
+			"error": "failed to stop deployment",
+		})
+	}
+
+	// Delete deployment
+	if err := g.backendRepo.DeleteDeployment(ctx.Request().Context(), deploymentWithRelated.Deployment); err != nil {
+		return ctx.JSON(http.StatusInternalServerError, map[string]interface{}{
+			"error": "failed to delete deployment",
```
nickpetrovic, line 178: Most of these errors start lower cased, but one doesn't. Can we choose a style and keep it consistent?

pkg/api/v1/deployment.go
jsun-m: nice, will add this

pkg/api/v1/deployment.go
jsun-m: yea you are right

nickpetrovic (approved):

PR #351, opened 2024-07-03
Add outputs to a task in the task api

nickpetrovic:
- Add a new `output` list to a `task` object that consists of outputs associated with the task
- Generates new public URLs for outputs that expire in 1h
- Reworks some previously private Output service methods to be public


nickpetrovic:
Example of what this looks like.

image

mernit (approved):

PR #360, opened 2024-07-10
Add crusoe and hydra boilerplate providers

luke-lombardi:


pkg/providers/base.go
```diff
@@ -63,8 +63,12 @@ func (p *ExternalProvider) Reconcile(ctx context.Context, poolName string) {
 		case <-ticker.C:
 			machines, err := p.ListMachinesFunc(ctx, poolName)
 			if err != nil {
-				log.Printf("<provider %s>: unable to list machines - %v\n", p.Name, err)
-				continue
+				if _, ok := err.(*types.ProviderNotImplemented); ok {
+					return
+				} else {
+					log.Printf("<provider %s>: unable to list machines - %v\n", "providerName", err)
```
luke-lombardi, line 69: If list isn't implemented - no reason to try reconciliation again.

mernit (approved):

pkg/providers/ec2.go
```diff
@@ -22,14 +22,14 @@ type EC2Provider struct {
 }
 
 func NewEC2Provider(ctx context.Context, appConfig types.AppConfig, providerRepo repository.ProviderRepository, workerRepo repository.WorkerRepository, tailscale *network.Tailscale) (*EC2Provider, error) {
-	cfg, err := common.GetAWSConfig(appConfig.Providers.EC2Config.AWSAccessKey, appConfig.Providers.EC2Config.AWSSecretKey, appConfig.Providers.EC2Config.AWSRegion, "")
+	cfg, err := common.GetAWSConfig(appConfig.Providers.EC2.AWSAccessKey, appConfig.Providers.EC2.AWSSecretKey, appConfig.Providers.EC2.AWSRegion, "")
```
nickpetrovic, line 25: When comparing this provider to Oracle and the others, would this provider be better named as `AWSProvider` instead of `EC2Provider` and the provider config `config.Providers.AWS.xxx` over `config.Providers.EC2.xxx`? Providers are AWS, OCI, GCP whereas their services are EC2 (Elastic Compute Cloud), Compute, and Compute Engine.

luke-lombardi: yea I've been thinking of that as well. I'll rename it in a follow-up PR today

PR #361, opened 2024-07-10
Add callback urls to containers and endpoints

jsun-m:
The way we add callbacks is not ideal but currently there are some nuances between abstractions that require task callbacks to be built case by case. 
One of these nuances is how we `end_task` in endpoint. It's location of handling is different than that of task queue, containers, and functions. 
We place `send_callback` after the `end_task` functions.

### DIFF
--- a/sdk/src/beta9/abstractions/endpoint.py
+++ b/sdk/src/beta9/abstractions/endpoint.py
@@ -105,6 +105,7 @@ class Endpoint(RunnerAbstraction):
         name: Optional[str] = None,
         authorized: Optional[bool] = True,
         autoscaler: Optional[Autoscaler] = QueueDepthAutoscaler(),
+        callback_url: Optional[str] = None,
     ):
         super().__init__(
             cpu=cpu,
@@ -122,6 +123,7 @@ class Endpoint(RunnerAbstraction):
             name=name,
             authorized=authorized,
             autoscaler=autoscaler,
+            callback_url=callback_url,
         )
 
         self._endpoint_stub: Optional[EndpointServiceStub] = None
--- a/sdk/src/beta9/abstractions/endpoint.py
+++ b/sdk/src/beta9/abstractions/endpoint.py
@@ -71,6 +71,8 @@ class Endpoint(RunnerAbstraction):
         autoscaler (Optional[Autoscaler]):
             Configure a deployment autoscaler - if specified, you can use scale your function horizontally using
             various autoscaling strategies (Defaults to QueueDepthAutoscaler())
+        callback_url (Optional[str]):
+            An optional URL to send a callback to when a task is completed, timed out, or cancelled.
     Example:
         ```python
         from beta9 import endpoint, Image
--- a/sdk/src/beta9/runner/common.py
+++ b/sdk/src/beta9/runner/common.py
@@ -13,7 +13,12 @@ from typing import Any, Callable, Optional, Union
 import requests
 from starlette.responses import Response
 
-from ..clients.gateway import GatewayServiceStub, SignPayloadRequest, SignPayloadResponse
+from ..clients.gateway import (
+    EndTaskRequest,
+    GatewayServiceStub,
+    SignPayloadRequest,
+    SignPayloadResponse,
+)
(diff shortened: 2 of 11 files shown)


sdk/src/beta9/abstractions/endpoint.py
```diff
@@ -105,6 +105,7 @@ def __init__(
         name: Optional[str] = None,
         authorized: Optional[bool] = True,
         autoscaler: Optional[Autoscaler] = QueueDepthAutoscaler(),
+        callback_url: Optional[str] = None,
```
nickpetrovic, line 108: Can we add an entry for this in the docstring?

sdk/src/beta9/runner/function.py
```diff
@@ -183,30 +189,24 @@ def main(channel: Channel):
 
         task_duration = time.time() - start_time
 
-        # End the task
-        end_task_response = gateway_stub.end_task(
-            EndTaskRequest(
+        end_task_response = end_task_and_send_callback(
```
nickpetrovic, line 192: Have we tested when a task times out or is cancelled now that this we send the callback before monitor_task.cancel() is called?

sdk/src/beta9/runner/common.py
```diff
@@ -188,6 +193,28 @@ def execute_lifecycle_method(name: str) -> Union[Any, None]:
         raise RunnerException()
 
 
+def end_task_and_send_callback(
+    *,
+    gateway_stub,
+    payload,
+    end_task_request: EndTaskRequest,
+):
```
luke-lombardi, line 201: let's add types here

sdk/src/beta9/runner/container.py
```diff
@@ -50,12 +51,25 @@ async def _run():
                     print(line.strip().decode("utf-8"))
 
                 if not self.killed:
-                    stub.end_task(
-                        EndTaskRequest(
+                    end_task_and_send_callback(
```
luke-lombardi, line 54: shouldn't we just send the callback regardless?

sdk/src/beta9/runner/endpoint.py
```diff
@@ -110,18 +118,23 @@ async def task_lifecycle(request: Request):
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR, detail="Failed to start task"
         )
 
+    task_lifecycle_data = TaskLifecycleData(
+        status=TaskStatus.Complete,
+        result=None,
```
luke-lombardi, line 123: won't this result always be None? what's the point of adding this dataclass wrapper?

luke-lombardi (changes requested):

sdk/src/beta9/runner/endpoint.py
jsun-m: the result can change. SImilar to how the task_status previously updated through the `yield`, we pass `task_lifecycle_data` by reference into the `api request` where it will update its values. Once this function gets passed the yield, the result is populated

sdk/src/beta9/runner/container.py
jsun-m: yea so if it was `killed` meaning it was canceled, we send a different callback with task status cancelled

sdk/src/beta9/runner/function.py
jsun-m: wdym? monitor tasks has its own send_callback functions

jsun-m: task cancel and timeout and cancelled still works properly

nickpetrovic: Does this invoke the callback just once if it times out?

sdk/src/beta9/runner/common.py
jsun-m: done

sdk/src/beta9/runner/function.py
jsun-m: should invoke it just once.

luke-lombardi (approved):

nickpetrovic (approved):

PR #362, opened 2024-07-10
Add latest version endpoints to deployment endpoints

nickpetrovic:
- Add support for `/` and `/latest` which points to the latest version of a deployment
- Fix a nil pointer error when not returning an error after GetDeploymentByNameAndVersion() returns one

Resolve BE-1532


mernit (approved):
Nice

jsun-m (approved):
lgtm

PR #363, opened 2024-07-10
GetLatestDeploymentByName should see deleted deployments

jsun-m:


pkg/repository/backend_postgres_migrations/010_add_deployment_deleted_at_constraint.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func upAddDeploymentDeletedAtUniqueConstraint(tx *sql.Tx) error {
+	_, err := tx.Exec(`ALTER TABLE deployment DROP CONSTRAINT deployment_name_version_workspace_id_stub_type_key RESTRICT`)
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.Exec(`ALTER TABLE deployment ADD CONSTRAINT unique_deploy UNIQUE (name, version, workspace_id, stub_type, deleted_at)`)
+	if err != nil {
```
nickpetrovic, line 20: Probably don't need this if statement.

nickpetrovic (commented):
lgtm. @luke-lombardi any reason we added the name here originally?

pkg/repository/backend_postgres_migrations/010_add_deployment_deleted_at_constraint.go
jsun-m: actually i removed this. Talked with luke, its probably better to keep the old constraint but new deployments are created with a bumped version on the latest (even if latest was deleted)

luke-lombardi (approved):

PR #369, opened 2024-07-11
Override callback urls by requests

jsun-m:


sdk/src/beta9/runner/common.py
```diff
@@ -210,18 +211,26 @@ def end_task_and_send_callback(
...
+    task_status: str,
+    override_callback_url: Optional[str] = None,
 ) -> None:
     """
     Send a signed callback request to an external host defined by the user
     """
-    if context.callback_url == "" or context.callback_url is None:
+
+    callback_url = override_callback_url or context.callback_url
+    if callback_url == "" or callback_url is None:
```
nickpetrovic, line 233: Can can we simplify this with `if not callback:`? That should cover these two conditions and any other data type condition that is falsy.

nickpetrovic (commented):
Can you help me understand why we need this? How would this override be set through the lifecycle data?

mernit:
> Can you help me understand why we need this? How would this override be set through the lifecycle data?

@nickpetrovic the reason we originally built this is because users with different dev/stage/prod environments want to fire a different callback depending on where the request is originating. We built this functionality into V1 based on that use case.

sdk/src/beta9/runner/common.py
jsun-m: yea you are right.

jsun-m:
The override be set at the API request.
```
{
    "kwargs": {
        "callback_url": "https://webhook.site/25114ac2-5a48-4d65-8538-546b91f057a0"
    }
}
```

PR #370, opened 2024-07-11
Collect and store container metrics from worker

nickpetrovic:
- Send container metrics to event repo (cpu, mem, gpu, and io)
- Add Elasticsearch to cluster-dev kustomize deployment

Resolve BE-1547

### DIFF
--- a/manifests/kustomize/components/monitoring/fluent-bit.yaml
+++ b/manifests/kustomize/components/monitoring/fluent-bit.yaml
@@ -51,7 +51,7 @@ valuesInline:
       [INPUT]
           Name   http
           Tag    events
-          Listen 0.0.0.0
+          Listen ::
           Port   9880
     outputs: |
       [OUTPUT]
@@ -73,6 +73,19 @@ valuesInline:
           use_put_object  On
           total_file_size 1M
           upload_timeout  1m
+      [OUTPUT]
+          Name                es
+          Match               events
+          Host                ${ELASTICSEARCH_HOST}
+          Port                ${ELASTICSEARCH_PORT}
+          Logstash_Format     On
+          Logstash_Prefix     events
+          Retry_Limit         False
+          Time_Key_Nanos      On
+          Suppress_Type_Name  On
+          Buffer_Size         False
+          Trace_Error         On
+          Replace_Dots        On
   volumeMounts:
   - name: config
     mountPath: /fluent-bit/etc/conf
--- a/pkg/worker/events.go
+++ b/pkg/worker/events.go
@@ -1,97 +1,247 @@
 package worker
 
 import (
-	"log"
+	"errors"
+	"fmt"
+	"path/filepath"
 	"time"
 
-	"github.com/beam-cloud/beta9/pkg/repository"
 	types "github.com/beam-cloud/beta9/pkg/types"
 	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/shirou/gopsutil/v4/net"
+	"github.com/shirou/gopsutil/v4/process"
 )
 
-func collectAndSendContainerMetrics(eventRepo repository.EventRepository, workerID string, request *types.ContainerRequest, spec *specs.Spec, pidChan <-chan int, done chan bool) {
-	gpuEnabled := request.Gpu != ""
-
-	ticker := time.NewTicker(types.ContainerResourceUsageEmissionInterval)
-	defer ticker.Stop()
-
+func (w *Worker) collectAndSendContainerMetrics(request *types.ContainerRequest, spec *specs.Spec, pidChan <-chan int, done chan bool) {
 	containerPid := <-pidChan
 	if containerPid == 0 {
 		return
 	}
 
-	containerProc, err := NewProcUtil(containerPid)
-	if err != nil {
-		log.Printf("Unable to get process for container %s: %v\n", request.ContainerId, err)
-		return
-	}
+	ticker := time.NewTicker(w.config.Monitoring.ContainerMetricsInterval)
+	defer ticker.Stop()
 
-	var prevProcCpuTime float64 = 0
-	var prevSystemCpuTime float64 = 0
+	monitor := NewProcessMonitor(containerPid, spec.Linux.Resources.Devices)
+
 	for {
 		select {
 		case <-done:
 			return
 
 		case <-ticker.C:
-			// Calculate GPU memory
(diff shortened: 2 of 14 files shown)


nickpetrovic:
Can confirm metrics are being shipped. This is my local env.

image

pkg/worker/events.go
```diff
@@ -0,0 +1,97 @@
...
+		case <-done:
+			return
+
+		case <-ticker.C:
+			// Calculate GPU memory
+			var gpuMemUsed int64
+			var gpuMemTotal int64
+			if gpuEnabled {
+				for _, device := range spec.Linux.Resources.Devices {
+					stats, err := GetGpuMemoryUsage(*device.Minor)
```
nickpetrovic, line 42: @luke-lombardi does this look right to you? I found that Minor should point to the device on disk as mentioned [here](https://man.archlinux.org/man/nvidia-smi.1.en#Minor_Number).

jsun-m: Maybe we can get the device and store it when worker starts up. I can investigate when i start on the multi-worker gpu nodes

pkg/worker/events.go
```diff
@@ -0,0 +1,97 @@
...
+		select {
+		case <-done:
+			return
+
+		case <-ticker.C:
+			// Calculate GPU memory
+			var gpuMemUsed int64
+			var gpuMemTotal int64
+			if gpuEnabled {
+				for _, device := range spec.Linux.Resources.Devices {
```
jsun-m, line 41: I know it's difficult to test locally but it seems like in the old repo we had a check for `nvmlActive`. What happened to that?

jsun-m (changes requested):

pkg/worker/events.go
nickpetrovic: This will go through all of the devices and attempt to use nvidia-smi to get this information. An error will be returned for those that don't work with nvidia-smi. This is kind of a brute force way of doing it, but I don't think we set `spec.Linux.Resources.Devices` for anything else at the moment. So this _should_ work.

pkg/worker/events.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/beam-cloud/beta9/pkg/repository"
+	types "github.com/beam-cloud/beta9/pkg/types"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func collectAndSendContainerMetrics(eventRepo repository.EventRepository, workerID string, request *types.ContainerRequest, spec *specs.Spec, pidChan <-chan int, done chan bool) {
+	gpuEnabled := request.Gpu != ""
+
+	ticker := time.NewTicker(types.ContainerResourceUsageEmissionInterval)
```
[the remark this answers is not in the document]
nickpetrovic: Updated.

pkg/worker/events.go
nickpetrovic: We got rid of the nvml dependency and opted for just using nvidia-smi instead. In v1, we had nvml and we had to have different go files per operating system e.g. darwin and linux. Because we develop on macos, we didn't get intellisense/code completion or static typing since we're developing for the linux side.

jsun-m: makes sense

manifests/kustomize/components/monitoring/fluent-bit.yaml
```diff
@@ -51,7 +51,7 @@ valuesInline:
       [INPUT]
           Name   http
           Tag    events
-          Listen 0.0.0.0
+          Listen ::
           Port   9880
```
jsun-m, line 55: isn't this just ipv6?

nickpetrovic: I believe it supports both. It works locally and we're ipv4 locally. 

```
k get svc -n monitoring
NAME                      TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)             AGE
fluent-bit                ClusterIP   10.43.146.100   <none>        2020/TCP,9880/TCP   4d
grafana                   ClusterIP   10.43.238.163   <none>        80/TCP              4d
victoria-logs-single      ClusterIP   None            <none>        9428/TCP            4d
victoria-metrics-single   ClusterIP   None            <none>        8428/TCP            4d
elasticsearch-master-hl   ClusterIP   None            <none>        9200/TCP,9300/TCP   23h
elasticsearch             ClusterIP   10.43.74.98     <none>        9200/TCP,9300/TCP   23h
elasticsearch-kibana      ClusterIP   10.43.16.106    <none>        5601/TCP            23h
```

pkg/worker/events.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+	}
+	return stat
+}
+
+// fetchNetworkIO gets network IO for all NICs globally, regardless of process.
+// Avoid using this data for external purposes.
+// TODO: Look into per-process network IO
+func (m *ProcessMonitor) fetchNetworkIO() (*net.IOCountersStat, error) {
+	counters, err := net.IOCountersByFile(false, filepath.Join("/proc", fmt.Sprint(m.pid), "net/dev"))
```
jsun-m, line 136: Interesting. Wouldn't this just get the network io for the process with `m.pid` so just the worker?

I'm confused how this file includes network io for all processes

jsun-m: Have you tried `findChildProcesses` and summing up the io of all the container processes?

nickpetrovic: Your thoughts are exactly right. I could just call the function IOCounters() instead if it clearer. 

I found that /proc/{pid}/nev/dev gives me the exact results as /proc/net/dev. So I did this just to show that it exists, but doesn't really matter.

jsun-m (approved):
overall looks really good. so much data is exported now

pkg/worker/events.go
nickpetrovic: I have - processes don't have a Net IO Counter function. Notice I'm using the `net` package here and this function in particular lets me point to a proc path.

jsun-m: interesting. Is it because maybe the worker is pid 1, so it already gets the sum of all the child io?

jsun-m: ah nvm, just tested it myself. its all the same it seems

nickpetrovic: I'm just going to switch to `counters, err := net.IOCounters(false)` to avoid the confusion and add a couple of comments about this.

PR #378, opened 2024-07-16
Add/Update stub related endpoints

jsun-m:


pkg/api/v1/stub.go
```diff
@@ -50,9 +60,37 @@ func (g *StubGroup) ListStubs(ctx echo.Context) error {
...
 	}
 }
+
+func (g *StubGroup) RetrieveStub(ctx echo.Context) error {
+	stubID := ctx.Param("stubId")
+	c, _ := ctx.(*auth.HttpAuthContext)
+
+	stub, err := g.backendRepo.GetStubByExternalId(ctx.Request().Context(), stubID)
+	if err != nil {
+		log.Println(err)
```
nickpetrovic, line 84: Should remove.

pkg/api/v1/stub.go
```diff
@@ -50,9 +60,37 @@ func (g *StubGroup) ListStubs(ctx echo.Context) error {
...
+		return HTTPInternalServerError("Failed to retrieve stub")
+	} else if stub == nil {
+		return HTTPNotFound()
+	}
+
+	if stub.Workspace.ExternalId != c.AuthInfo.Workspace.ExternalId && c.AuthInfo.Token.TokenType != types.TokenTypeClusterAdmin {
```
nickpetrovic, line 90: Can we check for this before the call to GetStubByExternalId? We have the stubid from the request and the token type.

nickpetrovic: NVM, I misread.

nickpetrovic (approved):

PR #383, opened 2024-07-17
Add delete volume command

nickpetrovic:
Resolve BE-1539


nickpetrovic:
@luke-lombardi @jsun-m I've written down some thoughts on this in [linear](https://linear.app/beam-dev/issue/BE-1539#comment-49996cb5). Curious about your thoughts.

luke-lombardi (approved):
lgtm

PR #385, opened 2024-07-18
Fix tmp file creation on windows (git bash)

kevin-cgc:


mernit:
Thanks for your contribution @kevin-cgc! We will review this shortly.

nickpetrovic:
Just tested this and seems to work on macOS as well.

PR #387, opened 2024-07-18
Add support for private AWS ECR images

nickpetrovic:
Resolve BE-1553


nickpetrovic:
An example that can be used with AWS IAM Identity Center (formerly AWS SSO) STS credentials.

```python
image = Image(
    base_image="111111111111.dkr.ecr.us-east-1.amazonaws.com/myapp:latest",
    base_image_creds=[
        "AWS_ACCESS_KEY_ID",
        "AWS_SECRET_ACCESS_KEY",
        "AWS_SESSION_TOKEN",
        "AWS_REGION",
    ],
)


@endpoint(image=image)
def square(i: int = 0) -> int:
    return i**2
```

pkg/abstractions/image/build.go
```diff
@@ -41,11 +41,12 @@ type BuildOpts struct {
 	BaseImageRegistry  string
 	BaseImageName      string
 	BaseImageTag       string
+	BaseImageCreds     string
```
jsun-m, line 44: Will all image creds always be resolved to being a single string token?

jsun-m, line 44: Will all image creds always be resolved to being a single string token?

pkg/abstractions/image/build.go
```diff
@@ -387,3 +367,28 @@ func (b *Builder) generatePipInstallCommand(opts *BuildOpts) string {
 	packages := strings.Join(escapedPackages, " ")
 	return fmt.Sprintf("%s -m pip install --root-user-action=ignore %s", opts.PythonVersion, packages)
 }
+
+var imageNamePattern = regexp.MustCompile(`^(?:([^/]+)(?:/))?((?:[^/:]+/)?[^/:]+)(?::([^/]+))?$`)
```
jsun-m, line 371: Is there a source for this? We should probably add a comment to some url where this verified

pkg/abstractions/image/credentials.go
```diff
@@ -0,0 +1,90 @@
...
+		return "", fmt.Errorf("error decoding token: %v", err)
+	}
+
+	parts := strings.SplitN(string(decodedToken), ":", 2)
+	if len(parts) != 2 {
+		return "", fmt.Errorf("decoded token format invalid, expected username:password")
+	}
+
+	// Return the username and password
+	var username = parts[0]
```
jsun-m, line 85: I suggest we use `:=` over `var` for single use local variables for most cases

jsun-m (changes requested):

pkg/abstractions/image/build.go
nickpetrovic: There isn't. It was trial and error. There are tests for this that should cover all cases if there's a concern.

pkg/abstractions/image/build.go
nickpetrovic: Yes, I believe so. Skopeo only has `--src-creds` or `--username` and `--password` flags.

How we get this username/password is where the details lie e.g. from an AWS account or from a GCP account.

pkg/abstractions/image/credentials.go
nickpetrovic: Updated.

pkg/abstractions/image/credentials.go
```diff
@@ -0,0 +1,90 @@
...
+	parts := strings.SplitN(string(decodedToken), ":", 2)
+	if len(parts) != 2 {
+		return "", fmt.Errorf("decoded token format invalid, expected username:password")
+	}
+
+	// Return the username and password
+	username := parts[0]
+	password := parts[1]
+
+	var token = fmt.Sprintf("%s:%s", username, password)
```
jsun-m, line 88: you missed one

jsun-m (approved):
just that last `var` and it's good.

pkg/abstractions/image/credentials.go
nickpetrovic: Updated.

PR #388, opened 2024-07-18
Reconfigure worker and agent config for multi-worker agents

jsun-m:
There are issues with multi-worker workloads stemming from shared access to host ports.
1. Remove container assignment of host port 9090 from all pods in external manifest. We don't need a prometheus metrics endpoint since we push metrics
2. Dynamically acquire any free host port for RunCServer. Currently all workers attempt to acquire port 1989

### DIFF
--- a/pkg/worker/runc_server.go
+++ b/pkg/worker/runc_server.go
@@ -55,7 +55,7 @@ func NewRunCServer(containerInstances *common.SafeMap[*ContainerInstance], image
 }
 
 func (s *RunCServer) Start() error {
-	listener, err := net.Listen("tcp", ":0") // Random port
+	listener, err := net.Listen("tcp", ":0") // // Random free port
 	if err != nil {
 		log.Fatalf("failed to listen: %v\n", err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/worker/runc_server.go
```diff
@@ -54,11 +55,14 @@ func NewRunCServer(containerInstances *common.SafeMap[*ContainerInstance], image
 }
 
 func (s *RunCServer) Start() error {
-	listener, err := net.Listen("tcp", fmt.Sprintf(":%d", defaultWorkerServerPort))
+	listener, err := net.Listen("tcp", ":0") // Random port
```
luke-lombardi, line 58: // Random free port

pkg/worker/worker.go
```diff
@@ -544,7 +544,7 @@ func (s *Worker) spawn(request *types.ContainerRequest, bundlePath string, spec
 	go s.updateContainerStatus(request)
 
 	// Set worker hostname
-	hostname := fmt.Sprintf("%s:%d", s.podAddr, defaultWorkerServerPort)
+	hostname := fmt.Sprintf("%s:%d", s.podAddr, s.runcServer.port)
```
luke-lombardi, line 547: is there nowhere else we need to adjust?

luke-lombardi (changes requested):

pkg/worker/worker.go
jsun-m: there is no where else to adjust for `defaultWorkerServerPort`. Other places get the worker runc server address through the containerRepo redis row

luke-lombardi (approved):

PR #389, opened 2024-07-20
Improve displaying invocation details

nickpetrovic:
- Print public URL when authorized is false
- Generate URL on Gateway side using the external URL in the config
- Fix type hints

Resolve BE-1544


nickpetrovic:
Example output.
image

PR #394, opened 2024-07-23
Multi-Worker node with Multi-GPUs

jsun-m:


jsun-m:
I had to extract the nvidia-smi command out into its own interface cause it caused failure in tests and i needed a way to override the functions involving the command

nickpetrovic (changes requested):

pkg/worker/nvidia_smi.go
```diff
@@ -0,0 +1,125 @@
...
+	UsedCapacity  int64
+	TotalCapacity int64
+}
+
+type NvidiaSMIClientInterface interface {
+	AvailableGPUDevices() ([]int, error)
+	GetGpuMemoryUsage(deviceIndex int) (GpuMemoryUsageStats, error)
+}
+
+type NvidiaSMIClient struct{}
```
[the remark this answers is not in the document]
jsun-m: It's really stupid, i mainly add this so that I can easily override functions for testing purposes. In this case (AvailableGPUDevices), the `nvidia-smi` call was causing tests to fail so i wanted a way to easily override this function for those tests.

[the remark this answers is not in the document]
nickpetrovic: I see. In that case, it might make sense to do the latter which is moving the functions to ContainerCudaManager then embedding an interface into ContainerCudaManager so that it requires AvailableGPUDevices and GetGpuMemoryUsage be implemented. 

This might be better long term in that if we do implement an AMD manager of some sort, it would also require AvailableGPUDevices and GetGpuMemoryUsage be implemented. 

What do you think? Can we give it a try?

pkg/worker/events.go
```diff
@@ -119,7 +120,7 @@ func (m *ProcessMonitor) fetchGPUMemory() *GPUInfoStat {
 		if device.Minor == nil || !device.Allow {
 			continue
 		}
-		stats, err := GetGpuMemoryUsage(int(*device.Minor))
+		stats, err := m.gpuInfoClient.GetGPUMemoryUsage(int(*device.Minor))
```
nickpetrovic, line 123: Were you able to confirm using Minor works for multi-gpu nodes? If so, this is great.

pkg/worker/gpu_info.go
```diff
@@ -0,0 +1,125 @@
...
+	"strconv"
+	"strings"
+)
+
+type GPUInfoClient interface {
+	AvailableGPUDevices() ([]int, error)
+	GetGPUMemoryUsage(deviceIndex int) (GpuMemoryUsageStats, error)
+}
+
+type GpuMemoryUsageStats struct {
```
nickpetrovic, line 18: Nitpick, can we change Gpu to GPU for this?

nickpetrovic (approved):
I think this looks good. 

@luke-lombardi could you take a look just in case?

pkg/worker/gpu_info.go
jsun-m: done

pkg/worker/events.go
jsun-m: I updated the function to use the device obtained from `AvailableDevices`. This way it's consistent with the code that assigned it.
```
func (m *ProcessMonitor) fetchGPUMemory() *GPUInfoStat {
	stat := &GPUInfoStat{}
	availableDevices, err := m.gpuInfoClient.AvailableGPUDevices()
	if err != nil {
		return stat
	}

	for _, device := range availableDevices {
		stats, err := m.gpuInfoClient.GetGPUMemoryUsage(device)
		if err == nil {
			stat.MemoryUsed += uint64(stats.UsedCapacity)
			stat.MemoryTotal += uint64(stats.TotalCapacity)
		}
	}
	return stat
}
```

PR #397, opened 2024-07-25
Allow dots and underscores in repo name

dleviminzi:
This image still fails for me because the repo name contains dots: `nvcr.io/nim/meta/llama-3.1-8b-instruct:1.1.0`. It can be fixed by updating the regex to: `(?P<Repo>[a-z0-9][-a-z0-9._]+)`


nickpetrovic (approved):
lgtm

PR #401, opened 2024-07-25
Add config for shared memory limit

nickpetrovic:
Resolve BE-1578


pkg/scheduler/pool.go
```diff
@@ -81,3 +87,29 @@ func freePoolCapacity(workerRepo repository.WorkerRepository, wpc WorkerPoolCont
 
 	return capacity, nil
 }
+
+// calculateMemoryQuantity calculates the memory based on the percentage and total memory.
+func calculateMemoryQuantity(percent float32, memoryTotal int64) resource.Quantity {
```
luke-lombardi, line 92: I feel like these functions could be consolidated, into maybe one function

luke-lombardi (changes requested):

pkg/scheduler/pool.go
nickpetrovic: I can do that if we want it to be specific to this. I started with it that way, but wasn't sure if I liked it.

nickpetrovic: Updated.

luke-lombardi (approved):

PR #402, opened 2024-07-26
Retry juicefs umount

jsun-m:


pkg/storage/juicefs.go
```diff
@@ -122,13 +123,29 @@ func (s *JuiceFsStorage) Format(fsName string) error {
...
 
 func (s *JuiceFsStorage) Unmount(localPath string) error {
-	cmd := exec.Command("juicefs", "umount", "--force", localPath)
+	juiceFsUmount := func() error {
+		cmd := exec.Command("juicefs", "umount", localPath)
 
-	output, err := cmd.CombinedOutput()
+		output, err := cmd.CombinedOutput()
+		if err != nil {
+			log.Printf("error executing juicefs umount: %v, output: %s", err, string(output))
```
luke-lombardi, line 131: Can you clean up this log

pkg/storage/juicefs.go
```diff
@@ -122,13 +123,29 @@ func (s *JuiceFsStorage) Format(fsName string) error {
...
+		log.Printf("JuiceFS filesystem unmounted from: '%s'\n", localPath)
+		return nil
+	}
+
+	err := backoff.Retry(juiceFsUmount, backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), 10))
+	if err == nil {
+		return nil
+	}
+
+	// Forcefully kill the fuse mount devices
```
luke-lombardi, line 144: Could this affect other fuse mounts of other workers running on multi-worker nodes?

luke-lombardi (changes requested):

pkg/storage/juicefs.go
jsun-m: It only kills pods that it finds using that file with fuser /dev/fuse which results to only files on that pod

pkg/storage/juicefs.go
jsun-m: it's the same log as [here](https://github.com/beam-cloud/beta9/pull/402/files#diff-c6dd2c832240e75f3b4b734ef955e8cfeba0121667164d22f6bc10da2e881118L129)

luke-lombardi (approved):

PR #404, opened 2024-07-30
Fix image builds

nickpetrovic:
- Create and delete OCI directory so Skopeo can copy OCI image
- Revert changing bundle path to include repo name


luke-lombardi:
@nickpetrovic should we keep this open? not sure this is really an issue

nickpetrovic:
> @nickpetrovic should we keep this open? not sure this is really an issue

It works as-is. I'll close this.

PR #409, opened 2024-07-31
remove CI env check in get_channel

dleviminzi:


Makefile
```diff
@@ -22,6 +22,11 @@ k3d-up:
 k3d-down:
 	bash bin/k3d.sh down
 
+restart:
```
nickpetrovic, line 25: Can we name this `k3d-rebuild` instead? Just want it to be consistent with the other k3d targets.

nickpetrovic (approved):
lgtm. Just added the one comment.

Makefile
dleviminzi: will do!

luke-lombardi (approved):

PR #418, opened 2024-08-01
Fix import error in function runner

luke-lombardi:
- Fixes an edge case if an import fails during the initial instantiation of FunctionHandler

If import failed, then kwargs/args would not be defined, causing the function container to never exit.


nickpetrovic (approved):
makes sense to me 👍

PR #420, opened 2024-08-01
SDK housekeeping

nickpetrovic:
- Fix context not being passed when running `beta9 deploy <fqfn> --name <name> --context <my-context>`
- Fix gpu literal types by using hyphens instead of underscores
- Add a Deployable mixin class for endpoint, function, and taskqueue to use
- Update various docstrings

### DIFF
--- a/sdk/src/beta9/abstractions/mixins.py
+++ b/sdk/src/beta9/abstractions/mixins.py
@@ -1,0 +1,48 @@
+from typing import Callable, Optional
+
+from .. import terminal
+from ..clients.gateway import DeployStubRequest, DeployStubResponse
+from ..config import ConfigContext
+from .base.runner import RunnerAbstraction
+
+
+class DeployableMixin:
+    func: Callable
+    parent: RunnerAbstraction
+    deployment_stub_type: str
+
+    def _validate(self):
+        if not hasattr(self, "func") or not isinstance(self.func, Callable):
+            raise AttributeError("func variable not set or is incorrect type")
+        if not hasattr(self, "parent") or not isinstance(self.parent, RunnerAbstraction):
+            raise AttributeError("parent variable not set or is incorrect type")
+        if not hasattr(self, "deployment_stub_type") or not self.deployment_stub_type:
+            raise AttributeError("deployment_stub_type variable not set")
+
+    def deploy(self, name: Optional[str] = None, context: Optional[ConfigContext] = None) -> bool:
+        self._validate()
+
+        name = name or self.parent.name
+        if not name or name == "":
+            terminal.error(
+                "You must specify an app name (either in the decorator or via the --name argument)."
+            )
+
+        if context is not None:
+            self.parent.config_context = context
+
+        if not self.parent.prepare_runtime(
+            func=self.func, stub_type=self.deployment_stub_type, force_create_stub=True
+        ):
+            return False
+
+        terminal.header("Deploying")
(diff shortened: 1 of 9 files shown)


sdk/src/beta9/abstractions/mixins.py
```diff
@@ -0,0 +1,48 @@
...
+
+class DeployableMixin:
+    func: Callable
+    parent: RunnerAbstraction
+    deployment_stub_type: str
+
+    def _validate(self):
+        if not hasattr(self, "func") or not isinstance(self.func, Callable):
+            raise AttributeError("func variable not set or is incorrect type")
+        if not hasattr(self, "parent") or not isinstance(self.parent, RunnerAbstraction):
```
luke-lombardi, line 17: nit pick but can we space this out a bit. kind of hard to read

sdk/src/beta9/abstractions/mixins.py
```diff
@@ -0,0 +1,48 @@
...
+            raise AttributeError("parent variable not set or is incorrect type")
+        if not hasattr(self, "deployment_stub_type") or not self.deployment_stub_type:
+            raise AttributeError("deployment_stub_type variable not set")
+
+    def deploy(self, name: Optional[str] = None, context: Optional[ConfigContext] = None) -> bool:
+        self._validate()
+
+        name = name or self.parent.name
+        if not name or name == "":
```
luke-lombardi, line 26: pretty sure this can just be if not name (name="" is falsey)

sdk/src/beta9/cli/deployment.py
```diff
@@ -112,13 +113,13 @@ def create_deployment(service: ServiceClient, name: str, entrypoint: str):
...
 
-    user_func = getattr(module, func_name, None)
+    user_func: Optional[DeployableMixin] = getattr(module, func_name, None)
     if user_func is None:
         terminal.error(
             f"Invalid handler function specified. Make sure '{module_path}' contains the function: '{func_name}'"
         )
 
-    if not user_func.deploy(name=name):  # type:ignore
+    if not user_func.deploy(name=name, context=service._config):  # type: ignore
```
luke-lombardi, line 122: why wasn't this required before?

luke-lombardi (changes requested):
lgtm, couple nit picks

sdk/src/beta9/cli/deployment.py
nickpetrovic: It never worked before for the `deploy` command, just for the other commands e.g. `task list` etc. This is because the deploy command is instantiated via the BaseAbstraction class whereas everything else in the CLI uses the ServiceClient gRPC stub wrapper. 

The ServiceClient is created at the CLI entrypoints where we use the `@pass_service_client` decorator which does look for `--context` and if it finds it, uses it to configure channel.

Basically, BaseAbstraction doesn't deal with contexts unless we provide it like we do in the ServiceClient class.

sdk/src/beta9/abstractions/mixins.py
nickpetrovic: Updated.

sdk/src/beta9/abstractions/mixins.py
nickpetrovic: Updated.

sdk/src/beta9/cli/deployment.py
luke-lombardi: Makes sense 👍

luke-lombardi (approved):

PR #427, opened 2024-08-05
Add drop-in config directory

nickpetrovic:
Since our config library (koanf) allows us to load and overwrite configs in the order they are loaded, the plan for this is to be able to generate a config (Secret) via helm and mount it under the drop-in config directory.

- Load configs from drop-in directory (/etc/beta9.d/)
- Reuse top-level config in taskqueue and endpoint abstractions
- Don't use global config format when load configs

Resolve BE-1593


luke-lombardi (approved):
lgtm 👍

PR #429, opened 2024-08-05
dlm/image-load-logs

dleviminzi:
This adds logs around image loading to the log file so that they are visible to users.


pkg/worker/worker.go
```diff
@@ -262,30 +262,39 @@ func (s *Worker) shouldShutDown(lastContainerRequest time.Time) bool {
...
+	f, err := NewLogger(request)
+	if err != nil {
+		return err
+	}
+	defer f.logFile.Close()
+
 	// Pull image
-	log.Printf("<%s> - lazy-pulling image: %s\n", containerID, request.ImageId)
-	err := s.imageClient.PullLazy(request)
+	f.Log("loading image: %s", request.ImageId)
```
luke-lombardi, line 276: I don't know that we need these logs sent to the user -- imo we only need the ones generated in the image client (that were tied directly to blobfs

pkg/worker/container_logger.go
```diff
@@ -0,0 +1,117 @@
+package worker
```
luke-lombardi, line 1: If we only capture those few logs in the image client surrounding the blobfs caching, do we need to refactor this out to a separate file / logger? Feels like overkill

luke-lombardi (changes requested):

pkg/worker/worker.go
dleviminzi: I bumped the logging down to be only in the image client.

pkg/worker/container_logger.go
dleviminzi: I wasn't sure if we would want to add similar logging in other places in the future. In any case, I've removed this refactoring for now.

pkg/worker/image.go
```diff
@@ -134,10 +142,17 @@ func blobfsAvailable(path string) bool {
 }
 
 func (c *ImageClient) PullLazy(request *types.ContainerRequest) error {
-	imageId := request.ImageId
+	f, err := newLogger(request)
```
luke-lombardi, line 145: Is the reason for the struct so that we can also log to stdout? Seems like we might be able to just have a function on the existing Logger similar to what you have below (`openLogFile`) which is like OpenContainerLogFile or something that just returns the file instead of a new struct

luke-lombardi (changes requested):

pkg/worker/image.go
dleviminzi: Ok I went ahead and put it all in a method on the `ContainerLogger`

luke-lombardi (approved):

PR #434, opened 2024-08-06
ASGI decorator & stub type

luke-lombardi:
- Adds an additional @asgi decorator to sdk (as a special case of the `endpoint` abstraction)
- Adds the ability to host arbitrary fastapi/ASGI compatible apps to beta9

```
from fastapi import FastAPI

from beta9 import Image, asgi, endpoint


def load_something_heavy():
    return [1, 2, 3, 4, 5]


@asgi(image=Image(python_version="python3.8"), on_start=load_something_heavy)
def handler(context):
    myapp = FastAPI()
    my_model = context.on_start_value

    @myapp.post("/nonsense")
    async def something():
        return my_model

    @myapp.post("/retrain")
    async def test2():
        return {"hello": True}

    return myapp
```


dleviminzi (approved):

sdk/src/beta9/runner/endpoint.py
```diff
@@ -158,6 +120,59 @@ async def _keep_worker_alive(self):
...
+                )
+
+            task_lifecycle_data = TaskLifecycleData(
+                status=TaskStatus.Complete, result=None, override_callback_url=None
+            )
+
+            try:
+                request.state.task_lifecycle_data = task_lifecycle_data
+                response = await call_next(request)
+                print(f"Task <{task_id}> finished")
```
jsun-m, line 156: Now that we get the `response` back directly, we don't need to do this convoluted way to passing the result and task.status. We can make it so we can extract it directly from the response

In this case, `task_lifecycle_data.result` is just `response.body`.

Also I still don't understand how we will get the task status update for user specified asgi endpoints.

PR #435, opened 2024-08-06
fix invocation url for asgi

dleviminzi:


luke-lombardi (approved):
Just need to bump the sdk version

PR #439, opened 2024-08-09
Add events for task.created and task.updated

jsun-m:
I'm looking for opinions on where to push events for these backend transactions. 
Rather than updating in 15 places where `backendRepo.UpdateTask` is called, I currently have the event happen after the database transaction is fulfilled.

Alternative to this, I can write a layer on top of `backendRepo` such as `backendRepoWithEvents` or something similar that wraps functions such as 
- `UpdateTask`
- `CreateTask`
- and other places where we might want events following a database transaction such as `CreateDeployment`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ proxy:
 
 runner:
 	for target in py312 py311 py310 py39 py38; do \
-		docker build . --no-cache --target $$target -f ./docker/Dockerfile.runner -t localhost:5001/beta9-runner:$$target-$(runnerTag); \
+		docker build . --no-cache --target $$target --platform=linux/amd64 -f ./docker/Dockerfile.runner -t localhost:5001/beta9-runner:$$target-$(runnerTag); \
 		docker push localhost:5001/beta9-runner:$$target-$(runnerTag); \
 	done
 
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ proxy:
 
 runner:
 	for target in py312 py311 py310 py39 py38; do \
-		docker build . --no-cache --target $$target --platform=linux/amd64 -f ./docker/Dockerfile.runner -t localhost:5001/beta9-runner:$$target-$(runnerTag); \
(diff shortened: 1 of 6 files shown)


pkg/types/event.go
```diff
@@ -15,6 +19,25 @@ var (
...
+		stub.deployed
+		stub.served
+
+		container.lifecycle.updated
+		worker.lifecycle.updated
+
+		the container.metrics event is slightly different because they are snapshots of a resource not a change in state
+		we might need to altered this or rethink concept of an event
+	*/
+	EventTaskUpdated = "task.update"
```
luke-lombardi, line 39: we should make this consisent, if it's past tense / present tense

pkg/repository/base.go
```diff
@@ -149,6 +149,8 @@ type EventRepository interface {
 	PushDeployStubEvent(workspaceId string, stub *types.Stub)
 	PushServeStubEvent(workspaceId string, stub *types.Stub)
 	PushRunStubEvent(workspaceId string, stub *types.Stub)
+	PushTaskUpdatedEvent(task *types.TaskWithRelated)
```
luke-lombardi, line 152: 👍

Makefile
```diff
@@ -42,7 +42,7 @@ proxy:
 
 runner:
 	for target in py312 py311 py310 py39 py38; do \
-		docker build . --no-cache --target $$target --platform=linux/amd64 -f ./docker/Dockerfile.runner -t localhost:5001/beta9-runner:$$target-$(runnerTag); \
```
luke-lombardi, line 45: put back

luke-lombardi (changes requested):

Makefile
jsun-m: I think i added this a while back when i was compiling workers for lambda locally. It dramatically increases compile time when building workers locally but was necessary before adding the github action

luke-lombardi: it's necessary for building the correct arch on m1s / arm hardware

luke-lombardi: this is for the runner image, not for the worker

luke-lombardi (changes requested):
left one more comment then good to go

luke-lombardi (approved):

PR #442, opened 2024-08-12
Enable multi-gpu containers

luke-lombardi:
- Adds SDK support for multiple GPUs per container

**NOTE**:  This limits the max gpu_count to 2 for now


jsun-m:
I don't think this will work out of the box? 
Currently we force all GPU's on the agent to be filled up to the max with 1 GPU workers.

luke-lombardi:
> I don't think this will work out of the box? Currently we force all GPU's on the agent to be filled up to the max with 1 GPU workers.

This is true. We'd have to update the pool sizing logic to specify the defaultGpuCount > 1 for this to work. We can set it to 2 and I believe it should work.

pkg/gateway/services/stub.go
```diff
@@ -30,37 +30,6 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
 
 	gpus := types.GPUTypesFromString(in.Gpu)
 
-	if len(gpus) > 0 {
```
luke-lombardi, line 33: I just moved this down so I can use the new RequiresGPU struct method

pkg/gateway/services/stub.go
```diff
@@ -106,6 +76,42 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
...
+				Ok:     false,
+				ErrMsg: "Failed to get GPU counts.",
+			}, nil
+		}
+
+		// T4s are currently in a different pool than other GPUs and won't show up in gpu counts
+		lowGpus := []string{}
+
+		for _, gpu := range gpus {
+			if gpuCounts[gpu.String()] <= 1 && gpu.String() != types.GPU_T4.String() {
```
nickpetrovic, line 105: I know this is probably out of scope, but should we remove this T4 check now that we have external provider T4 availability?

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -130,6 +132,12 @@ def __init__(
         )
         self.extra: dict = {}
 
+        if (self.gpu != "" or len(self.gpu) > 0) and self.gpu_count == 0:
+            self.gpu_count = 1
+
+        if self.gpu_count > 2:
+            raise ValueError("GPU count cannot be greater than 2")
```
nickpetrovic, line 139: I think this would be better on the server side - potentially under the stubLimits config - to prevent anyone from overriding this client side.

pkg/gateway/services/stub.go
nickpetrovic: That is true, but why would we have different behavior for A10Gs and T4s? A10Gs can also be scaled out in AWS since we have them as a lower priority worker pool.

nickpetrovic: We can work on this another time since its out of scope anyway.

nickpetrovic (approved):

PR #443, opened 2024-08-12
dlm/toggle-token

dleviminzi:
* Adds `/:workspaceId/:tokenId/toggle` route to the gateway.
  * Sets `active` field of token to true or false depending on current state.


pkg/api/v1/token.go
```diff
@@ -92,3 +95,20 @@ func (g *TokenGroup) ListWorkspaceTokens(ctx echo.Context) error {
...
+	workspaceId := ctx.Param("workspaceId")
+	workspace, err := g.backendRepo.GetWorkspaceByExternalId(ctx.Request().Context(), workspaceId)
+	if err != nil {
+		return HTTPBadRequest("Invalid workspace ID")
+	}
+
+	extTokenId := ctx.Param("tokenId")
+	token, err := g.backendRepo.ToggleToken(ctx.Request().Context(), workspace.Id, extTokenId)
+	if err != nil {
+		slog.Error("Failed to toggle token", "error", err)
```
nickpetrovic, line 109: We have I think 2 different types of logs right now. Do we really want to introduce another type of log?

dleviminzi: Whoops, that was for debugging. Removing it

nickpetrovic (approved):

luke-lombardi (approved):

PR #448, opened 2024-08-13
delete tokens and return full token on create

dleviminzi:
This PR adds support for additional support for tokens to the gateway. It allows us to delete tokens and we will now return the full token object upon creation. This is useful because it allows users to make subsequent requests using the external token ID without needing to fetch or re-fetch the full list of tokens.


luke-lombardi (approved):
lgtm 👍

PR #450, opened 2024-08-14
Add token command to the cli 

dleviminzi:
This PR adds token functionality to the beta9 cli. The added functionality will also be available in the UI on the beam platform. This change will let users `list`, `create`, `delete`, and `toggle` tokens. 
```
      # List tokens
      beta9 token list

      # List tokens and output in JSON format
      beta9 token list --format json
      
      # Create new token
      beta9 token create

      # Delete a token. 
      beta9 token delete 49554ce9-5861-4834-899b-ec301e5d8534

      # Toggle a token between active/inactive
      beta9 token toggle 49554ce9-5861-4834-899b-ec301e5d8534
```


sdk/src/beta9/cli/token.py
```diff
@@ -0,0 +1,183 @@
...
+      {cli_name} token toggle 49554ce9-5861-4834-899b-ec301e5d8534
+
+      \b
+    """,
+)
+@click.argument("token_id", type=click.STRING)
+@extraclick.pass_service_client
+def toggle_token(
+    service: ServiceClient,
+    token_id: int,
```
nickpetrovic, line 171: Should be a `str`.

sdk/src/beta9/cli/token.py
```diff
@@ -0,0 +1,183 @@
...
+    res: ToggleTokenResponse
+    res = service.gateway.toggle_token(ToggleTokenRequest(external_id=token_id))
+
+    if not res.ok:
+        terminal.error(res.err_msg)
+    else:
+        token = res.token.to_dict(casing=Casing.SNAKE)
+        if "active" not in token:
+            token["active"] = False
+        terminal.success("Token toggled successfully.")
```
nickpetrovic, line 182: Instead of sayings its been toggled, can we say something like, "Token is active" or "Token is disabled" for a message?

pkg/gateway/gateway.proto
```diff
@@ -365,3 +372,47 @@ message DeleteMachineResponse {
...
+
+message CreateTokenRequest {}
+
+message CreateTokenResponse {
+  bool ok = 1;
+  string err_msg = 2;
+  Token token = 3;
+}
+
+message ToggleTokenRequest { string external_id = 1; }
```
nickpetrovic, line 405: I think we want to use `id` or `token_id` for this. I don't think we typically expose the external_id as "external".

pkg/gateway/gateway.proto
```diff
@@ -365,3 +372,47 @@ message DeleteMachineResponse {
   bool ok = 1;
   string err_msg = 2;
 }
+
+message Token {
+  uint32 id = 1;
+  string external_id = 2;
```
nickpetrovic, line 378: `external_id` should be `id` or `token_id`. The `id` (regular int, not uuid) is an internal id that we don't typically expose externally.

pkg/gateway/gateway.proto
```diff
@@ -365,3 +372,47 @@ message DeleteMachineResponse {
...
+
+message ToggleTokenRequest { string external_id = 1; }
+
+message ToggleTokenResponse {
+  bool ok = 1;
+  string err_msg = 2;
+  Token token = 3;
+}
+
+message DeleteTokenRequest { string external_id = 1; }
```
nickpetrovic, line 413: Same comment as above - use `token_id` or `id` instead.

pkg/gateway/services/token.go
```diff
@@ -0,0 +1,130 @@
...
+	for _, token := range tokens {
+		var workspaceId uint32
+		if token.WorkspaceId != nil {
+			workspaceId = uint32(*token.WorkspaceId)
+		}
+
+		updatedAt := *timestamppb.New(token.UpdatedAt)
+
+		t = append(t, &pb.Token{
+			Id:                     uint32(token.Id),
```
nickpetrovic, line 35: I don't think we want this send out. Similar to the other comments.

pkg/gateway/services/token.go
```diff
@@ -0,0 +1,130 @@
...
+			ExternalId:             token.ExternalId,
+			Key:                    token.Key,
+			Active:                 token.Active,
+			Reusable:               token.Reusable,
+			WorkspaceId:            &workspaceId,
+			TokenType:              token.TokenType,
+			CreatedAt:              timestamppb.New(token.CreatedAt),
+			UpdatedAt:              &updatedAt,
+			DisabledByClusterAdmin: token.DisabledByClusterAdmin,
```
nickpetrovic, line 111: I don't know if we want this or not. @luke-lombardi what do you think?

nickpetrovic (changes requested):

pkg/gateway/services/token.go
dleviminzi: I removed it, but can always add it back

nickpetrovic: Sounds good.

sdk/src/beta9/cli/token.py
```diff
@@ -0,0 +1,183 @@
...
+    res = service.gateway.toggle_token(ToggleTokenRequest(external_id=token_id))
+
+    if not res.ok:
+        terminal.error(res.err_msg)
+    else:
+        token = res.token.to_dict(casing=Casing.SNAKE)
+        if "active" not in token:
+            token["active"] = False
+        terminal.success(
+            f"Token {res.token.token_id} {"enabled" if  token["active"] else "disabled"}."
```
nickpetrovic, line 181: Check if this works in Python 3.8. For some reason, I think they improved f-strings e.g. double quotes within double quotes in more recent versions.

nickpetrovic (approved):
lgtm. Just the one last comment about f-strings.

sdk/src/beta9/cli/token.py
dleviminzi: switched to single quotes

nickpetrovic (approved):

PR #451, opened 2024-08-14
Allow container networking to more configurable

luke-lombardi:
- Add a container network manager to abstract container network namespaces
- Add expose port methods to allow users to eventually expose specific ports from their containers


pkg/worker/network.go
```diff
@@ -1,13 +1,449 @@
...
+	"github.com/beam-cloud/beta9/pkg/repository"
+	"github.com/coreos/go-iptables/iptables"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	containerBridgeLinkName         string        = "br0"
```
nickpetrovic, line 26: I think you might be right about this. We should change this to something more specific to this project like br-beta9, or increment the bridge name until we hit a bridge that doesn't exist. I think the former is fine.

nickpetrovic (approved):

pkg/worker/network.go
luke-lombardi: Done

pkg/common/config.default.yaml
```diff
@@ -114,6 +114,7 @@ worker:
     #     minFreeGpu:
     #     minFreeMemory:
   # global pool attributes
+  exposeLocalNetwork: true
```
nickpetrovic, line 117: I would rename this to something like `useHostResolvConf`. It really just allows you to use the host's DNS setup.

luke-lombardi: updated

PR #452, opened 2024-08-15
Add deployment download route

dleviminzi:
This PR adds a route to the gateway for downloading deployment objects. It uses the stub ID to generate the download path and returns the file at that path. This will allow us to build a download option into the UI so that users can quickly access/verify their deployed code.

### DIFF
--- a/pkg/api/v1/deployment.go
+++ b/pkg/api/v1/deployment.go
@@ -1,15 +1,18 @@
 package apiv1
 
 import (
+	"fmt"
 	"net/http"
+	"path"
+
+	"github.com/labstack/echo/v4"
+	"k8s.io/utils/ptr"
 
 	"github.com/beam-cloud/beta9/pkg/auth"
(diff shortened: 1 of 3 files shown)


pkg/api/v1/deployment.go
```diff
@@ -179,6 +183,25 @@ func (g *DeploymentGroup) ListLatestDeployments(ctx echo.Context) error {
...
+	if err != nil {
+		return HTTPBadRequest("Invalid workspace ID")
+	}
+
+	object, err := g.backendRepo.GetObjectByExternalStubId(ctx.Request().Context(), stubId, workspace.Id)
+	if err != nil {
+		return HTTPInternalServerError("Failed to get object")
+	}
+	path := getPackagePath(workspace.Name, object.ExternalId)
+	fmt.Println(path)
```
nickpetrovic, line 200: Should remove this print.

nickpetrovic (approved):
One small comment. Looks good otherwise.

PR #454, opened 2024-08-16
Add scheduled jobs

nickpetrovic:
Resolve BE-1650


pkg/abstractions/function/function.go
```diff
@@ -354,6 +358,73 @@ func (fs *RunCFunctionService) genContainerId(taskId string) string {
...
+	}
+
+	currentDeployment, err := fs.backendRepo.GetDeploymentByExternalId(ctx, authInfo.Workspace.Id, req.DeploymentId)
+	if err != nil {
+		return &pb.FunctionScheduleResponse{Ok: false, ErrMsg: "Unable to find associated deployment"}, nil
+	}
+
+	if err := fs.backendRepo.DeletePreviousScheduledJob(ctx, &currentDeployment.Deployment); err != nil {
+		return &pb.FunctionScheduleResponse{Ok: false, ErrMsg: "Unable to delete previous scheduled job"}, nil
+	}
```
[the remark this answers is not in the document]
nickpetrovic: Yeah, interesting thought on that. Maybe we can do that in the future? Alternatively, you can define multiple handlers pointing to the same code though.


```python
def main():
    pass

@schedule(when="...")
def handler1():
    main()

@schedule(when="...")
def handler2():
    main()

```

pkg/abstractions/function/function.go
```diff
@@ -354,6 +358,73 @@ func (fs *RunCFunctionService) genContainerId(taskId string) string {
...
+					log.Println("Failed to unmarshal scheduled job payload:", err)
+					continue
+				}
+
+				authInfo := &auth.AuthInfo{Workspace: &types.Workspace{Name: payload.WorkspaceName}}
+				if _, err = fs.invoke(fs.ctx, authInfo, payload.StubId, &payload.TaskPayload); err != nil {
+					log.Printf("Failed to invoke scheduled job: %v\n", err)
+					continue
+				}
+			}
```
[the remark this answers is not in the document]
nickpetrovic: This inserts a task like any other task, so everything should be tracked from that perspective. So I think we could just count the failures of the task for a deployment and its version if we wanted to.

Were you thinking of something else though? Maybe I'm not following.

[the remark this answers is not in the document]
dleviminzi: I was more concerned about metrics to notify us if tasks are not being invoked at all for some reason. Like if we keep failing to read from the channel or a bad schema change causes the unmarshal to fail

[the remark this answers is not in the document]
nickpetrovic: Yeah makes sense. We do this kind of endless loop in a couple more places. I think we could probably come up with a better overall solution e.g. use Sentry or something. But yeah, I think this is a follow up.

pkg/repository/backend_postgres.go
```diff
@@ -1367,3 +1377,134 @@ func (r *PostgresBackendRepository) UpdateSecret(ctx context.Context, workspace
...
 	return &secret, nil
 }
+
+func (r *PostgresBackendRepository) CreateScheduledJob(ctx context.Context, scheduledJob *types.ScheduledJob) (*types.ScheduledJob, error) {
+	payloadJSON, err := json.Marshal(scheduledJob.Payload)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal scheduled job payload: %v", err)
+	}
+
+	// Remove invalid fields from schedule e.g. any field after 5th part
```
luke-lombardi, line 1387: can we add an example in the comment of what this looks like? not sure I follow

pkg/repository/backend_postgres.go
```diff
@@ -1367,3 +1377,134 @@ func (r *PostgresBackendRepository) UpdateSecret(ctx context.Context, workspace
...
+		return fmt.Errorf("failed to unschedule job: %v", err)
+	}
+
+	return nil
+}
+
+func (r *PostgresBackendRepository) unscheduleCron(ctx context.Context, jobId uint64) error {
+	unscheduleQuery := `SELECT cron.unschedule($1::integer);`
+	if _, err := r.client.ExecContext(ctx, unscheduleQuery, jobId); err != nil {
+		return fmt.Errorf("failed to call cron.unschedule function: %v", err)
```
luke-lombardi, line 1447: can we just return the err directly?

pkg/repository/backend_postgres.go
```diff
@@ -1367,3 +1377,134 @@ func (r *PostgresBackendRepository) UpdateSecret(ctx context.Context, workspace
...
+		}
+	}
+
+	return nil
+}
+
+func (r *PostgresBackendRepository) ListenToChannel(ctx context.Context, channel string) (<-chan string, error) {
+	dsn := GenerateDSN(r.config)
+
+	listener := pq.NewListener(dsn, 1*time.Second, 10*time.Second, func(ev pq.ListenerEventType, err error) {
```
luke-lombardi, line 1468: if these are constants we should move them up into `const` vars

pkg/repository/backend_postgres.go
```diff
@@ -1367,3 +1377,134 @@ func (r *PostgresBackendRepository) UpdateSecret(ctx context.Context, workspace
...
+		if err != nil {
+			log.Println("failed to create new listener:", err)
+		}
+	})
+
+	if err := listener.Listen(channel); err != nil {
+		return nil, fmt.Errorf("failed to listen to channel <%s>: %v", channel, err)
```
luke-lombardi, line 1475: same comment about just returning the err

pkg/repository/backend_postgres.go
```diff
@@ -1367,3 +1377,134 @@ func (r *PostgresBackendRepository) UpdateSecret(ctx context.Context, workspace
...
+}
+
+func (r *PostgresBackendRepository) DeleteScheduledJob(ctx context.Context, scheduledJob *types.ScheduledJob) error {
+	updateQuery := `
+	UPDATE scheduled_job
+	SET deleted_at = CURRENT_TIMESTAMP
+	WHERE id = $1;
+	`
+	if _, err := r.client.ExecContext(ctx, updateQuery, scheduledJob.Id); err != nil {
+		return fmt.Errorf("failed to delete scheduled job: %v", err)
```
luke-lombardi, line 1434: also can just return the err

pkg/repository/backend_postgres.go
```diff
@@ -1367,3 +1377,134 @@ func (r *PostgresBackendRepository) UpdateSecret(ctx context.Context, workspace
...
+			listener.Close()
+		}()
+
+		for {
+			select {
+			case n := <-listener.Notify:
+				if n != nil {
+					ch <- n.Extra
+				}
+			case <-time.After(10 * time.Second):
```
luke-lombardi, line 1493: move to const

sdk/src/beta9/abstractions/function.py
```diff
@@ -79,7 +87,7 @@ def __init__(
         image: Image = Image(),
         timeout: int = 3600,
         retries: int = 3,
-        callback_url: Optional[str] = "",
+        callback_url: str = "",
```
luke-lombardi, line 90: why remove the Optional type here?

sdk/src/beta9/abstractions/function.py
```diff
@@ -195,8 +204,148 @@ def _threaded_map(self, inputs: Sequence) -> Iterator[Any]:
...
+                    when=when_utc,
+                    deployment_id=self.deployment_id,
+                )
+            )
+            if not res.ok:
+                terminal.error(res.err_msg, exit=False)
+                return False
+        return deployed
+
+    def convert_schedule_to_utc(self, expression: str) -> str:
```
luke-lombardi, line 243: would probably be worth adding a test for this function

luke-lombardi (changes requested):
a few small comments but looking good 👍

sdk/src/beta9/abstractions/function.py
nickpetrovic: Hm, yeah upstream it defaults to None. When I see Optional but the default is "", it tells me something we got the Optional wrong. I'll change this to default to None.

pkg/repository/backend_postgres.go
nickpetrovic: Updated.

pkg/repository/backend_postgres.go
nickpetrovic: Updated.

pkg/repository/backend_postgres.go
nickpetrovic: Updated.

pkg/repository/backend_postgres.go
nickpetrovic: Updated.

pkg/repository/backend_postgres.go
nickpetrovic: Updated.

pkg/repository/backend_postgres.go
nickpetrovic: Updated.

sdk/src/beta9/abstractions/function.py
nickpetrovic: Removed the UTC conversion because if we convert the user's timezone to UTC, we're not/can't take into account whether its daylight savings or not. Its best to keep this as just UTC for everyone.

luke-lombardi (approved):

pkg/abstractions/function/function.go
```diff
@@ -397,6 +397,13 @@ func (fs *RunCFunctionService) FunctionSchedule(ctx context.Context, req *pb.Fun
 }
 
 func (fs *RunCFunctionService) listenForScheduledJobs() {
+	lock := common.NewRedisLock(fs.rdb)
+	if err := lock.Acquire(fs.ctx, functionScheduledJobsLock, common.RedisLockOptions{TtlS: 10, Retries: 3}); err != nil {
```
luke-lombardi, line 401: I think the lock should be on the message handling itself, not the listening portion

luke-lombardi (changes requested):

pkg/abstractions/function/function.go
nickpetrovic: Updated.

PR #464, opened 2024-08-20
Restrict stub id retrieval with workspace

jsun-m:


pkg/repository/backend_postgres.go
```diff
@@ -773,6 +773,30 @@ func (r *PostgresBackendRepository) GetStubByExternalId(ctx context.Context, ext
 	return &stub, nil
 }
 
+func (r *PostgresBackendRepository) GetStubByExternalIdAndWorkspaceId(ctx context.Context, externalId string, workspaceId uint) (*types.StubWithRelated, error) {
```
luke-lombardi, line 776: Can we add like an optional filter to the existing method or something? it feels like we keep adding new methods for just a slightly different query.

luke-lombardi (changes requested):

pkg/repository/backend_postgres.go
```diff
@@ -763,10 +763,23 @@ func (r *PostgresBackendRepository) GetStubByExternalId(ctx context.Context, ext
 	FROM stub s
 	JOIN workspace w ON s.workspace_id = w.id
 	LEFT JOIN object o ON s.object_id = o.id
-	WHERE s.external_id = $1;
+	WHERE s.external_id = $1
 	`
-	err := r.client.GetContext(ctx, &stub, query, externalId)
+
+	var stubFilters types.StubFilter
+	types.ParseFilterFromQueryFilters(&stubFilters, queryFilters...)
```
luke-lombardi, line 770: Should this be called `ParseConditionFromQueryFilters`

pkg/repository/backend_postgres.go
```diff
@@ -763,10 +763,23 @@ func (r *PostgresBackendRepository) GetStubByExternalId(ctx context.Context, ext
...
+
+	params := []interface{}{externalId}
+	if stubFilters.WorkspaceID != "" {
+		query += " AND w.external_id = $2"
```
luke-lombardi, line 774: this is a little confusing

luke-lombardi (changes requested):
mostly looks good, a couple comments

luke-lombardi (approved):

PR #465, opened 2024-08-20
gpu count endpoint

dleviminzi:
This PR adds an endpoint that returns the platform's GPU count.

### DIFF
--- a/pkg/api/v1/machine.go
+++ b/pkg/api/v1/machine.go
@@ -5,13 +5,14 @@ import (
 	"net/http"
 	"strconv"
 
+	"github.com/labstack/echo/v4"
+
 	"github.com/beam-cloud/beta9/pkg/auth"
 	"github.com/beam-cloud/beta9/pkg/network"
 	"github.com/beam-cloud/beta9/pkg/providers"
 	"github.com/beam-cloud/beta9/pkg/repository"
 	"github.com/beam-cloud/beta9/pkg/scheduler"
 	"github.com/beam-cloud/beta9/pkg/types"
-	"github.com/labstack/echo/v4"
 )
 
 type MachineGroup struct {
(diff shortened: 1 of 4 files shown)


pkg/repository/base.go
```diff
@@ -143,6 +143,7 @@ type ProviderRepository interface {
 	SetMachineLock(providerName, poolName, machineId string) error
 	RemoveMachineLock(providerName, poolName, machineId string) error
 	GetGPUAvailability(pools map[string]types.WorkerPoolConfig) (map[string]bool, error)
+	GetGPUCounts(pools map[string]types.WorkerPoolConfig) (map[string]int, error)
```
luke-lombardi, line 146: looking at the function above this, do we really need a separate repository function for this?

can't we use the results from GetGPUAvailability (maybe with something else added), and use that in the new endpoint?

luke-lombardi (changes requested):

pkg/repository/base.go
dleviminzi: `GetGPUAvailability` is just returning a map of GPU to boolean values and does not really count. If we just want to say there is no availability at all that could work, but it doesn't give us much flexibility to warn about low availability.

pkg/api/v1/machine.go
```diff
@@ -28,6 +29,7 @@ func NewMachineGroup(g *echo.Group, providerRepo repository.ProviderRepository,
 		config:       config,
 	}
 
+	g.GET("/:workspaceId/gpus", auth.WithWorkspaceAuth(group.GPUCounts))
```
luke-lombardi, line 32: not sure if you mentioned this, but why do we need workspace Id here?

luke-lombardi (changes requested):
one last comment on this one

pkg/api/v1/machine.go
dleviminzi: > not sure if you mentioned this, but why do we need workspace Id here?

If I remember correctly, there was an issue with workspace auth if it was not included.

luke-lombardi (approved):

PR #475, opened 2024-08-21
mount beam & beta9 to site-packages as well

dleviminzi:
This PR adds additional mounts to the base runc config so that beam and beta9 are placed in site-packages as well as dist-packages.

### DIFF
--- a/pkg/worker/base_runc_config.json
+++ b/pkg/worker/base_runc_config.json
@@ -302,6 +302,126 @@
 				"nosuid",
 				"nodev"
 			]
+		},
+		{
+			"destination": "/usr/local/lib/python3.8/site-packages/beam",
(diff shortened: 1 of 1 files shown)


nickpetrovic:
@dleviminzi This makes sense. Couple of questions though. What specific use case/issue does this solve and have you tested with custom images e.g. pulling something from ECR?

dleviminzi:
> @dleviminzi This makes sense. Couple of questions though. What specific use case/issue does this solve and have you tested with custom images e.g. pulling something from ECR?

This fixes issues with base python images. They only look at `site-packages`. I suspect some other images might behave similarly.

nickpetrovic (approved):

luke-lombardi (approved):

PR #476, opened 2024-08-21
Add expired status to task and expose task in-queue TTL to sdk

jsun-m:
Task dispatcher already automatically removes task from the list of pending tasks if it is pending up for 2 hours. However, the `STATUS` shown is the exact same status as when a `RUNNING` task times out (`TIMEOUT`). This results in a state that is confusing to the user as well developers trying to debug issues related to this task activity.


dleviminzi (approved):

mernit:
Nice, this is going to be helpful

pkg/abstractions/function/function.go
```diff
@@ -123,8 +123,23 @@ func (fs *RunCFunctionService) FunctionInvoke(in *pb.FunctionInvokeRequest, stre
 }
 
 func (fs *RunCFunctionService) invoke(ctx context.Context, authInfo *auth.AuthInfo, stubId string, payload *types.TaskPayload) (types.TaskInterface, error) {
-	policy := types.DefaultTaskPolicy
-	policy.Expires = time.Now().Add(time.Duration(functionDefaultTaskExpiration) * time.Second)
+	stub, err := fs.backendRepo.GetStubByExternalId(ctx, stubId)
```
luke-lombardi, line 126: Kind of a shame we have to query for this here on every function invocation -- aren't we also call this a level up?

```func (fs *RunCFunctionService) FunctionInvoke(in *pb.FunctionInvokeRequest, stream pb.FunctionService_FunctionInvokeServer) error {
	authInfo, _ := auth.AuthInfoFromContext(stream.Context())
	ctx := stream.Context()

	task, err := fs.invoke(ctx, authInfo, in.StubId, &types.TaskPayload{
		Args: []interface{}{in.Args},
	})
	if err != nil {
		return err
	}

	go func() {
		stub, err := fs.backendRepo.GetStubByExternalId(ctx, in.StubId)
		if err != nil {
			log.Printf("error getting stub: %v", err)
			return
		}

		go fs.eventRepo.PushRunStubEvent(authInfo.Workspace.ExternalId, &stub.Stub)
	}()

	return fs.stream(ctx, stream, authInfo, task)
}```

File: pkg/gateway/services/stub.go [hunk 2]
```diff
@@ -32,21 +33,35 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
 		autoscaler.TasksPerContainer = uint(in.Autoscaler.TasksPerContainer)
 	}
 
+	taskPolicy := types.TaskPolicy{
+		MaxRetries: uint(in.Retries),
+		Timeout:    int(in.Timeout),
+	}
+
+	if in.TaskPolicy.Expiration > 0 {
```
luke-lombardi, line 41: Can you clean up this casting? Feels like a lot of nested stuff

pkg/gateway/services/stub.go
```diff
@@ -32,21 +33,35 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
...
+		taskPolicy.Expiration = uint32(math.Min(float64(in.TaskPolicy.Expiration), float64(types.MaxTaskExpirationS)))
+	}
+
+	if in.TaskPolicy.MaxRetries > 0 {
```
luke-lombardi, line 45: same comment on these two

pkg/types/task.go
```diff
@@ -108,10 +108,14 @@ var DefaultTaskPolicy = TaskPolicy{
 	Timeout:    3600,
 }
 
+var MaxTaskExpirationS = 24 * 60 * 60
+var MaxTaskRetries = 5
+
 type TaskPolicy struct {
 	MaxRetries uint      `json:"max_retries" redis:"max_retries"`
 	Timeout    int       `json:"timeout" redis:"timeout"`
 	Expires    time.Time `json:"expires" redis:"expires"`
```
luke-lombardi, line 117: I think it's confusing that we're adding an additional field here on task policy, expires and expiration sound the same to me. I know that they represent different things but on the go side I wouldn't know that they did

luke-lombardi (changes requested):

pkg/gateway/services/stub.go
jsun-m: reorganized

pkg/gateway/services/stub.go
jsun-m: reorganized

pkg/abstractions/endpoint/buffer.go
```diff
@@ -349,9 +353,16 @@ func (rb *RequestBuffer) handleHttpRequest(req request) {
 
 	resp, err := httpClient.Do(httpReq)
 	if err != nil {
-		req.ctx.JSON(http.StatusInternalServerError, map[string]interface{}{
-			"error": "Internal server error",
-		})
+		var netErr net.Error
+		if errors.As(err, &netErr) && netErr.Timeout() {
```
luke-lombardi, line 357: this second condition could be cleaned up

pkg/abstractions/function/function.go
```diff
@@ -114,28 +115,30 @@ func (fs *RunCFunctionService) FunctionInvoke(in *pb.FunctionInvokeRequest, stre
...
+
+	policy := config.TaskPolicy
+	policy.Expires = time.Now().Add(time.Duration(policy.TTL) * time.Second)
 
 	task, err := fs.taskDispatcher.SendAndExecute(ctx, string(types.ExecutorFunction), authInfo, stubId, payload, policy)
 	if err != nil {
 		return nil, err
 	}
 
+	go fs.eventRepo.PushRunStubEvent(authInfo.Workspace.ExternalId, &stub.Stub)
```
luke-lombardi, line 141: 👍

pkg/gateway/services/stub.go
```diff
@@ -146,3 +147,29 @@ func (gws *GatewayService) DeployStub(ctx context.Context, in *pb.DeployStubRequ
...
 }
+
+func (gws *GatewayService) configureTaskPolicy(policy *pb.TaskPolicy, stubType types.StubType) types.TaskPolicy {
+	p := types.TaskPolicy{
+		MaxRetries: uint(math.Min(float64(policy.MaxRetries), float64(types.MaxTaskRetries))),
+		Timeout:    int(policy.Timeout),
+		TTL:        uint32(math.Min(float64(policy.Ttl), float64(types.MaxTaskTTL))),
+	}
+
+	if stubType.Kind() == types.StubTypeEndpoint {
```
luke-lombardi, line 158: can this be a switch?

luke-lombardi (changes requested):
left a couple more comments, getting close though 👍

pkg/abstractions/endpoint/buffer.go
jsun-m: cleaned up

pkg/task/dispatch.go
```diff
@@ -166,7 +166,7 @@ func (d *Dispatcher) monitor(ctx context.Context) {
 					continue
 				}
 
-				if !heartbeat && claimed {
```
jsun-m, line 169: this claimed check is redundant cause if its not claimed it skips

luke-lombardi (approved):
👍

PR #477, opened 2024-08-22
support env var references in commands

dleviminzi:
This change allows users to reference secrets as environment variables in their image build. 

For instance, lets say I create a secret called `TEST_TOKEN` like this: 
```beta9 secret create TEST_TOKEN hello```
and another to specify the numpy version I want:
```
beta9 secret create NUMPY_VERSION 1.22.3
```

We can now access those values as environment variables. 

```
from beta9 import endpoint, Image

@endpoint(
    cpu=1.0,
    memory=128,
    image=Image(
        commands=['echo $TEST_TOKEN'],
        python_packages=['numpy==$NUMPY_VERSION']
    )
)
def multiply(x, y, z):
    result = x * 2
    return {"result": result}
```

### DIFF
--- a/pkg/abstractions/image/build.go
+++ b/pkg/abstractions/image/build.go
@@ -35,6 +35,7 @@ type Builder struct {
 	registry      *common.ImageRegistry
 	containerRepo repository.ContainerRepository
 	tailscale     *network.Tailscale
+	backendRepo   repository.BackendRepository
 }
 
 type BuildOpts struct {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by nickpetrovic — APPROVED** [review]

File: pkg/abstractions/image/build.go [hunk 1]
```diff
@@ -140,6 +142,17 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
 		}
 	}
 
+	// Add workspace secrets to the environment
+	secrets, err := b.backendRepo.ListSecretsDecrypted(ctx, authInfo.Workspace)
```
luke-lombardi, line 146: PR looks good, my only concern is that people won't necessarily expect every secret they've defined to be loaded in their image builds. @mernit thoughts?

mernit: Yeah I agree, it feels precarious to just load all the decrypted secrets into the image build. Ideally it should only be loading the values they've asked for.

dleviminzi: We could add a "build secrets" option to allow users to dictate which ones to fetch. In my opinion that extra step doesn't really add value. 

My comment might be misleading but this doesn't load all of the environment variables into the environment as it is executing, just during build. For instance, `tt` will still be null in the code below unless the user adds `    secrets=["TEST_TOKEN"]` to the decorator. 

```
from beta9 import endpoint, Image

@endpoint(
    cpu=1.0,
    memory=128,
    image=Image(
        commands=['echo $TEST_TOKEN'],
        python_packages=['numpy==$NUMPY_VERSION']
    )
)
def multiply(x):
    import os
    tt = os.environ.get("TEST_TOKEN")
    result = x * 2
    return {"result": result, "tt": tt}
```

In the future it might make sense to have secrets that are name-spaced to deployments or something like that. However, as they are at the workspace level at the moment I think its fair to make them generally accessible to images being built in that workspace.

PR #479, opened 2024-08-23
Don't loop if container does not exist when STOP is received

luke-lombardi:


pkg/worker/worker.go
```diff
@@ -422,6 +422,12 @@ func (s *Worker) processStopContainerEvents() {
 		if err != nil {
 			log.Printf("<%s> - unable to stop container: %v\n", event.ContainerId, err)
 
+			if strings.Contains(err.Error(), "container does not exist") {
```
nickpetrovic, line 425: Guessing there isn't an error type or struct for this you could use instead?

nickpetrovic (approved):

PR #481, opened 2024-08-24
Cloud bucket abstraction

dleviminzi:
This PR gives users the ability to mount their own S3 buckets to containers. This is nice because it allows users to  choose their own storage providers. 

Tested with:
- [x] AWS S3
- [x] Tigris
- [x] Cloudflare R2


pkg/abstractions/common/mount.go
```diff
@@ -29,5 +29,20 @@ func ConfigureContainerRequestMounts(stubObjectId string, workspaceName string,
 		})
 	}
 
+	for _, s := range config.CloudBuckets {
```
luke-lombardi, line 32: let's try to consolidate this into volumes

pkg/gateway/gateway.proto
```diff
@@ -199,6 +202,19 @@ message Volume {
   string mount_path = 2;
 }
 
+message CloudBucket {
```
luke-lombardi, line 205: shouldn't need an additional message if we consolidate into volumes

pkg/storage/mountpoint.go
```diff
@@ -14,7 +13,7 @@ type MountPointStorage struct {
 	config   types.MountPointConfig
 }
 
-func NewMountPointStorage(config types.MountPointConfig) (Storage, error) {
+func NewMountPointStorage(config types.MountPointConfig) (*MountPointStorage, error) {
```
luke-lombardi, line 16: this should satisfy the interface

pkg/types/backend.go
```diff
@@ -175,19 +175,20 @@ type TaskStats struct {
...
+	OnStart         string            `json:"on_start"`
+	PythonVersion   string            `json:"python_version"`
+	KeepWarmSeconds uint              `json:"keep_warm_seconds"`
+	MaxPendingTasks uint              `json:"max_pending_tasks"`
+	CallbackUrl     string            `json:"callback_url"`
+	TaskPolicy      TaskPolicy        `json:"task_policy"`
+	Workers         uint              `json:"workers"`
+	Authorized      bool              `json:"authorized"`
+	Volumes         []*pb.Volume      `json:"volumes"`
+	CloudBuckets    []*pb.CloudBucket `json:"cloud_buckets"`
```
luke-lombardi, line 189: can remove if we consolidate

pkg/types/gateway.go
```diff
@@ -15,6 +15,7 @@ const (
 	FailedContainerThreshold           int           = 1
 	RequestTimeoutDurationS            time.Duration = 175 * time.Second
 	ContainerVolumePath                string        = "/volumes"
+	ContainerCloudBucketPath           string        = "/tmp"
```
luke-lombardi, line 18: should use the same path

pkg/worker/worker.go
```diff
@@ -46,6 +47,7 @@ type Worker struct {
 	workerId                string
 	eventBus                *common.EventBus
 	containerInstances      *common.SafeMap[*ContainerInstance]
+	mountPointPaths             *common.SafeMap[[]string]
```
luke-lombardi, line 50: this shouldn't be here. need to clean this up

luke-lombardi (changes requested):

pkg/storage/mountpoint.go
```diff
@@ -24,29 +23,41 @@ func (s *MountPointStorage) Mount(localPath string) error {
...
+			return err
+		}
+	}
+
+	endpoint := ""
+	if s.config.BucketURL != "" {
+		endpoint = fmt.Sprintf("--endpoint-url=%s", s.config.BucketURL)
+	}
 
+	// FIXME: cache? https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#caching-configuration
```
dleviminzi, line 38: We could give users the option to toggle caching on this mount. This removes read-after-write consistency guarantees, but would likely be useful in heavy read situations. 

We could also expose the option to include the overwrite tag, which allows users to overwrite files in the volume. By default, any attempt to overwrite a file will result in an error.

sdk/src/beta9/abstractions/volume.py
```diff
@@ -56,7 +88,18 @@ def get_or_create(self) -> bool:
...
 
     def export(self):
-        return VolumeConfig(
+        vol = VolumeGateway(
             id=self.volume_id,
             mount_path=self.mount_path,
         )
+
+        # FIXME: is the external flag necessary?
+        if self.config.external:
```
dleviminzi, line 97: We could just check for the presence of config and assume its external when provided.

pkg/common/crypto.go
```diff
@@ -65,6 +65,19 @@ func Decrypt(secretKey []byte, ciphertext64 string) (string, error) {
 	return string(plaintext), nil
 }
 
+// DecryptAll decrypts all the given secrets in the order they are given.
+func DecryptAll(signingKey []byte, secrets []string) ([]string, error) {
```
luke-lombardi, line 69: Let's make this DecryptAllSecrets for clarity

pkg/types/gateway.go
```diff
@@ -7,6 +7,7 @@ import (
 const (
 	DefaultExtractedObjectPath         string        = "/data/unpacked"
 	DefaultVolumesPath                 string        = "/data/volumes"
+	DefaultExternalVolumesPath         string        = "/data/external-volumes"
```
luke-lombardi, line 10: do we need this in gateway? Assuming we aren't allowing users to upload through cli, etc it might not matter?

pkg/types/worker.go
```diff
@@ -25,8 +25,9 @@ type ContainerResourceUsage struct {
...
 type Mount struct {
-	LocalPath string `json:"local_path"`
-	MountPath string `json:"mount_path"`
-	LinkPath  string `json:"link_path"`
-	ReadOnly  bool   `json:"read_only"`
+	LocalPath        string            `json:"local_path"`
+	MountPath        string            `json:"mount_path"`
+	LinkPath         string            `json:"link_path"`
+	ReadOnly         bool              `json:"read_only"`
+	MountPointConfig *MountPointConfig `json:"mountpoint_config"`
```
luke-lombardi, line 32: it may make sense to make this into a pointer to an interface{}, and add an additional Type string, then you can use a switch(mount.type) and case it to this struct. To keep this extensible.

luke-lombardi (changes requested):

pkg/gateway/services/stub.go
```diff
@@ -73,7 +73,41 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
 			Name:  secret.Name,
 			Value: secret.Value,
 		})
+	}
 
+	for i, volume := range in.Volumes {
```
luke-lombardi, line 78: Can we pull this volume stuff out into another method in this file?

pkg/storage/mountpoint.go
```diff
@@ -24,29 +27,33 @@ func (s *MountPointStorage) Mount(localPath string) error {
...
+			return err
+		}
+	}
 
-	s.mountCmd = exec.Command(
-		"mount-s3",
-		s.config.AWSS3Bucket,
-		localPath,
-	)
+	// FIXME: cache? https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#caching-configuration
```
luke-lombardi, line 37: can you clarify exactly what we need to fix? Is it enabling cache?

pkg/worker/worker.go
```diff
@@ -294,6 +297,12 @@ func (s *Worker) RunContainer(request *types.ContainerRequest) error {
 	}
 	log.Printf("<%s> - acquired port: %d\n", containerId, bindPort)
 
+	// Setup external S3 mounts
+	err = s.createMountpoints(request)
```
luke-lombardi, line 301: Let's move the createmountpoints stuff / mounts stuff into a separate file (worker/mount.go) - you can look at ContainerNetworkManager for an example. I think it might be time to encapsulate all container mount logic (worker side) in a separate struct.

luke-lombardi (changes requested):

pkg/types/worker.go
```diff
@@ -13,6 +13,8 @@ const (
 
 const (
 	BuildContainerPrefix string = "build-"
+	MountTypeJuiceFS     string = "juicefs"
```
luke-lombardi, line 16: I think we already have these types defined somewhere (maybe storage.go?)

pkg/abstractions/common/mount.go
```diff
@@ -3,31 +3,64 @@ package abstractions
 import (
 	"path"
 
+	"github.com/beam-cloud/beta9/pkg/common"
+	"github.com/beam-cloud/beta9/pkg/storage"
 	"github.com/beam-cloud/beta9/pkg/types"
 )
 
-func ConfigureContainerRequestMounts(stubObjectId string, workspaceName string, config types.StubConfigV1, stubId string) []types.Mount {
+const defaultExternalVolumesPath string = "/data/external-volumes"
```
luke-lombardi, line 11: Shouldn't this have a different root path? If it's mounted to /data, won't it be a subpath of the juicefs mount?

pkg/worker/mount.go
```diff
@@ -0,0 +1,91 @@
...
+		if err := mountPointS3.Unmount(m); err != nil {
+			log.Printf("<%s> - failed to unmount external s3 bucket: %v\n", containerId, err)
+		}
+	}
+
+	smm.mountPointPaths.Delete(containerId)
+}
+
+// RemoveWorkerStorage removes all mounts for a worker
+func (smm *StorageManager) RemoveWorkerStorage() error {
```
luke-lombardi, line 64: it makes sense to be able to remove all of them in some cases, but we probably don't want to leave a bunch of mountpoints in memory. I would probably argue for removing by container ID

pkg/worker/mount.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"log"
+
+	"github.com/beam-cloud/beta9/pkg/common"
+	"github.com/beam-cloud/beta9/pkg/storage"
+	"github.com/beam-cloud/beta9/pkg/types"
+)
+
+type StorageManager struct {
+	mountPointPaths *common.SafeMap[[]string]
```
luke-lombardi, line 13: should this be a map of container ID to the mount objects themselves? that way you can clear them by container ID. Also then its not mountPointPaths, it's just `containerMounts`

pkg/worker/worker.go
```diff
@@ -875,10 +882,9 @@ func (s *Worker) shutdown() error {
 	}
 
 	s.cancel()
-
-	err := s.storage.Unmount(s.config.Storage.FilesystemPath)
+	err := s.storageManager.RemoveWorkerStorage()
```
luke-lombardi, line 885: What is the benefit of this if we already cleaned up the mounts by container ID?

pkg/worker/mount.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func NewContainerMountManager() *containerMountManager {
+	return &containerMountManager{
+		mountPointPaths: common.NewSafeMap[[]string](),
+	}
+}
+
+// SetupContainerMounts initializes any external storage for a container
+func (smm *containerMountManager) SetupContainerMounts(containerId string, mounts []types.Mount) error {
```
luke-lombardi, line 22: let's rename this `smm` now that the struct has a different name

sdk/src/beta9/abstractions/volume.py
```diff
@@ -1,29 +1,73 @@
...
+            The S3 endpoint for the external provider.
+    """
+
+    external: bool = False
+    access_key: Optional[str] = None
+    secret_key: Optional[str] = None
+    endpoint: Optional[str] = None
+
+
 class Volume(BaseAbstraction):
```
luke-lombardi, line 32: Can we create a CloudBucket(Volume) class that basically just does exactly what you've done here with the config? I'm not saying replace the work you've done, just make it into a new name from a client perspective so it's clearly delineated.

luke-lombardi (approved):

PR #485, opened 2024-08-27
Add CunoFS support

luke-lombardi:
- Add a new storage option (cunofs)
- NOTE: only supported currently on x86

### DIFF
--- a/pkg/storage/cunofs.go
+++ b/pkg/storage/cunofs.go
@@ -28,13 +28,13 @@ func NewCunoFsStorage(config types.CunoFSConfig) (Storage, error) {
 
 func (s *CunoFsStorage) Mount(localPath string) error {
 	log.Printf("CunoFS filesystem mounting to: '%s'\n", localPath)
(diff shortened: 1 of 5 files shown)


pkg/storage/cunofs.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (s *CunoFsStorage) Format(fsName string) error {
+	activateCmd := exec.Command("sh", "-c", fmt.Sprintf("yes | echo %s | cuno creds activate", s.config.LicenseKey))
+	_, err := activateCmd.CombinedOutput()
+	if err != nil {
+		return err
+	}
+
+	templateStr := `aws_access_key_id={{.S3AccessKey}}
```
nickpetrovic, line 83: This seems required. Is cuno able to use the injected aws credentials via the service account?

nickpetrovic: Hm, looks like we'll use tigris for this anyway. Guess we can follow up later with using aws iam injected secrets.

PR #487, opened 2024-08-27
Check gpu limit before creating stub

dleviminzi:
This PR adds a check to `GetOrCreateStub` that ensures that the user has a valid GPU limit if they've requested to create a stub that includes a GPU.

### DIFF
--- a/pkg/gateway/services/stub.go
+++ b/pkg/gateway/services/stub.go
@@ -23,14 +23,7 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
 
 	if in.Gpu != "" {
 		concurrencyLimit, err := gws.backendRepo.GetConcurrencyLimitByWorkspaceId(ctx, authInfo.Workspace.ExternalId)
-		if err != nil {
-			return &pb.GetOrCreateStubResponse{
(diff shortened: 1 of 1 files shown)


pkg/gateway/services/stub.go
```diff
@@ -21,6 +21,23 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
...
+			return &pb.GetOrCreateStubResponse{
+				Ok:     false,
+				ErrMsg: "Failed to get concurrency limit for workspace",
+			}, nil
+		}
+
+		if concurrencyLimit.GPULimit <= 0 {
+			return &pb.GetOrCreateStubResponse{
+				Ok:     false,
+				ErrMsg: "GPU concurrency limit is 0. A payment method is required to deploy with GPUs.",
```
luke-lombardi, line 36: Since this in beta9, and there's no concept of "payment", I would just let the error be the first part of that message.

luke-lombardi (changes requested):
just one comment and good to merge

luke-lombardi (approved):

jsun-m (approved):

luke-lombardi (approved):

PR #489, opened 2024-08-27
Endpoints should at least transition from pending to expires

jsun-m:
One special condition we need to consider.

Current behavior:
If a request to endpoint is made, and it starts running on a container -> even if you disconnect the client request the task will still run.
However, if a request to endpoint is made but then client disconnects before the task goes inflight, the task will never run because the request's context is canceled before it ever reached he gunicorn server. Therefore a task like this is stuck in PENDING forever (which this PR helps transitions such tasks to expired)

This behavior of the request state is not anything new and is the current behavior in prod.

### DIFF
--- a/pkg/abstractions/function/function.go
+++ b/pkg/abstractions/function/function.go
@@ -29,7 +29,7 @@ type FunctionService interface {
 }
 
 const (
-	FunctionDefaultTaskTTL int = 3600 * 12 // 12 hours
+	DefaultFunctionDefaultTaskTTL int = 3600 * 12 // 12 hours
 
 	functionRoutePrefix             string        = "/function"
 	scheduleRoutePrefix             string        = "/schedule"
--- a/pkg/abstractions/function/function.go
+++ b/pkg/abstractions/function/function.go
@@ -29,7 +29,7 @@ type FunctionService interface {
 }
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/abstractions/function/function.go
```diff
@@ -29,7 +29,7 @@ type FunctionService interface {
 }
 
 const (
-	FunctionDefaultTaskTTL int = 3600 * 12 // 12 hours
+	DefaultFunctionDefaultTaskTTL int = 3600 * 12 // 12 hours
```
luke-lombardi, line 32: can we fix this name? Maybe DefaultFunctionTaskTTL

pkg/abstractions/taskqueue/taskqueue.go
```diff
@@ -46,7 +46,7 @@ type TaskQueueServiceOpts struct {
 }
 
 const (
-	TaskQueueDefaultTaskTTL int = 3600 * 2 // 2 hours
+	DefaultTaskQueueDefaultTaskTTL int = 3600 * 2 // 2 hours
```
luke-lombardi, line 49: same thing as above comment

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #490, opened 2024-08-28
log when mountpoint fails and cleanup mountpoint dir 

dleviminzi:
This PR adds a container log when a mountpoint setup fails. This way users will know if the credentials they provided were invalid. It also removes the temporary directory from the worker in the case of failure and during clean up.

### DIFF
--- a/pkg/storage/mountpoint.go
+++ b/pkg/storage/mountpoint.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


pkg/storage/mountpoint.go
```diff
@@ -46,7 +47,9 @@ func (s *MountPointStorage) Mount(localPath string) error {
 
 	output, err := s.mountCmd.CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("error: %+v, %s", err, string(output))
+		// Cleanup the temporary mountpoint directory if the mount fails
+		errors.Join(err, os.RemoveAll(localPath))
+		return fmt.Errorf("%+v, %s", err, string(output))
```
nickpetrovic, line 52: What happens if os.RemoveAll doesn't return an error? 

I kind of think these should be separated. I think its more clear this way.

```go
removeErr := os.RemoveAll()
if removeErr != nil {
  err = errors.Join(err, removeError)
}

return err
```

dleviminzi: I changed to just ignore the `os.RemoveAll` per @luke-lombardi's recommendation

luke-lombardi (approved):

PR #504, opened 2024-09-04
Allow using defaults when prompting for config values

nickpetrovic:
- Allow using gateway host and port defaults when prompting for config
- Require token when calling config create command (during prompt, not for cli command)
- Print message when user input prompts are invalid

Resolve BE-1752

### DIFF
--- a/sdk/src/beta9/config.py
+++ b/sdk/src/beta9/config.py
@@ -43,6 +43,7 @@ class SDKSettings:
     api_host: str = DEFAULT_API_HOST
     config_path: Path = Path("~/.beta9/config.ini").expanduser()
     ascii_logo: str = DEFAULT_ASCII_LOGO
+    use_defaults_in_prompt: bool = False
 
     def __post_init__(self, **kwargs):
         if p := os.getenv("CONFIG_PATH"):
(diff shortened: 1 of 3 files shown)


sdk/src/beta9/config.py
```diff
@@ -174,15 +175,25 @@ def prompt_for_config_context(
...
+            while not (gateway_host := prompt_gateway_host()) or not validate_ip_or_dns(
+                gateway_host
+            ):
+                terminal.warn("Gateway host is invalid.")
+
+            while not (gateway_port := prompt_gateway_port()) or not validate_port(gateway_port):
+                terminal.warn("Gateway port is invalid.")
+
+        if require_token:
+            while not (token := terminal.prompt(text="Token", default=None)) or len(token) < 64:
```
nickpetrovic, line 193: 64 is completely arbitrary. Is there a length that's guaranteed or should I keep it at less than what it could be (at least what I see in my account).

luke-lombardi (approved):

sdk/src/beta9/config.py
luke-lombardi: I think this should be 128

PR #508, opened 2024-09-04
Jupyter remote functions

dleviminzi:
This PR adds support for calling remote functions inside of jupyter notebooks. 

Screenshot 2024-09-04 at 17 26 59

### DIFF
--- a/sdk/src/beta9/abstractions/base/runner.py
+++ b/sdk/src/beta9/abstractions/base/runner.py
@@ -1,6 +1,8 @@
 import inspect
+import json
 import os
 import time
+import uuid
 from queue import Empty, Queue
 from typing import Callable, List, Optional, Union
 
@@ -197,7 +199,10 @@ class RunnerAbstraction(BaseAbstraction):
             return
 
         module = inspect.getmodule(func)  # Determine module / function name
-        if module:
+        if in_jupyter():
+            module_file = gather_executed_cells()
+            module_name = module_file.split("/")[-1].rstrip(".py")
+        elif module:
             module_file = os.path.relpath(module.__file__, start=os.getcwd()).replace("/", ".")
             module_name = os.path.splitext(module_file)[0]
         else:
@@ -300,6 +305,10 @@ class RunnerAbstraction(BaseAbstraction):
             if sync_result.success:
                 self.files_synced = True
                 self.object_id = sync_result.object_id
+
+                for file in os.listdir("./"):
+                    if file.startswith("tmp_beta9_"):
+                        os.remove(file)
             else:
                 terminal.error("File sync failed", exit=False)
                 return False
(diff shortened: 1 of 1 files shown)


sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -369,3 +378,40 @@ def prepare_runtime(
...
         return True
+
+
+def in_jupyter() -> bool:
+    try:
+        from IPython import get_ipython
+
+        shell = get_ipython().__class__.__name__
+        return shell == "ZMQInteractiveShell"
+    except NameError:
```
luke-lombardi, line 389: should probably also handle an ImportError here

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -369,3 +378,40 @@ def prepare_runtime(
...
+
+        with open(notebook_path, "r", encoding="utf-8") as f:
+            notebook_json = json.load(f)
+
+        executed_cells = []
+        for cell in notebook_json.get("cells", []):
+            if cell.get("cell_type") == "code" and cell.get("execution_count") is not None:
+                executed_cells.append(cell.get("source", ""))
+
+        tmp_file_path = f"./tmp_beta9_{uuid.uuid4()}.py"
```
luke-lombardi, line 408: let's move this prefix to a const since it's used in two places

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -369,3 +378,40 @@ def prepare_runtime(
...
+        with open(tmp_file_path, "w") as f:
+            for code in executed_cells:
+                if isinstance(code, list):
+                    code = "".join(code)
+                f.write(code + "\n\n")
+
+        return tmp_file_path
+    except NameError:
```
luke-lombardi, line 416: same thing here, we should handle the ImportError

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -300,6 +305,10 @@ def prepare_runtime(
             if sync_result.success:
                 self.files_synced = True
                 self.object_id = sync_result.object_id
+
+                for file in os.listdir("./"):
+                    if file.startswith("tmp_beta9_"):
```
luke-lombardi, line 310: move to const

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -197,7 +199,10 @@ def _map_callable_to_attr(self, *, attr: str, func: Callable):
             return
 
         module = inspect.getmodule(func)  # Determine module / function name
-        if module:
+        if in_jupyter():
```
luke-lombardi, line 202: can we make the jupyter case the non-default. I.e. move this to the else condition

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -300,6 +305,10 @@ def prepare_runtime(
             if sync_result.success:
                 self.files_synced = True
                 self.object_id = sync_result.object_id
+
+                for file in os.listdir("./"):
```
luke-lombardi, line 309: shouldn't this clean up everything regardless of if the sync was successful? Also do we really need to iterate through the entire directory?

luke-lombardi: I think there's another way to do this using tempfile: https://docs.python.org/3/library/tempfile.html

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -296,6 +306,7 @@ def prepare_runtime(
 
         if not self.files_synced:
             sync_result = self.syncer.sync()
+            self.remove_tmp_files()
```
luke-lombardi, line 309: Shouldn't this be called if it syncs _or_ fails to sync, regardless?

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/base/runner.py
dleviminzi: I'm calling it before checking whether the sync succeeded or failed. That way in any case it will be removed.

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -197,15 +199,23 @@ def _map_callable_to_attr(self, *, attr: str, func: Callable):
...
+            tmp_file = create_tmp_jupyter_file(module._ih)
+            module_name = tmp_file.name.split("/")[-1].removesuffix(".py")
+            self.tmp_files.append(tmp_file)
         else:
             module_name = "__main__"
 
         function_name = func.__name__
         setattr(self, attr, f"{module_name}:{function_name}")
 
+    def remove_tmp_files(self) -> None:
```
luke-lombardi, line 215: since this isn't called outside of the class, I would prefix it with _ (e.g. `_remove_tmp_files()`)

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #512, opened 2024-09-05
Add asgi websockets

jsun-m:


pkg/abstractions/endpoint/buffer.go
```diff
@@ -301,6 +306,31 @@ func (rb *RequestBuffer) getHttpClient(address string) (*http.Client, error) {
...
+	rb.availableContainersLock.RUnlock()
+
+	err := rb.incrementRequestsInFlight(c.id)
+	if err != nil {
+		return
+	}
+
+	err = proxyWebsocketConnection(req.ctx.Response().Writer, req.ctx.Request(), fmt.Sprintf("ws://%s/%s", c.address, req.ctx.Param("subPath")))
+	if err != nil {
+		req.done <- true
```
luke-lombardi, line 329: Shouldn't we set this to be done regardless of if there is an error?

luke-lombardi (changes requested):

nickpetrovic (approved):

PR #514, opened 2024-09-05
Stop previous deployment on new scheduled jobs deployment

nickpetrovic:
Resolve BE-1763

### DIFF
--- a/pkg/gateway/services/stub.go
+++ b/pkg/gateway/services/stub.go
@@ -161,6 +161,14 @@ func (gws *GatewayService) DeployStub(ctx context.Context, in *pb.DeployStubRequ
 
 	version := uint(1)
 	if lastestDeployment != nil {
+		if stub.Type == types.StubType(types.StubTypeScheduledJobDeployment) {
+			if err := gws.stopDeployments([]types.DeploymentWithRelated{*lastestDeployment}, ctx); err != nil {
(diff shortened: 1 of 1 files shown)


luke-lombardi (approved):

pkg/gateway/services/stub.go
```diff
@@ -161,6 +161,14 @@ func (gws *GatewayService) DeployStub(ctx context.Context, in *pb.DeployStubRequ
 
 	version := uint(1)
 	if lastestDeployment != nil {
+		if stub.Type == types.StubType(types.StubTypeScheduledJobDeployment) {
+			if err := gws.stopDeployments([]types.DeploymentWithRelated{*lastestDeployment}, ctx); err != nil {
```
jsun-m, line 165: Should this be after `deployment, err := gws.backendRepo.CreateDeployment` in case that fails?

nickpetrovic: What if CreateDeployment succeeds and stopDeployments fails? Someone could keep trying and create more Deployments.

nickpetrovic: `stopDeployments` also only returns an error if backendRepo.UpdateDeployment() fails. Errors are ignored for all other calls in that function.

I think this should be fine.

PR #515, opened 2024-09-05
Cache authorization token

jsun-m:


pkg/auth/grpc.go
```diff
@@ -50,9 +52,21 @@ func (ai *AuthInterceptor) validateToken(md metadata.MD) (*AuthInfo, bool) {
...
 	tokenKey := strings.TrimPrefix(md["authorization"][0], "Bearer ")
-	token, workspace, err := ai.backendRepo.AuthorizeToken(context.TODO(), tokenKey)
-	if err != nil {
-		return nil, false
+
+	var token *types.Token
+	var workspace *types.Workspace
+
+	token, workspace, err := ai.workspaceRepo.AuthorizeToken(tokenKey)
+	if err != nil || token == nil || workspace == nil {
```
nickpetrovic, line 60: I think it would be simpler to just return an error from AuthorizeToken even if the contents of the key is "".

Whats the benefit of checking each of these?

jsun-m: done

pkg/common/keys.go
```diff
@@ -236,6 +238,14 @@ func (rk *redisKeys) WorkspaceVolumePathDownloadToken(token string) string {
 	return fmt.Sprintf(workspaceVolumePathDownloadToken, token)
 }
 
+func (rl *redisKeys) WorkspaceAuthorizedToken(token string) string {
+	return fmt.Sprintf(workspaceAuthorizedToken, token)
+}
+
+func (rl *redisKeys) WorkspaceAuthorizedWorkspace(workspaceId string) string {
```
nickpetrovic, line 245: Where's this being used?

pkg/repository/workspace_redis.go
```diff
@@ -51,3 +55,46 @@ func (wr *WorkspaceRedisRepository) SetConcurrencyLimitByWorkspaceId(workspaceId
...
+	info := &AuthInfo{}
+	err = json.Unmarshal([]byte(res), &info)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return info.Token, info.Workspace, nil
+}
+
+func (wr *WorkspaceRedisRepository) CacheAuthorizationToken(token *types.Token, workspace *types.Workspace) error {
```
nickpetrovic, line 84: A nitpick, but can we say `Set` instead of `Cache` to keep with the naming convention of the other redis repo related functions?

pkg/repository/workspace_redis.go
```diff
@@ -16,6 +18,8 @@ func NewWorkspaceRedisRepository(r *common.RedisClient) WorkspaceRepository {
 	return &WorkspaceRedisRepository{rdb: r}
 }
 
+var CACHED_TOKEN_TTL_S = 600 // 10 minutes
```
nickpetrovic, line 21: Another nitpick, can this be `cachedTokenTTlS` instead? In addition can it be a `const`?

jsun-m: oh yea. I think i mixed up some conventions

pkg/common/keys.go
jsun-m: removed. I was testing a few things with hset that didnt work out

pkg/auth/http.go
```diff
@@ -25,9 +25,20 @@ func AuthMiddleware(backendRepo repository.BackendRepository) echo.MiddlewareFun
...
-				return echo.NewHTTPError(http.StatusUnauthorized)
+				token, workspace, err = backendRepo.AuthorizeToken(c.Request().Context(), tokenKey)
+				if err != nil {
+					return echo.NewHTTPError(http.StatusUnauthorized)
+				}
+
+				err = workspaceRepo.SetAuthorizationToken(token, workspace)
+				if err != nil {
+					return echo.NewHTTPError(http.StatusInternalServerError)
+				}
```
nickpetrovic, line 41: I think there might be an opportunity to streamline this logic. 

Could we turn this into a function and pass the function into AuthMiddleware and NewAuthInterceptor? This way we have one place for this logic and neither one of these need repos.

jsun-m: wouldn't the function need access to the repo still? So we will have to initialize the function in gateway and passing it down from Gateway -> abstraction -> authMiddleware

nickpetrovic: Its the same as passing workspaceRepo down, but this would dedup the auth logic. 

Going to approve this since it'll works as-is, like before.

nickpetrovic (approved):

pkg/auth/http.go
jsun-m: Thanks. I feel like this makes more sense because it's consistent with the scheme of passing down repos as necessary. 

Also WorkspaceRepo might be used in abstractions in the future since it is a general user-specific cache

PR #520, opened 2024-09-09
don't check count for T4

dleviminzi:


pkg/gateway/services/stub.go
```diff
@@ -43,7 +43,7 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
 			}, nil
 		}
 
-		if gpuCounts[in.Gpu] <= 1 {
+		if gpuCounts[in.Gpu] <= 1 && in.Gpu != types.GPU_T4.String() {
```
mernit, line 46: Add a comment explaining why this is necessary

PR #522, opened 2024-09-09
Allow running image build commands in order

nickpetrovic:
Adds new `add_commands()` and `add_python_packages()` to the `Image` abstraction.

This allows for running commands or installing python packages in order, but only after the initial values of the Image's attributes `commands` and `python_packages`. The reason for the exception is to maintain backwards compatibility with existing stubs/deployments.

Resolve BE-1675


pkg/abstractions/image/build.go
```diff
@@ -37,6 +40,11 @@ type Builder struct {
 	tailscale     *network.Tailscale
 }
 
+type ContainerCommand struct {
```
luke-lombardi, line 43: If we end up supporting things like `with_cuda_version` or something similar (things that aren't necessarily a container command), is this the right naming? I was thinking maybe something along the lines of `ContainerBuildStep` (or something like that)

Users would need to know either way, but for internal clarity

luke-lombardi (changes requested):
this looks good, just had a concern around naming

pkg/abstractions/image/build.go
nickpetrovic: Yeah good point. I'll change it.

nickpetrovic: Updated to `BuildStep`.

luke-lombardi (approved):

PR #525, opened 2024-09-10
Patch websockets with tailscale

jsun-m:


pkg/abstractions/endpoint/buffer.go
```diff
@@ -326,7 +324,11 @@ func (rb *RequestBuffer) handleWSRequest(req request) {
 
 	defer rb.afterRequest(req, c.id)
 
-	err = proxyWebsocketConnection(req.ctx.Response().Writer, req.ctx.Request(), fmt.Sprintf("ws://%s/%s", c.address, req.ctx.Param("subPath")))
+	dstDialer := websocket.Dialer{
+		NetDialContext: network.GetDialer(rb.tailscale, rb.tsConfig),
```
luke-lombardi, line 328: this should take into account if the address is a tailnet address. There's an example of this somewhere (otherwise you'd be using the tailnet dialer even in eks)

pkg/abstractions/endpoint/buffer.go
```diff
@@ -314,8 +314,6 @@ func (rb *RequestBuffer) handleWSRequest(req request) {
 		return
 	}
 
-	// Select an available container to forward the request to (whichever one has the lowest # of inflight requests)
```
luke-lombardi, line 317: Why'd you remove this comment?

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #526, opened 2024-09-12
skip command lines in jupyter cells

dleviminzi:


sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -399,7 +399,13 @@ def create_tmp_jupyter_file(input_history: List[str]) -> tempfile.NamedTemporary
     for code in input_history:
         if isinstance(code, list):
             code = "".join(code)
-        tmp_file.write(code + "\n\n")
+
+        # Skip command lines
+        for line in code.split("\n"):
+            if line.startswith("get_ipython"):
```
nickpetrovic, line 405: I'm no jupyter expert, but are there more commands we need to consider? If so, we might need to maintain a list of them and change how we exclude them here.

dleviminzi: I ran a little test of the various other magic jupyter things and it seems like they all start with the get_ipython call, so we should be covered. 

Screenshot 2024-09-13 at 15 55 49

nickpetrovic (approved):

PR #528, opened 2024-09-13
Add subdomain routing

nickpetrovic:
- Add subdomain middleware to route requests
- Add sql migration that adds a new column to the stub table, populates the column, and adds an index on said field
- Add tests

Resolve BE-1682


pkg/repository/backend_postgres.go
```diff
@@ -917,11 +938,48 @@ func (c *PostgresBackendRepository) GetDeploymentByNameAndVersion(ctx context.Co
 	return &deploymentWithRelated, nil
 }
 
+// GetDeploymentByStubGroup retrieves the a deployment by name, version, and stub group
```
dleviminzi, line 941: minor typo: `retrieves the a deployment` should just be `retrieves the deployment`

nickpetrovic: Fixed!

pkg/gateway/middleware.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	fields := make(map[string]string)
+	for i, name := range re.SubexpNames() {
+		if i != 0 && name != "" {
+			fields[name] = match[i]
+		}
+	}
+
+	if fields["name"] == "" || fields["hash"] == "" {
```
luke-lombardi, line 107: are these fields guaranteed to exist in the map?

PR #530, opened 2024-09-13
Shutdown threadpool executor when function errors

jsun-m:
It looks like a log of changes but all im really doing is wrapping the logic around a try-finally block where in the finally block, I shut down the thread pool executor

### DIFF
--- a/sdk/src/beta9/runner/function.py
+++ b/sdk/src/beta9/runner/function.py
@@ -3,7 +3,7 @@ import os
 import time
 import traceback
 from concurrent import futures
-from concurrent.futures import CancelledError, ThreadPoolExecutor
+from concurrent.futures import CancelledError
 from typing import Optional
 
 import cloudpickle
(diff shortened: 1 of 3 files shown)


nickpetrovic (approved):

sdk/src/beta9/runner/function.py
```diff
@@ -131,87 +131,90 @@ def main(channel: Channel):
...
-        futures.thread._threads_queues.clear()
-
-        task_status = TaskStatus.Complete
-        error: Optional[BaseException] = None
-
-        # Invoke function
-        try:
-            get_args_resp: FunctionGetArgsResponse = function_stub.function_get_args(
-                FunctionGetArgsRequest(task_id=task_id)
+    try:
```
luke-lombardi, line 134: can we do this without the indent somehow?

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #535, opened 2024-09-15
add gh action for automated changelog generation

mernit:


dleviminzi:
example of PR it will open https://github.com/slai-labs/beam-docs/pull/354/files

This only works if the PR names are in the format:

Fix: blah blah
Feat: blah blah

jsun-m:
I'm fine with merging this but is there no github actions plugin that already can do this? Or even a github app? @dleviminzi

dleviminzi:
> I'm fine with merging this but is there no github actions plugin that already can do this? Or even a github app? @dleviminzi

When we do the release there is the changelog generator that github has. As far as I can tell there is no plugin that automates scraping that and opening a formatted PR with the contents in another repo.

dleviminzi (approved):

PR #537, opened 2024-09-16
Sanitize secret's values

nickpetrovic:
- Remove (encrypted) secret values from deployment responses
- Move sanitize stub function from task to stub struct
- Update task response function calls to sanitize values; this means secrets come back in the task response, but the values are not returned.
- Add created and updated at dates to stub config of secret so they have valid dates in the frontend
- Remove "empty" attributes of secret type for workspace id and last updated by (both are private ids)

Resolve BE-1799

### DIFF
--- a/pkg/types/backend.go
+++ b/pkg/types/backend.go
@@ -144,23 +144,6 @@ type TaskWithRelated struct {
 	Stub      Stub         `db:"stub" json:"stub"`
 }
 
-func (t *TaskWithRelated) SanitizeStubConfig() error {
-	var stubConfig StubConfigV1
-	err := json.Unmarshal([]byte(t.Stub.Config), &stubConfig)
-	if err != nil {
-		return err
-	}
-
-	stubConfig.Secrets = []Secret{}
-
-	stubConfigBytes, err := json.Marshal(stubConfig)
-	if err != nil {
-		return err
-	}
-	t.Stub.Config = string(stubConfigBytes)
-	return nil
-}
-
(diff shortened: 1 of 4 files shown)


pkg/types/backend.go
```diff
@@ -265,6 +248,27 @@ func (s *Stub) UnmarshalConfig() (*StubConfigV1, error) {
...
+func (s *Stub) SanitizeConfig() error {
+	var config StubConfigV1
+	err := json.Unmarshal([]byte(s.Config), &config)
+	if err != nil {
+		return err
+	}
+
+	// Remove secret values from config
+	for i := range config.Secrets {
+		config.Secrets[i].Value = ""
```
dleviminzi, line 260: Should this be some placeholder value that indicates it exists despite not being shown here?

dleviminzi (approved):

pkg/types/backend.go
nickpetrovic: Yeah, could be helpful to the person seeing it in the browser. Something like `"*****"`? Did you have something in mind?

nickpetrovic: Thought about this a little more... I think I'll leave it empty, at least for now. This ends up being in both the deployment and the task API responses and I wouldn't want someone to do something like below and think there's actually a useful value there. 

```python
# someone might do this
if secret["value"]:
    # do something

# when they should do this
if secret["value"] == "*****":
    # do something
```

PR #538, opened 2024-09-16
ignore patterns like git

dleviminzi:
This PR contains a small change to match patterns the way that git does when ignoring files to sync. It uses the `pathspec` library which was already one of our deps.


sdk/src/beta9/sync.py
```diff
@@ -84,15 +90,9 @@ def _read_ignore_file(self) -> list:
         return patterns
 
     def _should_ignore(self, path: str) -> bool:
```
luke-lombardi, line 92: this seems reasonable, can we add a test case?

sdk/tests/test_filesyncer.py
```diff
@@ -0,0 +1,37 @@
+import os
+from unittest import TestCase
+from unittest.mock import MagicMock
+
+from beta9.sync import IGNORE_FILE_CONTENTS, FileSyncer
+
+
+class TestTaskQueue(TestCase):
```
luke-lombardi, line 8: should this be renamed?

luke-lombardi (changes requested):

sdk/tests/test_filesyncer.py
dleviminzi: Ohp yeah

sdk/tests/test_filesyncer.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class TestFileSyncer(TestCase):
+    def test_init_ignore_file(self):
+        syncer = FileSyncer(gateway_stub=MagicMock())
+        syncer._init_ignore_file()
+        self.assertTrue(syncer.ignore_file_path.exists())
+        os.remove(syncer.ignore_file_path)
+
+    def test_read_ignore_file(self):
```
luke-lombardi, line 15: nit-pick but I would clean up the spacing a bit on these tests, it's hard to read. otherwise good to go.

luke-lombardi (approved):
see last comment, otherwise good to merge

PR #544, opened 2024-09-17
Use query params to pass in token for websocket

jsun-m:


pkg/abstractions/endpoint/buffer.go
```diff
@@ -458,7 +458,13 @@ func (rb *RequestBuffer) afterRequest(req request, containerId string) {
 }
 
 func proxyWebsocketConnection(w http.ResponseWriter, req *http.Request, dialer websocket.Dialer, dstAddress string) error {
-	upgrader := websocket.Upgrader{}
+	upgrader := websocket.Upgrader{
+		CheckOrigin: func(r *http.Request) bool {
+			// Allow all connections
```
luke-lombardi, line 463: // Allow all origins?

pkg/auth/http.go
```diff
@@ -17,12 +17,17 @@ type HttpAuthContext struct {
...
+			var tokenKey string
 			req := c.Request()
 			authHeader := req.Header.Get("Authorization")
-			tokenKey := strings.TrimPrefix(authHeader, "Bearer ")
+			tokenKey = strings.TrimPrefix(authHeader, "Bearer ")
 
 			if authHeader == "" || tokenKey == "" {
-				return next(c)
+				// Check for the token in the Sec-WebSocket-Protocol header
+				tokenKey = parseAuthFromWebsocketProtocolHeader(req.Header.Get("Sec-WebSocket-Protocol"))
```
luke-lombardi, line 27: what about unauthorized websockets?

luke-lombardi (changes requested):

pkg/auth/http.go
jsun-m: we just extract the token, we still verify it through the normal authorize token workflow

luke-lombardi (approved):

PR #546, opened 2024-09-17
stop container endpoint

dleviminzi:


pkg/api/v1/container.go
```diff
@@ -84,3 +85,26 @@ func (c *ContainerGroup) StopAllWorkspaceContainers(ctx echo.Context) error {
...
+		return HTTPBadRequest("invalid workspace id")
+	}
+
+	err = c.scheduler.Stop(containerId)
+	if err != nil {
+		return HTTPInternalServerError("failed to stop container")
+	}
+
+	return ctx.JSON(http.StatusOK, map[string]interface{}{
+		"message": "container stopped",
```
nickpetrovic, line 108: Just a couple of nitpicks. 

Can we capitalize the messages? e.g. "Invalid container ID"

I think the OK response should just be a NoContent response.

dleviminzi: updated. also made the get responses capitalized

PR #552, opened 2024-09-18
use pathspec based logic for ignore file

dleviminzi:
Since actually using pathspec caused an issue, this PR copies the logic of pathspec into class methods for our FileSyncer. This is permissible under MPL 2.0.


jsun-m:
Instead of doing this, can you add pathspec to the runner image?

luke-lombardi:
> Instead of doing this, can you add pathspec to the runner image?

If we update, that will force us to rebuild all images for a dep that solves one very small problem. Doesn't feel worth it.

sdk/src/beta9/sync.py
```diff
@@ -114,6 +117,160 @@ def _calculate_sha256(file_path: str, chunk_size: int = CHUNK_SIZE) -> str:
                 hasher.update(chunk)
         return hasher.hexdigest()
 
+    # The methods _match_file, _pattern_to_regex, and _translate_segment_glob are copied
```
luke-lombardi, line 120: Maybe we should vendor this? This seems like overkill -- this isn't a big enough problem to warrant this much code.

luke-lombardi (changes requested):

sdk/src/beta9/sync.py
dleviminzi: It makes it easier to keep up to date as well. Good with me

PR #558, opened 2024-09-20
Add CPU and memory limits to runc containers

nickpetrovic:
This will prevent workloads from using up all the memory available on a host which can cause a worker to get OOM'd by the kernel leaving the pod in a bad state.

Resolve BE-1780

### DIFF
--- a/pkg/common/redis.go
+++ b/pkg/common/redis.go
@@ -18,6 +18,7 @@ import (
 )
 
 var (
+	ErrNilMessage       = errors.New("redis: nil message")
 	ErrChannelClosed    = errors.New("redis: channel closed")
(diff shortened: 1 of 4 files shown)


pkg/common/redis.go
```diff
@@ -238,6 +239,11 @@ func (r *RedisClient) handleChannelSubs(
 				return
 			}
 
+			if message == nil {
+				errCh <- ErrNilMessage
+				return
+			}
```
nickpetrovic, line 245: I saw a worker crash with a nil pointer error in the EventBus' receive function. It was trying to consume a message that was `nil`.

jsun-m (approved):

PR #565, opened 2024-09-24
Add message handling for when image build fails to pull and archive initial image

jsun-m:


pkg/abstractions/image/build.go
```diff
@@ -321,6 +324,32 @@ func (b *Builder) genContainerId() string {
...
+	// to terminate the build and notify the user.
+	for {
+		_, err := b.containerRepo.GetContainerAddress(containerId)
+		if err == nil {
+			return
+		}
+
+		_, err = b.containerRepo.GetContainerExitCode(containerId)
+		if err == nil {
+			outputChan <- common.OutputMsg{Done: true, Success: false, Msg: "Base image failed to load.\n"}
```
nickpetrovic, line 340: I think we need to check if the exit code is > 0 because it looks like terminateContainer() calls SetContainerExitCode(). We also pass a reference of the exit code to terminateContainer(), which can be set from s.runcHandle.Run().

pkg/abstractions/image/build.go
```diff
@@ -321,6 +324,32 @@ func (b *Builder) genContainerId() string {
 	return fmt.Sprintf("%s%s", types.BuildContainerPrefix, uuid.New().String()[:8])
 }
 
+func (b *Builder) checkForInitialImagePullError(containerId string, outputChan chan common.OutputMsg) {
+	// If the initial image pull of custom base image fails, the container will never receive
+	// a container address for the runc client to connect to.
+	// If it fails without a container address, we can still look up the exit code
+	// to terminate the build and notify the user.
+	for {
```
nickpetrovic, line 332: We might want to pass the calling function's context and check to see if its done using a select{} in the for loop. Doing this should guarantee that if someone disconnects or hits Ctl-C, this will exit regardless whats happening in the build workers.

pkg/abstractions/image/build.go
```diff
@@ -199,6 +199,9 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
 		return err
 	}
 
+	// Get the hostname of the build container
+	go b.checkForInitialImagePullError(containerId, outputChan)
```
luke-lombardi, line 203: I think we can probably do this without an extra goroutine

luke-lombardi (changes requested):

pkg/abstractions/image/build.go
jsun-m: Ideally I check in the gateway to see if the image exists or not. But the packages that help do this require a shitton of other packages

luke-lombardi (changes requested):
@jsun-m Is this still required with the inspect and verify stuff now?

PR #566, opened 2024-09-24
Functions in Colab

dleviminzi:
This PR fixes functions in Colab. The main issue was that beam was not found in the last frame in the stack did not include a beam import. As a result, the beta9 config was being picked up. 

To fix this, I added a condition to check if beam is in the sys modules (its presence there would imply it was imported). I think we might be able to replace the frame logic entirely with this check. Aside from that, I've added some error handling to the file syncing and added some default ignore paths for google drive.

### DIFF
--- a/sdk/src/beta9/abstractions/base/__init__.py
+++ b/sdk/src/beta9/abstractions/base/__init__.py
@@ -1,5 +1,6 @@
 import inspect
 import os
+import sys
 from abc import ABC
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


sdk/src/beta9/abstractions/base/__init__.py
```diff
@@ -69,9 +70,10 @@ def __init_subclass__(cls, /, **kwargs):
         frames = inspect.stack()
         frame = frames[-1]
 
-        if frame.code_context and any(
-            substr in frame.code_context[0] for substr in ("import beam", "from beam")
-        ):
+        if (
+            frame.code_context
+            and any(substr in frame.code_context[0] for substr in ("import beam", "from beam"))
+        ) or "beam" in sys.modules:
```
dleviminzi, line 76: @nickpetrovic @jsun-m I'm curious what you guys think about removing the frame logic and just checking the sys modules.

nickpetrovic: I think that's a good idea. If its there every time when beam is calling beta9 and its not there when only beta9 is being invoked via the CLI, then I say we do it.

nickpetrovic (approved):

sdk/src/beta9/abstractions/base/__init__.py
dleviminzi: I am going to test this a bit more, but I am fairly sure its safe

PR #567, opened 2024-09-24
Rework and fix subdomain routing

nickpetrovic:
- Replace stub.group with deployment.subdomain column in the database. This simplifies creating stubs again and moves the subdomain to the model (deployment) that its actually for.
- Fix generating subdomain hash by using deployment.name instead of stub.name. This makes deployment subdomains unique by deployment.name, stubType, and workspaceId.
- Introduce a new setting to control the invoke url type end-users should see in the UI and CLI. This also removes the localhost hardcode and simplifies URL building logic.
- Add a column check constraint on deployment.subdomain
- Move subdomain middleware to a new middleware package
- Update all tests

Resolve BE-1849 BE-1848

### DIFF
--- a/pkg/common/url.go
+++ b/pkg/common/url.go
@@ -8,42 +8,37 @@ import (
 )
 
 const (
-	InvokeUrlTypePath       string = "path"
-	InvokeUrlTypeSubdomain  string = "subdomain"
-	invokeUrlIgnoreHostname string = "localhost" // Ignore hostname when building local urls
+	InvokeUrlTypePath string = "path"
+	InvokeUrlTypeHost string = "host"
 )
 
-func BuildDeploymentURL(externalUrl, invokeUrlType string, stub *types.StubWithRelated, deployment *types.Deployment) string {
+func BuildDeploymentURL(externalUrl, urlType string, stub *types.StubWithRelated, deployment *types.Deployment) string {
 	parsedUrl, err := url.Parse(externalUrl)
 	if err != nil {
 		return ""
 	}
 
-	isLocalOrPathBased := parsedUrl.Hostname() == invokeUrlIgnoreHostname || invokeUrlType == InvokeUrlTypePath
+	if urlType == InvokeUrlTypeHost {
+		return fmt.Sprintf("%s://%s-v%d.%s", parsedUrl.Scheme, deployment.Subdomain, deployment.Version, parsedUrl.Host)
+	}
+
 	stubConfig, err := stub.UnmarshalConfig()
 	isPublic := err == nil && !stubConfig.Authorized
 
-	if isLocalOrPathBased {
-		if isPublic {
-			return fmt.Sprintf("%s://%s/%s/public/%s", parsedUrl.Scheme, parsedUrl.Host, stub.Type.Kind(), stub.ExternalId)
-		}
-		return fmt.Sprintf("%s://%s/%s/%s/v%d", parsedUrl.Scheme, parsedUrl.Host, stub.Type.Kind(), deployment.Name, deployment.Version)
+	if isPublic {
+		return fmt.Sprintf("%s://%s/%s/public/%s", parsedUrl.Scheme, parsedUrl.Host, stub.Type.Kind(), stub.ExternalId)
 	}
-
-	return fmt.Sprintf("%s://%s-v%d.%s", parsedUrl.Scheme, stub.Group, deployment.Version, parsedUrl.Host)
+	return fmt.Sprintf("%s://%s/%s/%s/v%d", parsedUrl.Scheme, parsedUrl.Host, stub.Type.Kind(), deployment.Name, deployment.Version)
 }
 
-func BuildServeURL(externalUrl, invokeUrlType string, stub *types.StubWithRelated) string {
+func BuildServeURL(externalUrl, urlType string, stub *types.StubWithRelated) string {
 	parsedUrl, err := url.Parse(externalUrl)
 	if err != nil {
 		return ""
 	}
 
-	isLocalOrPathBased := parsedUrl.Hostname() == invokeUrlIgnoreHostname || invokeUrlType == InvokeUrlTypePath
-
-	if isLocalOrPathBased {
-		return fmt.Sprintf("%s://%s/%s/id/%s", parsedUrl.Scheme, parsedUrl.Host, stub.Type.Kind(), stub.ExternalId)
+	if urlType == InvokeUrlTypeHost {
+		return fmt.Sprintf("%s://%s.%s", parsedUrl.Scheme, stub.ExternalId, parsedUrl.Host)
 	}
-
-	return fmt.Sprintf("%s://%s.%s", parsedUrl.Scheme, stub.ExternalId, parsedUrl.Host)
+	return fmt.Sprintf("%s://%s/%s/id/%s", parsedUrl.Scheme, parsedUrl.Host, stub.Type.Kind(), stub.ExternalId)
 }
--- a/pkg/common/url_test.go
+++ b/pkg/common/url_test.go
@@ -7,113 +7,90 @@ import (
 	"github.com/tj/assert"
 )
 
-func TestBuildInvokeURL(t *testing.T) {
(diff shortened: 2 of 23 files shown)


pkg/common/url.go
```diff
@@ -8,40 +8,40 @@ import (
...
 
-func BuildDeploymentURL(externalUrl, invokeUrlType string, stub *types.StubWithRelated, deployment *types.Deployment) string {
+func BuildDeploymentURL(externalUrl, urlType string, stub *types.StubWithRelated, deployment *types.Deployment) string {
 	parsedUrl, err := url.Parse(externalUrl)
 	if err != nil {
 		return ""
 	}
 
-	isLocalOrPathBased := parsedUrl.Hostname() == invokeUrlIgnoreHostname || invokeUrlType == InvokeUrlTypePath
+	if urlType == "" {
```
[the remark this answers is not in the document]
jsun-m:
```
if urlType == "" {
return fmt.Sprintf("%s://%s-v%d.%s", parsedUrl.Scheme, deployment.Subdomain, deployment.Version, parsedUrl.Host)
}
```

pkg/common/url_test.go
```diff
@@ -7,113 +7,90 @@ import (
...
-			invokeType: InvokeUrlTypePath,
-			want:       "https://app.example.com/asgi/id/aff86f02-c968-47a9-9132-0bde826b0aca",
+			name: "returns public path-based URL",
+			stub: &types.StubWithRelated{Stub: types.Stub{
+				Type:       types.StubType(types.StubTypeEndpointDeployment),
+				Config:     `{"authorized": false}`,
+				ExternalId: "e9c29586-c465-4a67-9c9b-25293d1ce77b",
+			}},
+			invokeType:  InvokeUrlTypePath,
+			expectedURL: "http://app.example.com/endpoint/public/e9c29586-c465-4a67-9c9b-25293d1ce77b",
```
jsun-m, line 50: are there public subdomain-based urls?

jsun-m: I feel like we talked about this but I forgot the reason why we can't

jsun-m (changes requested):

pkg/common/url.go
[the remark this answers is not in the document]
nickpetrovic: Yeah it probably is. If I remove this and the urlType is != InvokeUrlTypeHost then it'll fallback to the path-based URL which is probably what we want. 

I'll remove this bit.

pkg/common/url_test.go
nickpetrovic: There used to be, but @mernit and I talked about it and we thought it made sense to just have the host-based URL be the same for public and private endpoints. This keeps to what v1 was doing IIRC.

For example, http://my-app-1234567-v8.app.example.com will be the same whether `authorized` is true or false. We also won't print out the auth header if `authorized` is false, same as before.

pkg/common/url.go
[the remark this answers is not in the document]
nickpetrovic: Just updated. 

I removed that if statement and made the BuildServeURL function follow the same logic e.g. check if host type first else return the path-based URL.

pkg/repository/backend_postgres.go
```diff
@@ -941,7 +920,7 @@ func (c *PostgresBackendRepository) GetDeploymentByNameAndVersion(ctx context.Co
 
 // GetDeploymentByStubGroup retrieves the deployment by name, version, and stub group
```
dleviminzi, line 921: Method names need to be updated still

nickpetrovic: Good catch. Just Fixed it.

pkg/common/url_test.go
jsun-m: ok so http://my-app-1234567-v8.app.example.com/ can still be seen as a public endpoint

nickpetrovic: Yes, public and private (authorized/unauthorized) endpoints will have the same host-based URL.

pkg/gateway/middleware/subdomain.go
```diff
@@ -98,14 +95,30 @@ type SubdomainFields struct {
...
+	if err != nil {
+		h = host // If error, assume no port
 	}
+
+	// Normalize to lower case
+	normalizedHost := strings.ToLower(h)
+	normalizedBaseDomain := strings.ToLower(baseDomain)
+
+	// If host equals baseDomain, return empty string
+	if normalizedHost == normalizedBaseDomain {
```
jsun-m, line 113: i assume this is for path routing instead of subdomain routing.

pkg/gateway/middleware/subdomain_test.go
```diff
@@ -222,20 +222,20 @@ func TestParseSubdomainFields(t *testing.T) {
 		{
 			subdomain: "app1-9a7dbcc",
 			expected: &SubdomainFields{
-				StubGroup: "app1-9a7dbcc",
+				Subdomain: "app1-9a7dbcc",
```
jsun-m, line 225: This doesn't have a version. Should the version of this also be latest?

dleviminzi (approved):

jsun-m (changes requested):

pkg/gateway/middleware/subdomain.go
nickpetrovic: Yes. If there's no difference between these two, then there isn't a subdomain to parse.

pkg/gateway/middleware/subdomain_test.go
nickpetrovic: No because parseSubdomainFields() doesn't set defaults. It only extracts named groups from the subdomain regex.

So in this case, because version doesn't exist in the subdomain when its parsed, the Version field will be an empty string. This is eventually passed to getStubForSubdomain(). In there, the version is parsed and 0 will be the default version if its not parseable. 

From there, the query defined in GetDeploymentBySubdomain has a case statment that will check if the version is 0. If so, it will find the latest version on the deployment CTE/subquery. If its not 0, it'll use the version provided.

nickpetrovic: Sorry, so technically parseSubdomainFields() does set defaults, but they're empty strings.

Thinking about this more, it could make more sense to move parseVersion(fields.Version) out of getStubForSubdomain() and to the parseSubdomainFields() function. Mostly to keep parsing logic closer together. What do you think?

nickpetrovic: I tested this out and it does seem to work well. The one major difference is that Version would be an uint and if its parsed as 0 (meaning latest), then we will always look for and invoke a handler that has "latest" in the path e.g. `/endpoint/:name/latest`.

For example: 

```go
g.GET("/:deploymentName", handler)           // no longer invoked by the middleware
g.GET("/:deploymentName/latest", handler)
g.GET("/:deploymentName/v:version", handler)
```

This isn't an issue, just pointing the differences out. Otherwise, the functionality seems to work the same, but the handler path is more explicit.

Last thing to note is the buildHandlerPath function will have to be updated to handle the Version uint.

```
func buildHandlerPath(stub *types.Stub, fields *SubdomainFields, extraPaths ...string) string {
	pathSegments := []string{"/" + stub.Type.Kind()}

	if stubConfig, err := stub.UnmarshalConfig(); err == nil && !stubConfig.Authorized {
		pathSegments = append(pathSegments, "public", stub.ExternalId)
	} else if fields.StubId != "" {
		pathSegments = append(pathSegments, "id", fields.StubId)
	} else {
		pathSegments = append(pathSegments, fields.Name)

		if fields.Version > 0 {
			pathSegments = append(pathSegments, fmt.Sprintf("v%d", fields.Version))
		} else {
			pathSegments = append(pathSegments, "latest")
		}
	}

	pathSegments = append(pathSegments, extraPaths...)
	return path.Join(pathSegments...)
}
```

I think I like this better too. Mostly for its explicitness in calling a handler with a :latest path param.

nickpetrovic: Just pushed this change.

PR #571, opened 2024-09-26
Add init lock for gateway

nickpetrovic:
Initially this would be used for database migrations. It is generic enough that it can be used in the future for other initialization locks.

In the database repo, this attempts to acquire a lock and when it does, a function is returned. This function is used to release the lock. If a lock is unable to be obtained, then the error is ignored and the system resumes normal operation. 

Resolve BE-1861

### DIFF
--- a/pkg/repository/backend_postgres.go
+++ b/pkg/repository/backend_postgres.go
@@ -52,7 +52,7 @@ func GenerateDSN(config types.PostgresConfig) string {
 	)
 }
 
-type LockMigrationFunc func(string) (func(), error)
+type MigrationLockFunc func(string) (func(), error)
 
 type PostgresBackendRepository struct {
 	client    *sqlx.DB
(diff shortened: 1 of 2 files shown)


pkg/repository/backend_postgres.go
```diff
@@ -72,8 +74,11 @@ func NewBackendPostgresRepository(config types.PostgresConfig, eventRepo EventRe
 		eventRepo: eventRepo,
 	}
 
-	if err := repo.migrate(); err != nil {
-		log.Fatalf("failed to run backend migrations: %v", err)
+	if release, err := obtainMigrationLock("postgres"); err == nil {
+		defer release()
+		if err := repo.migrate(); err != nil {
+			log.Fatalf("Failed to run backend migrations: %v", err)
+		}
```
dleviminzi, line 81: Ignoring the error here when its not nil means we are skipping for now if we fail to acquire the lock?

dleviminzi: If we were to err here, the startup would fail entirely, which seems like that might keep us in a more consistent state. The new logic wouldn't potentially be out ahead of the migration being complete. Also, it should retry rolling out the update anyway right?

dleviminzi: Another benefit is that we could do the locking logic in `NewGateway` instead of passing it down into repo.

nickpetrovic: Hm yeah, I'll move it out and keep it in `NewGateway`.

Can you explain more what you mean in your 2nd comment? I'm not sure I follow.

dleviminzi (approved):

pkg/repository/backend_postgres.go
nickpetrovic: Made migrate a public function and now its being wrapped by the lock.

PR #572, opened 2024-09-26
Add semaphore to request buffer

jsun-m:
## Problem Statement

When multiple requests arrive simultaneously at the request buffer, they are sent to the Gunicorn worker without respecting the `worker` limit, leading to a race condition on Gunicorn’s middleware request interceptor.

## Solution Overview

We aim to enhance handling at the request buffer level to manage and balance requests more effectively. Key considerations include:

- **Performance**: Utilizing locks must not degrade performance as we scale up with more gateways and concurrent requests.
- **Consistency**: Should a gateway crash while holding a lock, the lock must expire naturally to avoid buildup of unreachable containers due to false occupancy.

## Approach: Redis-Based Semaphores

Redis, with its single-threaded nature and in-memory capabilities, provides a high-performance solution for managing transactional requests.

### Blocking List Approach

Using `BLPOP` and `BLPUSH` as a counter, we initialize a list for each `container_id` with a predefined number of values (locks) corresponding to the number of tasks each container can handle.

#### Pros
- **Simple**: No need for custom Lua scripting.
- **Performant**: Requires only one Redis query.

#### Cons
- **Reliability**: If the process holding a lock crashes, the lock is never reinserted, potentially starving the semaphore.

### ZSETs - Redis Sorted Set Semaphore (Advanced Solution)

Adapted from "Redis in Action" by Josiah L. Carlson, this approach uses `ZREMRangeByScore` within a Redis Sorted Set to handle lock expiration efficiently. This method compares a numeric value (timestamp) against all values in the ZSET and deletes them if they meet the specified condition. This operation is performed within O(log(N) + M), where N is the number of elements in the set and M is the number of elements removed.

#### Advantages
- **Efficiency**: Highly efficient for our use case, making it suitable for high-load environments.
- **Reliability**: This method addresses the reliability issues associated with shared mutexes by allowing each lock to be individually referenced and updated.
- **Self-cleaning**: This method will expired locks that are no longer used even if the holder of the lock crashed. It will expire if it does not refresh

### Implementation in Request Buffer

Each request must acquire a semaphore lock before proceeding, ensuring that each container handles requests up to its predefined limit.

```markdown
Each container will manage its own semaphore, ensuring that resource allocation adheres strictly to defined limits.

### DIFF
--- a/pkg/abstractions/endpoint/buffer.go
+++ b/pkg/abstractions/endpoint/buffer.go
@@ -52,6 +52,7 @@ type RequestBuffer struct {
 	buffer                  *abstractions.RingBuffer[request]
 	availableContainers     []container
 	availableContainersLock sync.RWMutex
+	maxTaskPerContainer     int
 
 	length atomic.Int32
 
@@ -88,6 +89,12 @@ func NewRequestBuffer(
 		tsConfig:  tsConfig,
 
 		isASGI: isASGI,
+	}
+
+	if stubConfig.Autoscaler == nil || stubConfig.Autoscaler.TasksPerContainer == 0 {
+		b.maxTaskPerContainer = 1
+	} else {
+		b.maxTaskPerContainer = int(stubConfig.Autoscaler.TasksPerContainer)
 	}
 
 	go b.discoverContainers()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jsun-m** [comment]
Will require a lot of load testing

**Comment by nickpetrovic** [comment]
I think this makes sense. Can we add constants for the `time.Duration`s?

File: pkg/abstractions/endpoint/buffer.go [hunk 1]
```diff
@@ -71,14 +72,14 @@ func NewRequestBuffer(
...
-		availableContainers: []container{},
-
+		ctx:                     ctx,
+		rdb:                     rdb,
+		workspace:               workspace,
+		stubId:                  stubId,
+		stubConfig:              stubConfig,
+		buffer:                  abstractions.NewRingBuffer[request](size),
+		availableContainers:     []container{},
+		maxWorkersPerContainer:  int(stubConfig.Workers),
```
nickpetrovic, line 82: stubConfig is a pointer. Should we check if this is null first?

pkg/abstractions/endpoint/buffer.go
```diff
@@ -315,7 +316,26 @@ func (rb *RequestBuffer) handleRequest(req request) {
...
+		for {
+			select {
+			case <-req.ctx.Request().Context().Done():
+				return
+			case <-time.After(30 * time.Second):
+				rb.rdb.RefreshSemaphore(req.ctx.Request().Context(), c.id, lockId, time.Duration(1*time.Minute))
+			}
+		}
+	}()
+
```
nickpetrovic, line 335: Several arbitrary durations. Can we add constants for them?

nickpetrovic (approved):

jsun-m:
!!! Project put on hold due to causing high resource usage in Redis

PR #574, opened 2024-10-01
Remove running tasks from endpoint serve autoscaler

dleviminzi:
This fixes the issue where users with tasks pending have zombie endpoint serves.

### DIFF
--- a/pkg/abstractions/endpoint/autoscaler.go
+++ b/pkg/abstractions/endpoint/autoscaler.go
@@ -72,7 +72,7 @@ func endpointServeScaleFunc(i *endpointInstance, sample *endpointAutoscalerSampl
 		}
 	}
 
-	if sample.TotalRequests == 0 && exists == 0 {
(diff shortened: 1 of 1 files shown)


pkg/abstractions/endpoint/autoscaler.go
```diff
@@ -72,7 +72,7 @@ func endpointServeScaleFunc(i *endpointInstance, sample *endpointAutoscalerSampl
 		}
 	}
 
-	if sample.TotalRequests == 0 && exists == 0 {
+	if exists == 0 {
```
nickpetrovic, line 75: Are we basically just saying if there is no serve lock, scale the containers to zero?

What happens with the requests that may be in flight?

dleviminzi: They will just expire which I think is reasonable

nickpetrovic (approved):

PR #577, opened 2024-10-02
Ping database in health check

nickpetrovic:
- Fails the HTTP health check if it cannot ping the database
- Pass readiness probe on just 1 successful check
- Lower readiness probe timeout to 1 second

Resolve BE-1866

### DIFF
--- a/deploy/charts/beta9/templates/_helpers.tpl
+++ b/deploy/charts/beta9/templates/_helpers.tpl
@@ -53,10 +53,11 @@ controllers:
             enabled: true
             custom: true
             spec:
+              successThreshold: 1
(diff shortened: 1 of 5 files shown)


deploy/charts/beta9/templates/_helpers.tpl
```diff
@@ -53,10 +53,11 @@ controllers:
             enabled: true
             custom: true
             spec:
+              successThreshold: 1
```
nickpetrovic, line 56: Default was 1. Setting this so its explicit.

dleviminzi (approved):
lgtm!

PR #582, opened 2024-10-02
Fix: error handling in function runner

nickpetrovic:
- Marks the task status as error when it fails
- Fixes or prevents container from running until task expires e.g. "container still running"
- Fixes issue where CLI never exits - still mentions "Waiting..." even though the error has been printed
- Fixes scoping with some vars

Resolve BE-1904

### DIFF
--- a/sdk/src/beta9/runner/function.py
+++ b/sdk/src/beta9/runner/function.py
@@ -128,13 +128,12 @@ def main(channel: Channel):
     if not task_id:
         raise RunnerException("Invalid runner environment")
 
-    start_time = time.time()
-
     with StdoutJsonInterceptor(task_id=task_id):
         container_id = config.container_id
         container_hostname = config.container_hostname
 
         # Start the task
+        start_time = time.time()
         start_task_response = start_task(gateway_stub, task_id, container_id)
         if not start_task_response.ok:
             raise RunnerException("Unable to start task")
@@ -202,6 +201,7 @@ def invoke_function(function_stub, context, task_id):
 def complete_task(gateway_stub, result, task_id, container_id, container_hostname, start_time):
     task_status = TaskStatus.Complete
     task_duration = time.time() - start_time
+    keep_warm_seconds = 0
 
     end_task_response = end_task_and_send_callback(
         gateway_stub=gateway_stub,
@@ -212,7 +212,7 @@ def complete_task(gateway_stub, result, task_id, container_id, container_hostnam
             task_status=task_status,
             container_id=container_id,
             container_hostname=container_hostname,
-            keep_warm_seconds=0,
+            keep_warm_seconds=keep_warm_seconds,
         ),
         override_callback_url=None,
     )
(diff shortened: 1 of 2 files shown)


sdk/src/beta9/runner/function.py
```diff
@@ -131,87 +125,120 @@ def main(channel: Channel):
     gateway_stub: GatewayServiceStub = GatewayServiceStub(channel)
     task_id = config.task_id
 
-    with StdoutJsonInterceptor(task_id=task_id), ThreadPoolExecutorOverride() as thread_pool:
+    if not task_id:
+        raise RunnerException("Invalid runner environment")
+
+    start_time = time.time()
```
jsun-m, line 131: Why did  we need to move the start_time out?

sdk/src/beta9/runner/function.py
```diff
@@ -131,87 +125,120 @@ def main(channel: Channel):
...
-                task_id=task_id,
-                task_duration=task_duration,
-                task_status=task_status,
-                container_id=container_id,
-                container_hostname=container_hostname,
-                keep_warm_seconds=0,
-            ),
-            override_callback_url=kwargs.get("callback_url"),
-        )
+def start_task(gateway_stub, task_id, container_id):
```
jsun-m, line 176: did this functioanlly change anything?

jsun-m (changes requested):
This looks much cleaner. However what was the issue that this is solving in terms of the hanging function? How does moving the threadpoolexecutor further into the scope of the log interceptor help aid it in shutting down a container?

sdk/src/beta9/runner/function.py
nickpetrovic: No, but it works with the theme of having the functions described as invoke, handle error, and complete task. It looks cleaner is all. Functionally, this does not worsen the code, it just makes it easier to read.

jsun-m: makes sense

sdk/src/beta9/runner/function.py
```diff
@@ -131,87 +125,120 @@ def main(channel: Channel):
...
+            try:
+                result = invoke_function(function_stub, context, task_id)
+            except BaseException as e:
+                handle_task_failure(e, gateway_stub, task_id, container_id, container_hostname)
+                raise
+            finally:
+                monitor_future.cancel()
+
+            # End the task and send callback
+            complete_task(
```
dleviminzi, line 165: Is there no need to handle exceptions on these calls? If we fail to complete the task should we re-try?

nickpetrovic: No, its what I thought too but found out that `end_task_and_send_callback` (which is called by complete_task) has this decorator on it.

```python
@retry_grpc_call(exception_to_check=BaseException, tries=4, delay=5, backoff=2)

```

nickpetrovic: I mean we might still need it, but I don't knwo what we'd do with the error since we can complete the task.

dleviminzi: Ahh I see

sdk/src/beta9/runner/function.py
nickpetrovic: Just moved a lot around. Its moved back to where it was.

jsun-m (approved):

PR #584, opened 2024-10-03
Send errors to callback url in function invocations

nickpetrovic:
- Send exceptions/errors to callback if they happen
- Make Config fields required (not optional)
- Cleanup errors with custom exception classes
- Improve type annotations

Resolve BE-1904

### DIFF
--- a/sdk/src/beta9/runner/common.py
+++ b/sdk/src/beta9/runner/common.py
@@ -17,6 +17,7 @@ from starlette.responses import Response
 
 from ..clients.gateway import (
     EndTaskRequest,
+    EndTaskResponse,
     GatewayServiceStub,
     SignPayloadRequest,
     SignPayloadResponse,
@@ -30,18 +31,18 @@ USER_CACHE_DIR = "/cache"
 
 @dataclass
 class Config:
-    container_id: Optional[str]
-    container_hostname: Optional[str]
-    stub_id: Optional[str]
-    stub_type: Optional[str]
-    workers: Optional[int]
-    keep_warm_seconds: Optional[int]
-    timeout: Optional[int]
+    container_id: str
+    container_hostname: str
+    stub_id: str
+    stub_type: str
+    workers: int
+    keep_warm_seconds: int
+    timeout: int
     python_version: str
     handler: str
-    on_start: Optional[str]
-    callback_url: Optional[str]
-    task_id: Optional[str]
+    on_start: str
+    callback_url: str
+    task_id: str
     bind_port: int
     volume_cache_map: Dict
 
@@ -232,7 +233,7 @@ def end_task_and_send_callback(
     payload: Any,
     end_task_request: EndTaskRequest,
     override_callback_url: Optional[str] = None,
-):
+) -> EndTaskResponse:
     resp = gateway_stub.end_task(end_task_request)
 
     send_callback(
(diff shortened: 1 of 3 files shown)


sdk/src/beta9/runner/common.py
```diff
@@ -30,18 +31,18 @@
...
+    stub_id: str
+    stub_type: str
+    workers: int
+    keep_warm_seconds: int
+    timeout: int
     python_version: str
     handler: str
-    on_start: Optional[str]
-    callback_url: Optional[str]
-    task_id: Optional[str]
```
nickpetrovic, line 44: This is to correct the linting issues.

TODO: Need to test endpoints, taskqueues, and serves because dataclasses do care about something being Optional or not.

dleviminzi (approved):
Code looks good, but can you help me understand the before and after behavior?

sdk/src/beta9/runner/function.py
```diff
@@ -161,10 +169,13 @@ def main(channel: Channel):
...
                 result = invoke_function(function_stub, context, task_id)
-            except BaseException as e:
-                handle_task_failure(e, gateway_stub, task_id, container_id, container_hostname)
+                if result.exception:
+                    raise result.exception
+            except BaseException:
+                handle_task_failure(result, gateway_stub, task_id, container_id, container_hostname)
                 raise
             finally:
                 monitor_future.cancel()
```
dleviminzi, line 181: Does the `monitor_future.cancel()` need to happen after `handle_task_failure(result, gateway_stub, task_id, container_id, container_hostname)`?

nickpetrovic: Here's the docstring on the cancel function.

> Returns True if the future was cancelled, False otherwise. A future cannot be cancelled if it is running or has already completed.

This is almost a noop because it always has started for me, but if for some reason the future hasn't already started, we can cancel or prevent it from ever running.

I debated on even calling this function at all. Before these changes, this was called after the code thats now in `complete_task()`. Reason being that because this is within a ThreadPoolExector() context, thread_pool.shutdown() is called. I'm wondering if it makes more sense to call thread_pool.shutdown(wait=False) explicitly after `complete_task()` since the default shutdown() uses wait=True.

nickpetrovic: The more I've read about this the more I do think it's better that we explicitly set `.shutdown(wait=False)`. I've updated the code to do this now.

dleviminzi: Yeah I think it will be cancelled when the thread pool context is exited anyway. Doesn't hurt to be explicit, but if omitted, I think it could be refactored to look like this: 

```python
with ThreadPoolExecutor() as thread_pool:
    monitor_future = thread_pool.submit(
        _monitor_task,
        context=context,
        stub_id=config.stub_id,
        task_id=task_id,
        container_id=container_id,
        function_stub=function_stub,
        gateway_stub=gateway_stub,
    )

    result = invoke_function(function_stub, context, task_id)

    if result.exception:
        handle_task_failure(result, gateway_stub, task_id, container_id, container_hostname)
        raise result.exception

    # End the task and send callback
    complete_task(
        gateway_stub,
        result,
        task_id,
        container_id,
        container_hostname,
        start_time,
    )
```

nickpetrovic: Hm. That does look cleaner. Let me test that real quick.

PR #587, opened 2024-10-03
Print error logs in function callable

nickpetrovic:
Fixes printing error logs in the function callable. It was removed during a refactor because the error logs were being printed twice, once in function:main() and once in with_runner_context.

To simplify this, we're decoupling the channel decorators and making it optional that the upstream error handler prints a stack trace. Also including a minor change in reordering params for handle_task_failure to be consistent with complete_task.

Resolve BE-1905

### DIFF
--- a/sdk/src/beta9/runner/function.py
+++ b/sdk/src/beta9/runner/function.py
@@ -135,8 +135,8 @@ def _monitor_task(
             os._exit(0)
 
 
-@pass_channel
 @handle_error(print_traceback=False)
+@pass_channel
 def main(channel: Channel):
     function_stub: FunctionServiceStub = FunctionServiceStub(channel)
     gateway_stub: GatewayServiceStub = GatewayServiceStub(channel)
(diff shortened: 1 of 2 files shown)


sdk/src/beta9/runner/function.py
```diff
@@ -171,8 +172,9 @@ def main(channel: Channel):
         # Invoke the function and handle its result
         result = invoke_function(function_stub, context, task_id)
         if result.exception:
-            handle_task_failure(result, gateway_stub, task_id, container_id, container_hostname)
             thread_pool.shutdown(wait=False)
+            handle_task_failure(gateway_stub, result, task_id, container_id, container_hostname)
+            print(result.exception)
```
jsun-m, line 177: cleanup

jsun-m (approved):

sdk/src/beta9/runner/function.py
nickpetrovic: We want this. This needs to be picked up by the log interceptor so the log gets a task id.

PR #588, opened 2024-10-04
Implement decorator for json output interceptor

nickpetrovic:
- Make a version of the output interceptor as a decorator
- Use new decorator on function runner entrypoint to capture all logs and reduce indentation

Resolve BE-1905

### DIFF
--- a/sdk/src/beta9/logging.py
+++ b/sdk/src/beta9/logging.py
@@ -1,10 +1,12 @@
+import functools
 import io
 import json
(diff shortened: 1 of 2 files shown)


nickpetrovic:
@jsun-m I wanted to figure out how to capture all logs, I think this will do it. Plus it deindents the code. This should be it.

dleviminzi (approved):
nice

PR #590, opened 2024-10-05
Add worker management command to CLI

nickpetrovic:
- Add worker management command group
- Add list, cordon, uncordon, and drain worker commands
- Prevent requests from being scheduled on disabled workers
- Prevent disabled workers from being calculated in pool capacity functionality

Resolve BE-1601


pkg/common/redis.go
```diff
@@ -314,16 +314,20 @@ func (l *RedisLock) Release(key string) error {
...
 		if dstField.IsValid() && dstField.CanSet() {
-			dstField.Set(srcVal.Field(i))
+			if srcField.Type() == dstField.Type() {
+				dstField.Set(srcVal.Field(i))
+			} else if srcField.Kind() == reflect.String && dstField.Kind() == reflect.String {
+				dstField.Set(reflect.ValueOf(srcField.String()))
+			}
 		}
 	}
 }
```
dleviminzi, line 333: We could use [mapstructure](https://pkg.go.dev/github.com/go-viper/mapstructure/v2) which is forked from [here](https://github.com/mitchellh/mapstructure) to do this struct copying. It can be used like this 
```go
type Source struct {
	Name    string
	Age     int
	Active  bool
	Address string
}

type Destination struct {
	Name    string
	Age     float64
	Active  bool
	Address *string
}
```
```go
	source := Source{
		Name:    "John Doe",
		Age:     30,
		Active:  true,
		Address: "123 Main St",
	}

	var destination Destination

	config := &mapstructure.DecoderConfig{
		WeaklyTypedInput: true,
		Result:           &destination,
	}

	decoder, err := mapstructure.NewDecoder(config)
    // check error

	err = decoder.Decode(source)
	// check error

    // destination will now be {Name:John Doe Age:30 Active:true Address:123 Main St} 
```
This seems preferable to rolling our own imo

nickpetrovic: I'd prefer it too honestly, but I'll have to do some testing on this to make sure the behavior is the same. I'll check it out today/tomorrow.

nickpetrovic: Actually, I'm going to follow up in a different PR with this. So for now I'll just assign the values directly.

dleviminzi: Ok sounds good!

luke-lombardi (approved):

PR #591, opened 2024-10-05
Improve endpoint request buffering

luke-lombardi:
Some quick fixes for endpoint request buffering, implements a lock-less flow control setup using tokens

- Instead of relying on locking, invert request counter to use token based flow control where max requests per container is equal to the number of available workers in the endpoint. This is very similar to the way it was before, but with more checks around capacity / combining set/get operations into a single atomic operation

- Check available containers without lock, _before_ trying to pop the request from the buffer to prevent unnecessary pops from fifo

- If requests are cancelled client side, cancel the tasks in the database / task dispatcher immediately

- Add priority flag to ring buffer to prioritize failed request retries before newer requests (also add some test cases for it)

- Some clean up and stuff

**NOTE**: Currently this uses the number of workers available to the endpoint as the request limit per container, but this could be something else that we allow users to set on the stub config.

### DIFF
--- a/pkg/abstractions/common/ring_buffer.go
+++ b/pkg/abstractions/common/ring_buffer.go
@@ -25,10 +25,11 @@ func (rb *RingBuffer[T]) Push(request T, priority bool) {
 	rb.mu.Lock()
 	defer rb.mu.Unlock()
 
+	// If priority is true, move head backward to insert at the front
 	if priority {
-		// Priority insert: move head backward to insert at the front
 		rb.head = (rb.head - 1 + rb.size) % rb.size
 		rb.buffer[rb.head] = request
+
 		if rb.count < rb.size {
 			rb.count++
 		} else {
@@ -36,11 +37,12 @@ func (rb *RingBuffer[T]) Push(request T, priority bool) {
 			rb.tail = (rb.tail - 1 + rb.size) % rb.size
 		}
 	} else {
-		// Normal FIFO insert: insert at the tail
+		// Normal FIFO insert, just insert at the tail
 		rb.buffer[rb.tail] = request
 		rb.tail = (rb.tail + 1) % rb.size
+
+		// Buffer is full, move head forward to overwrite oldest element
 		if rb.count == rb.size {
-			// Buffer is full, move head forward to overwrite oldest element
 			rb.head = (rb.head + 1) % rb.size
 		} else {
 			rb.count++
--- a/pkg/abstractions/common/ring_buffer.go
+++ b/pkg/abstractions/common/ring_buffer.go
@@ -26,21 +26,25 @@ func (rb *RingBuffer[T]) Push(request T, priority bool) {
 	defer rb.mu.Unlock()
 
 	if priority {
+		// Priority insert: move head backward to insert at the front
 		rb.head = (rb.head - 1 + rb.size) % rb.size
 		rb.buffer[rb.head] = request
 		if rb.count < rb.size {
 			rb.count++
+		} else {
+			// Buffer is full, adjust tail as the oldest element is dropped
+			rb.tail = (rb.tail - 1 + rb.size) % rb.size
 		}
 	} else {
-		// Normal FIFO behavior, insert at tail
+		// Normal FIFO insert: insert at the tail
+		rb.buffer[rb.tail] = request
+		rb.tail = (rb.tail + 1) % rb.size
 		if rb.count == rb.size {
+			// Buffer is full, move head forward to overwrite oldest element
 			rb.head = (rb.head + 1) % rb.size
 		} else {
 			rb.count++
 		}
-
-		rb.buffer[rb.tail] = request
-		rb.tail = (rb.tail + 1) % rb.size
 	}
 }
 
--- a/pkg/abstractions/common/ring_buffer.go
+++ b/pkg/abstractions/common/ring_buffer.go
@@ -20,19 +20,28 @@ func NewRingBuffer[T any](size int) *RingBuffer[T] {
 	}
 }
 
-// Push adds a new request to the buffer
-func (rb *RingBuffer[T]) Push(request T) {
+// Push adds a new request to the buffer. If priority is true, it gets inserted at the front of the buffer
+func (rb *RingBuffer[T]) Push(request T, priority bool) {
 	rb.mu.Lock()
 	defer rb.mu.Unlock()
 
-	if rb.count == rb.size {
-		rb.head = (rb.head + 1) % rb.size
+	if priority {
+		rb.head = (rb.head - 1 + rb.size) % rb.size
+		rb.buffer[rb.head] = request
+		if rb.count < rb.size {
+			rb.count++
+		}
 	} else {
-		rb.count++
-	}
+		// Normal FIFO behavior, insert at tail
+		if rb.count == rb.size {
+			rb.head = (rb.head + 1) % rb.size
+		} else {
+			rb.count++
+		}
 
-	rb.buffer[rb.tail] = request
-	rb.tail = (rb.tail + 1) % rb.size
+		rb.buffer[rb.tail] = request
+		rb.tail = (rb.tail + 1) % rb.size
+	}
 }
 
 // Pop retrieves and removes the oldest request from the buffer
--- a/pkg/abstractions/common/ring_buffer.go
+++ b/pkg/abstractions/common/ring_buffer.go
@@ -25,19 +25,19 @@ func (rb *RingBuffer[T]) Push(request T, priority bool) {
 	rb.mu.Lock()
 	defer rb.mu.Unlock()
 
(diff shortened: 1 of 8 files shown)


pkg/abstractions/endpoint/buffer.go
```diff
@@ -232,43 +236,62 @@ func (rb *RequestBuffer) discoverContainers() {
...
+	tokenKey := Keys.endpointRequestTokens(rb.workspace.Name, rb.stubId, containerId)
+
+	val, err := rb.rdb.Get(rb.ctx, tokenKey).Int()
 	if err != nil && err != redis.Nil {
 		return 0, err
 	} else if err == redis.Nil {
-		return 0, nil
+		maxTokens := int(rb.stubConfig.Workers)
+
+		err = rb.rdb.SetNX(rb.ctx, tokenKey, maxTokens, 0).Err()
```
nickpetrovic, line 253: This function is being called by multiple goroutines and could result in a race condition. Its possible that one or more goroutines could both get redis.Nil and attempt to call `SetNX`, but with just one goroutine succeeding. Even using `Set` here could overwrite another goroutines call to it. 

From what I can tell `SetNX` is atomic, so we should probably call it first with .Result() and check for the error and if the key was created. Then if the key was created, we can return `maxTokens` or if it wasn't we can return `Get().Int()`

nickpetrovic (approved):
This seems to work well from my testing. I've noticed some pauses when sending requests. I can't quite tell what it is, but regardless, this doesn't keep sending requests until the the health check fails which is an improvement.

I added a couple of comments. One I think is a race condition, another just for clarification.

nickpetrovic (changes requested):
Accidentally approved it earlier. I was going to approve it until I saw the potential race condition.

pkg/abstractions/common/ring_buffer.go
```diff
@@ -20,19 +20,34 @@ func NewRingBuffer[T any](size int) *RingBuffer[T] {
...
+		if rb.count == rb.size {
+			// Buffer is full, overwrite the oldest element at head without moving pointers
+			// No adjustment to head or tail, since the count remains the same
+			rb.buffer[rb.head] = request
+		} else {
+			// Move head backward to insert at front
+			rb.head = (rb.head - 1 + rb.size) % rb.size
+			rb.buffer[rb.head] = request
+			rb.count++
+		}
```
dleviminzi, line 38: If priority is for failed requests and tons of requests start failing it seems like we'll just start dropping a lot of those. Is that ok? Not sure if I'm misunderstanding the use case/if that's intended.

pkg/abstractions/endpoint/buffer.go
```diff
@@ -127,14 +121,20 @@ func (rb *RequestBuffer) processRequests() {
...
 			if !ok {
-				time.Sleep(time.Millisecond * 100)
+				time.Sleep(requestProcessingInterval)
 				continue
 			}
 
 			if req.ctx.Request().Context().Err() != nil {
-				// Context has been cancelled
+				// Request context has been cancelled, mark task as cancelled
+				req.task.Cancel(context.Background(), types.TaskCancellationReason(types.TaskRequestCancelled))
```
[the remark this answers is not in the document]
luke-lombardi: Currently, if a request is cancelled client side, the task associated with that request will still remain pending until the task expiration time elapses. The side effect of this is that the autoscaler won't spin down their container until all of those tasks expire -- with the current logic that could be up to 20 minutes. With this change, if the actual http request is cancelled, the task object associated with that request immediately goes to `CANCELLED` and the task dispatcher removes the task.

pkg/abstractions/common/ring_buffer.go
luke-lombardi: The current behavior is the failed request will be inserted at the back of the FIFO. The failure mode we're improving here is that a container was temporarily busy, and we want to rebuffer that request with high priority so its handled as soon as possible (since it was requested first). But yes, if tons of requests are failing, tons of requests will still fail. Not much we can do about that.

pkg/abstractions/endpoint/buffer.go
luke-lombardi: Yeah probably true. Might even make sense to just create this when the endpoint container is scheduled in instance.go... gotta think about that a little more though

luke-lombardi: Eh, nah I went with the setnx/get. Should be good now.

pkg/abstractions/endpoint/buffer.go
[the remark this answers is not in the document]
nickpetrovic: That makes sense.  We'll want to communicate or document this in our docs. We brought this up last week and it was undecided on whether or not we wanted this to happen.

nickpetrovic (approved):

pkg/abstractions/endpoint/buffer.go
```diff
@@ -232,43 +236,71 @@ func (rb *RequestBuffer) discoverContainers() {
...
 
-	err = rb.rdb.Expire(rb.ctx, Keys.endpointRequestsInFlight(rb.workspace.Name, rb.stubId, containerId), time.Duration(rb.stubConfig.TaskPolicy.Timeout)*time.Second).Err()
+	// If the token count is negative, we exceeded our threshold of
+	// available request tokens, just reverse the operation
+	if tokenCount < 0 {
+		rb.rdb.Incr(rb.ctx, tokenKey)
+		return errors.New("too many in-flight requests")
+	}
+
+	err = rb.rdb.Expire(rb.ctx, tokenKey, time.Duration(rb.stubConfig.TaskPolicy.Timeout)*time.Second).Err()
```
jsun-m, line 287: Let's say 2 gateways are running: G1 and G2. G1 has 3/5 tokens and G2 has 2/5 tokens. 
If G1 crashes. 3 tokens are just gone. Since G2 keeps refreshing the expiration, the key never expires and resets

jsun-m (changes requested):

pkg/abstractions/common/ring_buffer.go
```diff
@@ -20,19 +20,34 @@ func NewRingBuffer[T any](size int) *RingBuffer[T] {
...
-	}
+		// Normal FIFO insert at tail
+		rb.buffer[rb.tail] = request
+		rb.tail = (rb.tail + 1) % rb.size
 
-	rb.buffer[rb.tail] = request
-	rb.tail = (rb.tail + 1) % rb.size
+		// Buffer is full, move head forward to overwrite oldest element
+		if rb.count == rb.size {
+			rb.head = (rb.head + 1) % rb.size
```
jsun-m, line 46: Also if we do overwrite it, it doesn't seem to cancel the users request. The request needs to be canceled or the user client will be left hanging

pkg/abstractions/common/ring_buffer.go
```diff
@@ -20,19 +20,34 @@ func NewRingBuffer[T any](size int) *RingBuffer[T] {
...
+			// Move head backward to insert at front
+			rb.head = (rb.head - 1 + rb.size) % rb.size
+			rb.buffer[rb.head] = request
+			rb.count++
+		}
 	} else {
-		rb.count++
-	}
+		// Normal FIFO insert at tail
+		rb.buffer[rb.tail] = request
```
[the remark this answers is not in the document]
luke-lombardi: That would only happen if the buffer was full.

pkg/abstractions/endpoint/buffer.go
luke-lombardi: The tokens are by container, not by gateway - but yes it's true that if a gateway crashes, it wouldn't decrement the tokens so they would hang around until the expiration

pkg/abstractions/common/ring_buffer.go
luke-lombardi: Yeah but it's the same as the fifo implementation was previously. Since it's a ring buffer it would always overwrite the first element when it's full. In practice it won't ever be full because the buffer size is likely much larger than max pending tasks

pkg/abstractions/common/ring_buffer.go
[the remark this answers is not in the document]
luke-lombardi: @jsun-m just checked on this. I think the simplest solution is just to ensure the requestBufferSize is always greater than max pending tasks. I think this works:

```
	requestBufferSize := int(stubConfig.MaxPendingTasks) + 1 // Changed this to add 1
	if requestBufferSize < endpointMinRequestBufferSize {
		requestBufferSize = endpointMinRequestBufferSize
	}
```

pkg/abstractions/endpoint/buffer.go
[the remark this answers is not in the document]
luke-lombardi: Good point - @mernit can you add this to docs somewhere? Not sure what the best place is

PR #606, opened 2024-10-09
Fix: Added middleware and passed request headers down from buffer for websockets

jsun-m:
The TaskMiddleware used the BasicMiddleware class which happened to skip the dispatch function if the type of http connection is a websocket conenction. This adds a new middleware to handle the task lifecycle functionality for just websockets

### DIFF
--- a/sdk/src/beta9/runner/endpoint.py
+++ b/sdk/src/beta9/runner/endpoint.py
@@ -4,17 +4,14 @@ import os
 import signal
 import traceback
 from contextlib import asynccontextmanager
-from dataclasses import dataclass
 from http import HTTPStatus
 from typing import Any, Dict, Optional, Tuple
 
-from fastapi import Depends, FastAPI, HTTPException, Request
+from fastapi import Depends, FastAPI, Request
 from fastapi.responses import JSONResponse, Response
 from gunicorn.app.base import Arbiter, BaseApplication
 from starlette.applications import Starlette
-from starlette.middleware.base import BaseHTTPMiddleware
-from starlette.requests import HTTPConnection
-from starlette.types import ASGIApp, Receive, Scope, Send
+from starlette.types import ASGIApp
 from uvicorn.workers import UvicornWorker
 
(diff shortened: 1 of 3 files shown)


luke-lombardi (changes requested):
Can we move the middleware out to a separate file? Feels like it's getting really messy

sdk/src/beta9/runner/endpoint.py
```diff
@@ -127,51 +128,67 @@ class TaskLifecycleData:
     override_callback_url: Optional[str] = None
 
 
-class TaskLifecycleMiddleware(BaseHTTPMiddleware):
-    async def dispatch(self, request: Request, call_next):
-        if request.url.path == "/health":
-            return await call_next(request)
+async def runTask(request, func, func_args):
```
luke-lombardi, line 131: let's stick to style conventions here `handle_task` maybe

sdk/src/beta9/runner/endpoint.py
```diff
@@ -209,6 +226,7 @@ def __init__(self, logger: logging.Logger, worker: UvicornWorker) -> None:
             self.app = FastAPI(lifespan=self.lifespan)
 
         self.app.add_middleware(TaskLifecycleMiddleware)
+        self.app.add_middleware(WebsocketTaskLifecycleMiddleware)
```
luke-lombardi, line 229: I guess we need to add both - they don't conflict do they? Like we're not going to accidentally be using the websocket middleware for non-websockets. Also another concern is overhead of two layers of middleware, are they both executed (or attempted to be executed) everytime?

jsun-m: both are being executed but it immediately skips the websocket one cause of this
```
if scope["type"] != "websocket":
            await self.app(scope, receive, send)
            return

```

The TaskMiddleware behind the scenes already do
```
if scope["type"] != "http":
            await self.app(scope, receive, send)
            return
```

luke-lombardi (approved):

PR #609, opened 2024-10-10
Fix: fix any gpu scheduling

jsun-m:


dleviminzi (approved):

luke-lombardi (approved):

luke-lombardi (changes requested):
@jsun-m do we need to clean this one up now that we merged failover logic? maybe there's a cleaner way to do it now?

pkg/scheduler/scheduler.go
```diff
@@ -309,8 +309,20 @@ func filterWorkersByResources(workers []*types.Worker, request *types.ContainerR
 
 		if requiresGPU {
 			// Validate GPU resource availability
+
 			priorityModifier, validGpu := gpuRequestsMap[worker.Gpu]
-			if !validGpu || worker.FreeGpuCount < request.GpuCount {
+			if !validGpu {
```
luke-lombardi, line 314: can't we simplify this by saying if requiresGpu && isGpuWorker, we filter it in?

luke-lombardi (changes requested):

PR #610, opened 2024-10-10
Fix: Forward query parameters for endpoints and ASGI apps

dleviminzi:
Resolve BE-1941

Currently we are not forwarding query parameters for asgi apps. I discovered this while hosting a simple test asgi app with the endpoint 

```python
    @myapp.get("/sleep")
    async def sleep(seconds: int):
        time.sleep(seconds)
        return {"message": f"Slept for {seconds} seconds"}
```

Calling `asgi-test-app/v2/sleep?seconds=0` yielded

```json
{"detail":[{"type":"missing","loc":["query","seconds"],"msg":"Field required","input":null,"url":"https://errors.pydantic.dev/2.5/v/missing"}]}
```

With this change I get the expected behavior.


pkg/task/serialize_test.go
```diff
@@ -95,11 +105,76 @@ func TestSerializeHttpPayload(t *testing.T) {
...
+		},
+		{
+			name:        "mix of strings and numbers",
+			queryParams: url.Values{"sleep": []string{"Today", "200.2", "300"}},
+			wantPayload: &types.TaskPayload{
+				Args:   nil,
+				Kwargs: map[string]interface{}{"sleep": []string{"Today", "200.2", "300"}},
+			},
+			wantErr: false,
+		},
```
dleviminzi, line 172: With fastapi, you can define an endpoint like below to take a list of mixed types.  
```python
@app.get("/endpoint")
async def endpoint(stuff: List[Union[int, str]] = Query(None)):
    return {"stuff": stuff}
```

With our endpoints, that will not work currently. Everything will become a string. We could try to create a new type that we then unravel into the correct types in the runner. Alternatively, we can just document this divergence and leave it as is.

pkg/task/serialize.go
```diff
@@ -51,6 +58,42 @@ func SerializeHttpPayload(ctx echo.Context) (*types.TaskPayload, error) {
...
+			if allFloat {
+				taskPayload.Kwargs[key] = floats
+			} else {
+				taskPayload.Kwargs[key] = values
+			}
+		}
+	}
+}
+
+func convertToNumericIfPossible(s string) interface{} {
```
nickpetrovic, line 94: This is probably a personal preference and a nitpick, but I'm not a fan of the "IfPossible" part lol. What about naming this to something like `tryParseNumeric`?

pkg/task/serialize.go
```diff
@@ -51,6 +58,42 @@ func SerializeHttpPayload(ctx echo.Context) (*types.TaskPayload, error) {
...
-	return taskPayload, nil
+func parseRequestArgs(queryParams url.Values, taskPayload *types.TaskPayload) {
+	if taskPayload.Kwargs == nil {
+		taskPayload.Kwargs = make(map[string]interface{})
+	}
+
+	for key, values := range queryParams {
+		if len(values) == 1 {
+			taskPayload.Kwargs[key] = convertToNumericIfPossible(values[0])
+		} else {
```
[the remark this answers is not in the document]
nickpetrovic: If we still need it, it might be cleaner to use `continue` instead of an if/else.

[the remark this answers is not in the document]
dleviminzi: This is checking whether the query parameter is singular or a list. I will change it to use `continue` though e

pkg/task/serialize.go
dleviminzi: yeah that is fair lol. I will change it to `tryParseNumeric`

nickpetrovic (approved):
lgtm. Can we update our example apps (or the integration tests) to validate the use of query params before we go out to prod?

dleviminzi:
> lgtm. Can we update our example apps (or the integration tests) to validate the use of query params before we go out to prod?

yeah, I'm not sure where the integration tests live, but I will update the docs today with a few things

PR #611, opened 2024-10-10
Feat: Add ability to provide backup GPUs which containers can be scheduled on 

jsun-m:
This update introduces a new feature that allows users to specify multiple GPUs for fallback purposes. In scenarios where the primary GPU becomes unavailable due to low capacity or failure, the system will now automatically fallback to the specified alternative GPUs.

#### Key Changes:
•	GPU Parameter Flexibility:
•	The gpu parameter in all abstractions now accepts both a single GPU as a string and a list of GPUs as strings.
•	This change enhances flexibility in GPU management, enabling users to specify a prioritized list of fallback GPUs.

#### Usage Examples:
1.	Single GPU Specification:
```
@function(
    gpu="A10G",
)
def test():
    ...
```
2.	Multiple GPU Specification:
```
@function(
    gpu=["A10G", "T4"]
)
def test():
    ...
```

#### Priority Handling:
The order of GPUs specified in the list determines their priority. The GPU at the earliest index in the list is given the highest priority, allowing for efficient fallback management.

This enhancement aims to improve resource allocation and ensure more reliable GPU utilization within our system.


pkg/scheduler/scheduler.go
```diff
@@ -237,6 +259,7 @@ func (s *Scheduler) scheduleRequest(worker *types.Worker, request *types.Contain
 	go s.schedulerMetrics.CounterIncContainerScheduled(request)
 	go s.eventRepo.PushContainerScheduledEvent(request.ContainerId, worker.Id, request)
 
+	request.ActualGpu = worker.Gpu
```
luke-lombardi, line 262: this is super confusing

pkg/types/scheduler.go
```diff
@@ -67,24 +67,28 @@ type ContainerState struct {
...
+	ContainerId string   `json:"container_id"`
+	EntryPoint  []string `json:"entry_point"`
+	Env         []string `json:"env"`
+	Cpu         int64    `json:"cpu"`
+	Memory      int64    `json:"memory"`
+	Gpu         string   `json:"gpu"`
+	BackupGpus  []string `json:"backup_gpus"`
+	// TODO: Ideally Gpu and BackupGpus should be merged into a single field
+	// However we need to do it in a backward compatible way
+	ActualGpu        string          `json:"actual_gpu"` // The actual GPU that was chosen for the container
```
luke-lombardi, line 79: this field shouldn't be necessary

pkg/abstractions/container/container.go
```diff
@@ -158,12 +159,21 @@ func (cs *CmdContainerService) ExecuteCommand(in *pb.CommandExecutionRequest, st
 		gpuCount = 1
 	}
 
+	gpuTypes := strings.Split(string(stubConfig.Runtime.Gpu), ",")
```
luke-lombardi, line 162: can you consolidate this logic into a helper function or something?

luke-lombardi (changes requested):

pkg/scheduler/scheduler.go
```diff
@@ -161,23 +161,37 @@ func (s *Scheduler) Stop(containerId string) error {
...
 
-func (s *Scheduler) getController(request *types.ContainerRequest) (WorkerPoolController, error) {
-	var ok bool
-	var workerPool *WorkerPool
+func (s *Scheduler) getControllers(request *types.ContainerRequest) ([]WorkerPoolController, error) {
+	controllers := []WorkerPoolController{}
 
 	if request.PoolSelector != "" {
-		workerPool, ok = s.workerPoolManager.GetPool(request.PoolSelector)
+		wp, ok := s.workerPoolManager.GetPool(request.PoolSelector)
```
jsun-m, line 168: will look into simplifying this.

luke-lombardi:
@jsun-m I'm still reviewing this, but can you rename this PR and add more context to the PR description?

pkg/abstractions/container/container.go
```diff
@@ -154,16 +154,21 @@ func (cs *CmdContainerService) ExecuteCommand(in *pb.CommandExecutionRequest, st
 	env = append(secrets, env...)
 
 	gpuCount := 0
-	if stubConfig.Runtime.Gpu != "" {
+	if len(stubConfig.Runtime.Gpus) > 0 {
 		gpuCount = 1
 	}
 
+	gpuRequest := types.GpuTypesToStrings(stubConfig.Runtime.Gpus)
+	if stubConfig.Runtime.Gpu != "" {
```
luke-lombardi, line 162: this is the backwards compatibility check correct?

pkg/repository/container_redis.go
```diff
@@ -161,6 +161,39 @@ func (cr *ContainerRedisRepository) UpdateContainerStatus(containerId string, st
 	return nil
 }
 
+func (cr *ContainerRedisRepository) UpdateContainerGPU(containerId string, gpuType string) error {
```
luke-lombardi, line 164: maybe UpdateAssignedContainerGPU?

pkg/types/scheduler.go
```diff
@@ -74,18 +74,19 @@ type ContainerRequest struct {
...
 	ImageId          string          `json:"image_id"`
 	StubId           string          `json:"stub_id"`
 	WorkspaceId      string          `json:"workspace_id"`
 	Workspace        Workspace       `json:"workspace"`
-	Stub             StubWithRelated `json:"stub"`
 	Timestamp        time.Time       `json:"timestamp"`
 	Mounts           []Mount         `json:"mounts"`
 	RetryCount       int             `json:"retry_count"`
 	PoolSelector     string          `json:"pool_selector"`
+	Stub             StubWithRelated `json:"stub"`
```
luke-lombardi, line 89: why did you move this? I thought it sort of made sense next to Workspace

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -272,6 +272,14 @@ def _sync_content(
         except Exception as e:
             terminal.warn(str(e))
 
+    def _parse_gpu(self, gpu: Union[GpuTypeAlias, List[GpuTypeAlias]]) -> str:
```
luke-lombardi, line 275: does this handle type assertions / validation on the input?

pkg/scheduler/scheduler.go
```diff
@@ -252,12 +278,35 @@ func filterWorkersByPoolSelector(workers []*types.Worker, request *types.Contain
...
+		}
+
+		if len(gpuRequestsMap) > 0 {
+			// Validate GPU resource availability
+			priorityModifier, validGpu := gpuRequestsMap[worker.Gpu]
+			if !validGpu || worker.FreeGpuCount < request.GpuCount {
+				continue
+			}
+
+			worker.Priority -= int32(priorityModifier) // This will account for the preset priorities for the pool type as well as the order of the GPU requests
```
luke-lombardi, line 306: hm interesting. will this always work? i think it'll only work if priorities are just increasing by 1?

luke-lombardi (changes requested):

pkg/abstractions/container/container.go
jsun-m: yes

pkg/types/scheduler.go
jsun-m: not sure why i moved it. moving back

pkg/scheduler/scheduler.go
jsun-m: right for now it will work well if priorities are incremental. Once we think of more complex priorities, we would need to update this scoring function. But this additive/subtractive behavior works well with the default priorities.

sdk/src/beta9/abstractions/base/runner.py
jsun-m: yes

pkg/repository/container_redis.go
jsun-m: done

luke-lombardi (changes requested):
Can you add a note about that here?

pkg/scheduler/scheduler.go
```diff
@@ -161,23 +161,36 @@ func (s *Scheduler) Stop(containerId string) error {
 	return nil
 }
 
-func (s *Scheduler) getController(request *types.ContainerRequest) (WorkerPoolController, error) {
-	var ok bool
-	var workerPool *WorkerPool
+func (s *Scheduler) getControllers(request *types.ContainerRequest) ([]WorkerPoolController, error) {
+	controllers := []WorkerPoolController{}
+	combinedRequestedGpu := request.GpuRequest
```
luke-lombardi, line 166: Can you come up with a better name for this car

luke-lombardi (changes requested):
two more comments

luke-lombardi (approved):

PR #613, opened 2024-10-11
Feat: Allow concurrent requests to ASGI apps

dleviminzi:
Resolve BE-1919

Previously running the following two curls in order would result in the second request taking 10 seconds because it waits until the first request is handled. With this change, a user can now specify `concurrent_requests` for their asgi application. In the example below, setting this to 2 allows the app to receive and process the second request while the first is still sleeping

```bash
# first
curl -X POST 'http://localhost:1994/asgi/asgi-test-app/latest/sleep-with-body' \
-H 'Connection: keep-alive' \
-H 'Content-Type: application/json' \
-H 'Authorization: Bearer <token>' \
-d '{"seconds": 10, "nums": [1, 2, 3]}'
```
```
# second
curl -X POST 'http://localhost:1994/asgi/asgi-test-app/latest/sleep-with-body' \
-H 'Connection: keep-alive' \
-H 'Content-Type: application/json' \
-H 'Authorization: Bearer <token>' \
-d '{"seconds": 0, "nums": [1, 2, 3]}'
```

Endpoint being called above
```python
    @myapp.post("/sleep-with-body")
    async def sleep_with_body(body: dict = Body(...)):
        await asyncio.sleep(body["seconds"])
        return {"message": f"Slept for {body['seconds']} seconds", "nums": body["nums"]}
```


sdk/src/beta9/abstractions/endpoint.py
```diff
@@ -117,6 +117,8 @@ def __init__(
         callback_url: Optional[str] = None,
         task_policy: TaskPolicy = TaskPolicy(),
     ):
+        # Default to 1 concurrent requests for endpoints
+        self.concurrent_requests = 1
```
jsun-m, line 121: do we need this? Seems like the && isASGI will just ignore this anyways

jsun-m (changes requested):

sdk/src/beta9/abstractions/endpoint.py
dleviminzi: Yeah this is actually in the wrong place, but we do need to set the default. `ASGI` calls the `Endpoint` init method and overrides the concurrent_requests with whatever it was passed. For endpoints we don't want to expose the setting in the decorator.

sdk/src/beta9/abstractions/endpoint.py
```diff
@@ -98,6 +98,8 @@ def multiply(**inputs):
         ```
     """
 
+    concurrent_requests = 1
```
jsun-m, line 101: I think i was wondering if we even needed to initialize for endpoints. Seems unnecessary

dleviminzi: If we don't initialize then when we are deploying an endpoint we get an attribute not found error because we pass `self.concurrent_requests` into the runner's `init`.

jsun-m: doesnt it get initialize here with a default?
https://github.com/beam-cloud/beta9/pull/613/files#diff-330a31c80611143ecc706c48c89d09bd4f73e343ae53bcfa4a606048da630b46R102

dleviminzi: It won't ever reach that line because [this line](https://github.com/beam-cloud/beta9/pull/613/files#diff-7925030b37f47ede4bed18d7a45561a55ce011de54bbb098e0437c2bfe0a915fR140) will cause a missing attribute error. 

For ASGI apps it won't cause a missing attribute error because it has a default value in the ASGI `init` method. However, we don't want this to be an option on the Endpoint decorator, so there is no default set in the `init` because its not in the `init` method.  

By setting this default in the Endpoint, we guarantee that `concurrent_requests` is an attribute that can be passed into the runner's init method. When it is set in the ASGI `init`, it overrides this default value.

jsun-m: can this line be 
```
concurrent_requests=1,
```

instead of 
```
concurrent_requests=self.concurrent_requests,
```

since endpoints doesnt care too much about it?

dleviminzi: We can't set that because then L256 `self.concurrent_requests = concurrent_requests` will have no effect. The value of `concurrent_requests` passed into the ASGI `init` method will just be overwritten with `1` every time  because ASGI init -> Endpoint init -> Runner init. 

Basically, by using `self.concurrent_requests`, we are able to override that value in the ASGI `init`, but it will still always be `1` for endpoint.

jsun-m: hmm ok

jsun-m (approved):

sdk/src/beta9/abstractions/endpoint.py
```diff
@@ -166,24 +252,27 @@ def __init__(
         name: Optional[str] = None,
         authorized: bool = True,
         autoscaler: Autoscaler = QueueDepthAutoscaler(),
+        task_policy: TaskPolicy = TaskPolicy(),
```
luke-lombardi, line 255: do we want this here @jsun-m @dleviminzi ? Retries don't work for endpoints so might not be worth exposing

luke-lombardi (changes requested):
one small comment, otherwise looks good

sdk/src/beta9/abstractions/endpoint.py
dleviminzi: whoops removed!

luke-lombardi: 👍

luke-lombardi (approved):

PR #621, opened 2024-10-15
Fix: Increase wait on logs as well as split logs in worker stdout

jsun-m:


pkg/abstractions/image/build.go
```diff
@@ -299,7 +299,7 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
 				errMsg = err.Error() + "\n"
 			}
 
-			time.Sleep(1 * time.Second) // Wait for logs to be passed through
```
jsun-m, line 302: This is not long enough if the user's connection is slow due to internet and/or distance from our datacenter

pkg/worker/logger.go
```diff
@@ -103,10 +103,12 @@ func (r *ContainerLogger) CaptureLogs(containerId string, outputChan chan common
 				instance.LogBuffer.Write([]byte(msg.Message))
 			}
 
-			if msg.TaskID != nil && msg.Message != "" {
-				log.Printf("<%s>:<%s> - %s\n", containerId, *msg.TaskID, msg.Message)
-			} else if msg.Message != "" {
-				log.Printf("<%s> - %s\n", containerId, msg.Message)
+			for _, line := range strings.Split(msg.Message, "\n") {
```
jsun-m, line 106: This doesn't affect user code but it affects the logs captured by fluent-bit from workers for `snoop`. Some lines won't have container-ids since if they had a newline so we must split the multiline logs and add them ourselves. Shouldn't increase # of ES entries there are

luke-lombardi (approved):

Makefile
```diff
@@ -47,10 +47,10 @@ runner:
 	done
 
 start:
-	cd hack && okteto up --file okteto.yml
```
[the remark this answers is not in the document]
jsun-m: yea

PR #628, opened 2024-10-17
Feat: Add max TTL to map values of 7 days and add max size to payload

nickpetrovic:
- Allow for a TTL (in seconds) to be set on a map key
- Set a max TTL of 7 days (also use as default TTL)
- Set a max map value size of 1MiB
- Add docstring to map.set function

Resolve BE-1953


sdk/src/beta9/abstractions/map.py
```diff
@@ -69,11 +69,28 @@ def stub(self) -> MapServiceStub:
...
+            key: The key to set.
+            value: The value to set.
+            ttl: The time to live for the key in seconds. Cannot be more than 7 days. Defaults to 7 days.
+
+        Raises:
+            ValueError: If the key could not be set.
+
+        Returns:
+            bool: True if the key was set, False otherwise.
+        """
```
dleviminzi, line 86: I think we should mention the payload size limit here

nickpetrovic: Good call. Just updated.

dleviminzi (approved):
Looks good, but I do think its a little odd that the default TTL is equal to the max TTL. Are we planning on lowering the default at a later date or something?

pkg/abstractions/map/map_redis.go
```diff
@@ -26,7 +32,15 @@ func NewRedisMapService(rdb *common.RedisClient) (MapService, error) {
 func (m *RedisMapService) MapSet(ctx context.Context, in *pb.MapSetRequest) (*pb.MapSetResponse, error) {
 	authInfo, _ := auth.AuthInfoFromContext(ctx)
 
-	err := m.rdb.Set(ctx, Keys.MapEntry(authInfo.Workspace.Name, in.Name, in.Key), in.Value, 0).Err()
+	if len(in.Value) > maxMapValueSize {
+		return &pb.MapSetResponse{Ok: false, ErrMsg: "Value cannot be larger than 1 MiB"}, nil
+	}
+
+	if time.Duration(in.Ttl)*time.Second > maxMapValueTtls {
+		return &pb.MapSetResponse{Ok: false, ErrMsg: "TTL cannot be larger than 1 week"}, nil
```
luke-lombardi, line 40: Change to say longer than 1 week probably

luke-lombardi (changes requested):

pkg/abstractions/map/map_redis.go
nickpetrovic: Updated.

nickpetrovic:
> Looks good, but I do think its a little odd that the default TTL is equal to the max TTL. Are we planning on lowering the default at a later date or something?

I'm not sure. I think we wanted to play it safe with the users currently using it. Could be that we increase the max value if we end up using another backend to store these. I don't know if redis is a good long term solution given the limited amount of memory.

PR #634, opened 2024-10-17
Feat: Add worker version to worker list command

nickpetrovic:
Resolve BE-1939


pkg/common/redis.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/beam-cloud/beta9/pkg/types"
 	"github.com/bsm/redislock"
+	"github.com/mitchellh/mapstructure"
```
dleviminzi, line 17: Can we pull in this one instead? https://pkg.go.dev/github.com/go-viper/mapstructure/v2

The original one is no longer maintained. Viper is a very active project so it should be in good.

pkg/repository/worker_redis.go
```diff
@@ -310,18 +310,19 @@ func (r *WorkerRedisRepository) UpdateWorkerCapacity(worker *types.Worker, reque
...
+	currentWorker, err := r.getWorkerFromKey(key)
 	if err != nil {
 		return fmt.Errorf("failed to get worker state <%v>: %v", key, err)
 	}
 
+	sourceWorker := currentWorker // worker from the Redis store
+	if sourceWorker == nil {
+		sourceWorker = worker // worker from the argument
+	}
+
```
dleviminzi, line 322: I think you can just declare sourceWorker in the assignment such that L318 isn't needed

dleviminzi (approved):
Looks good! Left two minor comments

pkg/repository/worker_redis.go
nickpetrovic: We did something in v1 where this logic needed to be this way. Could have been an oversight, but I don't think I want to rock the boat on this too much. I just wanted to avoid having to write and return in CopyStruct twice.

PR #637, opened 2024-10-18
Add more tests for CopyStruct

nickpetrovic:


pkg/common/redis_test.go
```diff
@@ -87,12 +87,133 @@ func TestRedisLockWithTTLAndRetry(t *testing.T) {
...
+				Status:       types.WorkerStatusPending,
+				TotalCpu:     20,
+				TotalMemory:  0,
+				BuildVersion: "0.1.0",
+			},
+			expected: &types.Worker{
+				Id:                   "123",
+				Status:               types.WorkerStatusAvailable,
+				TotalCpu:             10,
+				TotalMemory:          0,
```
luke-lombardi, line 145: nit-pick but i think it would be worth making all of these non-zero values

luke-lombardi (changes requested):

pkg/common/redis_test.go
```diff
@@ -87,12 +87,133 @@ func TestRedisLockWithTTLAndRetry(t *testing.T) {
...
+			},
+		},
+		{
+			name: "copy struct overwriting fields on destination part 2",
+			src: &types.Worker{
+				Id:       "123",
+				Status:   types.WorkerStatusAvailable,
+				TotalCpu: 10,
+			},
+			dest: &types.Worker{
```
luke-lombardi, line 134: @nickpetrovic can you add a test for machine state as well

pkg/common/redis_test.go
dleviminzi: This is a good library if you want to automate creating a bunch of fake structs a bit: https://github.com/go-faker/faker

nickpetrovic: Yeah good call. The test after this tests for defaults. I've updated this.

pkg/common/redis_test.go
nickpetrovic: Added.

PR #641, opened 2024-10-18
Fix: Add lock in sdk when creating stub to prevent races when using multiple threads

dleviminzi:
Below you can see the get or create stub request to the gateway fail. This occurs because multiple requests are made with the same object id. Whichever call extracts the object first will succeed and any others will fail (assuming they go down the path of creating the stub). To fix this problem, I added a thread lock to the `get_or_create_stub` call in `prepare_runtime`. This guarantees that only one of the threads will try to create stub and the remaining threads will just retrieve the previously created stub instead of potentially trying to create one themselves. 
```bash
=> Building image 
=> Building image 
=> Building image 
=> Using cached image 
=> Syncing files 
=> Using cached image 
Reading .beta9ignore file
=> Using cached image 
Collecting files from /Users/minzi/Dev/beam/baps/thread-fn
Added /Users/minzi/Dev/beam/baps/thread-fn/lalal
Added /Users/minzi/Dev/beam/baps/thread-fn/app.py
Added /Users/minzi/Dev/beam/baps/thread-fn/bloooper
Collected object is 969.00 B
=> Uploading 
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 969/969 bytes 0:00:00
=> Files synced 
=> Files already synced 
=> Files already synced 
Failed to get or create stub
=> Running function: <app:f1> 
=> Running function: <app:f3> 
f1
f3
=> Function complete <798bb012-4f93-4b13-9137-05256d1c3f48> 
=> Function complete <bd558c36-f538-420a-b631-dc6821c49309> 
[None, 'f1', 'f3']
```

### DIFF
--- a/sdk/src/beta9/abstractions/base/runner.py
+++ b/sdk/src/beta9/abstractions/base/runner.py
@@ -2,6 +2,7 @@ import inspect
 import os
 import sys
 import tempfile
+import threading
 import time
 from queue import Empty, Queue
 from typing import Callable, List, Optional, Union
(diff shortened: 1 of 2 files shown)


sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -52,6 +53,20 @@
 ASGI_SERVE_STUB_TYPE = "asgi/serve"
 FUNCTION_SERVE_STUB_TYPE = "function/serve"
 
+_stub_creation_lock = threading.Lock()
+_stub_created_for_workspace = False
+
+
+def _get_stub_created_for_workspace() -> bool:
```
nickpetrovic, line 60: I think using `_is_` vs `_get_` will make a better name for this.

nickpetrovic (approved):
I almost wonder if there's a better way to handle multiple locks. If more locks are required later, we might want to see about trying something else.

Added one comment on naming convention, otherwise I think this looks good.

nickpetrovic:
@dleviminzi just noticed `Building image` is being printed multiple times. Could this be an issue in the future?

dleviminzi:
> @dleviminzi just noticed `Building image` is being printed multiple times. Could this be an issue in the future?

hmm that is a good point. I will investigate.

PR #643, opened 2024-10-20
Feat: Realtime abstraction built on top of asgi

luke-lombardi:
This is an experimental decorator that configures a basic websocket streaming example under the hood. The idea is to have a drop-in-place feature that lets users create streaming handlers that can do image generation, chat, etc. without writing the ASGI boilerplate.

**TODO**:
- [x] Add automatic type inference
- [ ] Create client examples

```
from beta9 import realtime


def load_model():
    return "my_model"


@realtime(
    cpu=1,
    memory=128,
    timeout=180,
    concurrent_requests=10,
    on_start=load_model,
    authorized=False,
)
def handler(*, context, input: str):
    print(context.on_start_value)
    return input

```

The output from this echo server looks like this:
```
luke@Lukes-MBP beta9 % websocat ws://localhost:1994/asgi/public/bbfb4433-9685-4f79-9d92-68fb57b240c3
do something
do something
echo
echo
here
here
```

### DIFF
--- a/sdk/src/beta9/abstractions/endpoint.py
+++ b/sdk/src/beta9/abstractions/endpoint.py
@@ -334,7 +334,6 @@ class RealtimeASGI(ASGI):
             try:
                 while True:
                     data = await websocket.receive_text()
-                    websocket.send_bytes
                     internal_asgi_app.input_queue.put(data)
 
                     if not internal_asgi_app.input_queue.empty():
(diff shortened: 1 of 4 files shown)


mernit (approved):
Nice

sdk/src/beta9/abstractions/endpoint.py
```diff
@@ -276,6 +279,191 @@ def __init__(
...
+        callback_url (Optional[str]):
+            An optional URL to send a callback to when a task is completed, timed out, or cancelled.
+    Example:
+        ```python
+        from beta9 import realtime
+
+        def generate_text():
+            return ["this", "could", "be", "anything"]
+
+        @realtime(
```
dleviminzi, line 348: Are you against just calling this abstraction "websocket"? In my mind its a bit more direct and potentially easier for users to find in the sdk.

luke-lombardi: I'm open to it, let's talk about it tomorrow. It's a very particular use case which is like "process frame" from queue, and keep going vs. some more advanced websocket use cases that the name `websocket` might imply

nickpetrovic (approved):
lgtm. We can rename it or alias it to websocket if we want. Either is fine to me though.

PR #645, opened 2024-10-21
Message on fn quit and update b9ignore

dleviminzi:
Resolve BE-1966

Now, when a user ctrl-c while a remote function invocation is under way they see the following: 

```python
~/Dev/beam/baps/slow-fn main ❯ python app.py                                                                                                                                 beta9-py3.12
=> Building image
=> Using cached image
=> Syncing files
Reading .beta9ignore file
Collecting files from /Users/minzi/Dev/beam/baps/slow-fn
Added /Users/minzi/Dev/beam/baps/slow-fn/pyproject.toml
Added /Users/minzi/Dev/beam/baps/slow-fn/app.py
Collected object is 646.00 B
=> Files already synced
=> Running function: <app:slow>
Application interrupted. Remote function is still running.
```

Also, the ignore file has been updated to ignore everything aside from a bunch of (I think) sensible file types


mernit:
can we change the text to say: `Exiting Shell. Your function will continue running remotely.`

sdk/src/beta9/sync.py
```diff
@@ -63,6 +81,12 @@ def get_workspace_object_id() -> str:
 .ruff_cache
 .dockerignore
 .ipynb_checkpoints
+.env.local
+.envrc
+**/__pycache__/
+**/.pytest_cache/
+**/node_modules/
+*.pyc
```
nickpetrovic, line 89: If we negate, do we even need any rules after the last rule on line 66?

sdk/src/beta9/abstractions/function.py
```diff
@@ -147,8 +148,12 @@ def __call__(self, *args, **kwargs) -> Any:
...
             return
 
-        with terminal.progress("Working..."):
-            return self._call_remote(*args, **kwargs)
+        try:
+            with terminal.progress("Working..."):
+                return self._call_remote(*args, **kwargs)
+        except KeyboardInterrupt:
+            terminal.error("Exiting Shell. Your function will continue running remotely.")
+            os._exit(0)
```
nickpetrovic, line 156: `terminal.error` will call `sys.exit(1)`.

Just wondering what the behavior would be if we just did something like `return terminal.error("...")`. Does it exit cleanly and quickly?

nickpetrovic (changes requested):

sdk/src/beta9/sync.py
dleviminzi: I left these here because I wasn't sure what would happen if some of these folders contained allowed file types. For instance, .vscode will have some json in it

sdk/src/beta9/abstractions/function.py
dleviminzi: I removed the exit call. Still exists cleanly. The `terminal.error` call doesn't return anything so I'm just calling it, not returning it

nickpetrovic: Got it. I usually try returning because the linter doesn't know its exiting and will make some incorrect assumptions about other vars and what not. Since this is the end of the function, it should be fine.

PR #646, opened 2024-10-21
Feat: Allow Worker to shutdown when signaled with SIGTERM

nickpetrovic:
## Summary

This PR mostly does 2 things.

- Move code around so worker.go isn't over 1k lines long
- Allow for the worker to terminate when a SIGTERM is received. This is done by having goroutines use select with ctx.Done to complete their loops and stopping (SIGKILL) all containers in the shutdown function.

## Changes

- Move logic in processStopContainerEvents func to stopContainer func so stopContainer can be called elsewhere
- When calling stopContainer, do not check for container exit code. Container could still be running, terminate it anyway.
- Added listenForShutdown
- Forcefully stop all containers in the shutdown function. If we've gotten this far, containers should not be running otherwise unmounting clip images will hang.
- Moved calling of cancel() above stopping containers in shutdown. This stops all goroutines from running before stopping containers.
- Shutdown worker if worker is not found while in manage capacity goroutine. Usually thousands of "unable to process completed request" messages would be printed. This stops that.
- Change DeleteContainerState so it receives a containerId vs a ContainerRequest - we only used the containerId.
- Move container-specific and helper functions to lifecycle.go
- Rename terminateContainer to finalizeContainer to confusion with stopContainer func
- Remove containerErr var since it wasn't being used
- Move CPU and Memory Linux resource calculations to runc_resources.go

Resolve BE-1954

### DIFF
--- a/pkg/repository/base.go
+++ b/pkg/repository/base.go
@@ -45,7 +45,7 @@ type ContainerRepository interface {
 	SetContainerAddress(containerId string, addr string) error
 	GetContainerAddress(containerId string) (string, error)
 	UpdateContainerStatus(string, types.ContainerStatus, time.Duration) error
-	DeleteContainerState(*types.ContainerRequest) error
+	DeleteContainerState(containerId string) error
 	SetWorkerAddress(containerId string, addr string) error
 	SetContainerStateWithConcurrencyLimit(quota *types.ConcurrencyLimit, request *types.ContainerRequest) error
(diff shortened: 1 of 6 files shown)


nickpetrovic:
@luke-lombardi starting to think we should've moved the code in a follow up PR. Let me know if this is too much to read.

luke-lombardi (approved):

PR #647, opened 2024-10-23
Feat: Add a public token type that has only READ access to calling deployments

jsun-m:
Added a new `token_type` called `public` that gives READONLY for making HTTP requests to deployments, but prevents access to other resources.

This will allow our users to safely distribute Beta9 api tokens to their own users


pkg/types/backend.go
```diff
@@ -28,6 +28,7 @@ const (
 	TokenTypeWorkspace    string = "workspace"
 	TokenTypeWorker       string = "worker"
 	TokenTypeMachine      string = "machine"
+	TokenTypePublic       string = "public" // TODO: naming could be improved
```
luke-lombardi, line 31: what should we name it?

luke-lombardi (changes requested):

pkg/types/backend.go
nickpetrovic: I think `TokenTypeReadOnly` because its being used as just that. If we decide to have permissions on tokens, then having a "Public" token is redundant with a "Workspace" token. So this feels more like a temporary feature to me until we have something more elaborate. 

If we want to get more specific, `TokenTypeDeploymentReadOnly`. This creates room for other resource types like tasks or volumes, but it could make the auth closure logic more complicated to follow, challenging to add more token types, and difficult to maintain.

jsun-m: what about `TokenTypeRequestOnly`?

jsun-m: or `TokenTypeCallAPIOnly`

nickpetrovic: I think those are less clear. Neither "Request" nor "CallAPIOnly" indicate to me that this is a read-only token. In both names, one could still assume they have write permissions.

jsun-m: `TokenTypeDeploymentReadOnly` sounds like they can list deployments as well. I can't think of the perfect name for this.

what about `TokenTypeAPIRestricted`

PR #652, opened 2024-10-23
Feat: Conda support via micromamba base images

dleviminzi:
This PR adds support for conda via micromamba base images that we build.  

You can now create apps that use the anaconda ecosystem like this:

```python
from beta9 import Image, function


@function(
    image=Image(python_version="python3.11")
    .micromamba()
    .add_micromamba_packages(["pandas", "numpy"])
    .add_python_packages(["triton"]) # pip installs still works in conda environment
)
def func():
    import sys

    import numpy as np
    import pandas as pd
    import triton

    return {
        "python_version": sys.version,
        "numpy_version": np.__version__,
        "pandas_version": pd.__version__,
        "triton_version": triton.__version__,
    }


if __name__ == "__main__":
    print(func.remote())
```


pkg/abstractions/image/image.go
```diff
@@ -55,20 +57,23 @@ func NewRuncImageService(
...
-		ExistingImageUri:  in.ExistingImageUri,
+		BaseImageTag:       baseImageTag,
+		BaseImageName:      is.config.ImageService.Runner.BaseImageName,
+		BaseImageRegistry:  is.config.ImageService.Runner.BaseImageRegistry,
+		PythonVersion:      in.PythonVersion,
+		PythonPackages:     in.PythonPackages,
+		Commands:           in.Commands,
+		BuildSteps:         convertBuildSteps(in.BuildSteps),
+		ExistingImageUri:   in.ExistingImageUri,
+		Micromamba:         in.Micromamba,
```
luke-lombardi, line 75: Can we add this to buildsteps instead of adding additional parameters for mamba?

pkg/abstractions/image/image.go
```diff
@@ -100,21 +107,33 @@ func (is *RuncImageService) BuildImage(in *pb.BuildImageRequest, stream pb.Image
 		BuildSteps:         convertBuildSteps(in.BuildSteps),
 		ExistingImageUri:   in.ExistingImageUri,
 		ExistingImageCreds: in.ExistingImageCreds,
+		Micromamba:         in.Micromamba,
```
luke-lombardi, line 110: same comment as above

pkg/abstractions/image/image.go
```diff
@@ -139,3 +158,17 @@ func convertBuildSteps(buildSteps []*pb.BuildStep) []BuildStep {
 	}
 	return steps
 }
+
+func (is RuncImageService) getBaseImageTag(pythonVersion string, micromambaEnabled bool) (string, bool) {
```
luke-lombardi, line 162: feels weird to parameterize the function like this, is there a more elegant way we can map to the tags?

luke-lombardi (changes requested):

pkg/abstractions/image/image.go
dleviminzi: I ended up removing these fields and setting the python version to micromamba in the sdk. That can be used in place of this flag. Also, channels can just be added as build steps to be parsed as flags, so I did that as well.

pkg/abstractions/image/image.go
dleviminzi: In cleaning up the above, I got rid of this

pkg/abstractions/image/build.go
```diff
@@ -504,6 +463,50 @@ func ExtractImageNameAndTag(imageRef string) (BaseImage, error) {
...
+	if len(flagLines) > 0 {
+		command += " " + strings.Join(flagLines, " ")
+	}
+	if len(packages) > 0 {
+		command += " " + strings.Join(packages, " ")
+	}
+
+	return command
+}
+
```
dleviminzi, line 509: None of this changed. I moved it because it was a method on the BuildService, but didn't need to be.

pkg/abstractions/image/build.go
```diff
@@ -512,3 +515,75 @@ func hasAnyPrefix(s string, prefixes []string) bool {
...
+		} else {
+			packages = append(packages, fmt.Sprintf("%q", pkg))
+		}
+	}
+	return flagLines, packages
+}
+
+// Generate the commands to run in the container. This function will coalesce pip and mamba commands
+// into a single command if they are adjacent to each other.
+func generateCommands(buildSteps []BuildStep, pythonVersion string) []string {
```
nickpetrovic, line 536: Can we add a test for this? Want to ensure we're grouping adjacent commands and not all of the commands so that we can ensure the ordering is still intact.

pkg/abstractions/image/image.go
```diff
@@ -90,8 +91,10 @@ func (is *RuncImageService) VerifyImageBuild(ctx context.Context, in *pb.VerifyI
 func (is *RuncImageService) BuildImage(in *pb.BuildImageRequest, stream pb.ImageService_BuildImageServer) error {
 	log.Printf("incoming image build request: %+v", in)
 
+	baseImageTag := is.config.ImageService.Runner.Tags[in.PythonVersion]
```
nickpetrovic, line 94: This doesn't look right... I thought we fixed this already? Shouldn't we check to see if the string is empty or if the ok/exists value is false?

docker/Dockerfile.runner
```diff
@@ -195,3 +195,76 @@ rm -rf /usr/lib/python2.7 && rm -rf /usr/lib/python3.6
...
+rm -rf /usr/share/man
+rm -rf /usr/share/locale
+rm -rf /root/.cache/*
+rm -rf /tmp/*
+EOT
+
+
+VOLUME ["/volumes", "/snapshot"]
+
+# Micromamba 3.10
```
nickpetrovic, line 250: Nitpick probably, but is "3.10" is the python version in this? Can we remove that? Just better to focus on maintaining the build arg and not the comment.

dleviminzi: Oh yeah I should remove that

pkg/abstractions/image/image.go
dleviminzi: I think we don't check here because we already checked in the verify above. I just followed what is in the main branch: https://github.com/beam-cloud/beta9/blob/main/pkg/abstractions/image/image.go#L94

pkg/abstractions/image/build.go
dleviminzi: Yeah I'll write up some tests

pkg/abstractions/image/image.go
nickpetrovic: Oh right. Its on the API side of things. So why did we change this at all then?

dleviminzi: Yeah I should revert. I had different logic and then didn't put it back in the right place when I undid all that

Makefile
```diff
@@ -42,9 +42,13 @@ proxy:
 
 runner:
 	for target in py312 py311 py310 py39 py38; do \
-		docker build . --no-cache --target $$target --platform=linux/amd64 -f ./docker/Dockerfile.runner -t localhost:5001/beta9-runner:$$target-$(runnerTag); \
+		docker build . --target $$target --platform=linux/amd64 -f ./docker/Dockerfile.runner -t localhost:5001/beta9-runner:$$target-$(runnerTag); \
 		docker push localhost:5001/beta9-runner:$$target-$(runnerTag); \
 	done
+	for version in "3.12" "3.11" "3.10" "3.9" "3.8"; do \
```
luke-lombardi, line 48: I think we need to push these to ECR in the github action. might be worth manually pushing it first

luke-lombardi (approved):
I think this is ready to merge 👍

PR #656, opened 2024-10-24
Fix: Bug with scheduling logic for CPU workers

jsun-m:


pkg/scheduler/scheduler.go
```diff
@@ -283,8 +283,14 @@ func filterWorkersByResources(workers []*types.Worker, request *types.ContainerR
 		gpuRequestsMap[gpu] = index
 	}
 
+	requiresGPU := len(gpuRequestsMap) > 0
+
 	filteredWorkers := []*types.Worker{}
 	for _, worker := range workers {
+		if requiresGPU && worker.Gpu == "" {
+			continue
```
luke-lombardi, line 291: I still think this is unclear. This handles the case where you're skipping workers where there is no GPU if the request requires a GPU, but it's not clear what happens in the reverse case - that you don't require a GPU and the worker has a GPU. I'd make that explicit

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/taskqueue.py
```diff
@@ -43,7 +43,7 @@ class TaskQueue(RunnerAbstraction):
             The type or name of the GPU device to be used for GPU-accelerated tasks. If not
             applicable or no GPU required, leave it empty.
             You can specify multiple GPUs by providing a list of GpuTypeAlias. If you specify multiple
-            GPUs, the container will load balance across them with equal priority.
+            GPUs, the earlier GPUs will be prioritized over the later ones.
```
luke-lombardi, line 46: earlier implies time, this is just based on the _order_ of the list

pkg/scheduler/scheduler.go
```diff
@@ -295,7 +297,11 @@ func filterWorkersByResources(workers []*types.Worker, request *types.ContainerR
 			continue
 		}
 
-		if len(gpuRequestsMap) > 0 {
+		if requiresGPU && worker.Gpu == "" {
+			// If the worker doesn't have a GPU and the request requires one, skip
```
luke-lombardi, line 301: i would add an explicit else if !requiresGPU && worker.Gpu != ""

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #659, opened 2024-10-25
Experiment: Use uv+pip to install packages during build

dleviminzi:
When building the image for the app below on the main branch (using pip) it currently takes 31 seconds. Building the same image using [uv](https://github.com/astral-sh/uv) takes 14s. This might be controversial because we are adding a dependency instead of just using the base python tooling. However, the speed up is pretty impressive in my opinion and given the tweak -> build -> tweak -> build nature of users' applications I think the improvement will be felt. 

There is also an additional optimization bundled in that could be separated out. I changed the build step command processing so that it resolves all dependencies at once instead of doing a bunch of individual pip installs. However, if somebody had two different `add_python_packages`, my current implementation of this optimization would break. If we like it, I'll make a more robust version that allows for multiple groups of packages. 

```python
from beta9 import endpoint, Image

@endpoint( image=Image()
    .add_commands(
        ["echo 4"] # trigger new builds
    )
    .add_python_packages(
        ["numpy", "pandas", "seaborn", "transformers", "matplotlib", "scipy", "scikit-learn", "opencv-python"]
    ),
    name="slow-build"
)
def handler():
    return "hello"
```

One other change you might notice is the addition of `imageArchitectures` in the config. The primary reason I added this was so that we could use arm runner images when developing locally (they are much faster). I added an optional flag to the makefile so that you can decide which architecture to build the runner with.


pkg/abstractions/image/build.go
```diff
@@ -282,14 +282,19 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
...
 		case shellCommandType:
 			opts.Commands = append(opts.Commands, step.Command)
 		case pipCommandType:
-			opts.Commands = append(opts.Commands, b.generatePipInstallCommand([]string{step.Command}, opts.PythonVersion))
+			opts.Commands = append(opts.Commands, "")
+			pipIndex = len(opts.Commands) - 1
+			pipCommands = append(pipCommands, step.Command)
 		}
 	}
+	opts.Commands[pipIndex] = b.generatePipInstallCommand(pipCommands)
```
nickpetrovic, line 297: One of the benefits of `add_python_packages` was to be able to install packages before or after shell commands due to external dependency requirements. If we resolve and install all python packages in one go, this defeats the purpose of having an ordered list of commands.

pkg/abstractions/image/build.go
```diff
@@ -451,7 +456,7 @@ func (b *Builder) generatePipInstallCommand(pythonPackages []string, pythonVersi
 		}
 	}
 
-	command := fmt.Sprintf("%s -m pip install --root-user-action=ignore", pythonVersion)
+	command := "uv pip install --system --link-mode=copy"
```
nickpetrovic, line 459: Can you copy/paste what this looks like when building a container?

docker/Dockerfile.worker
```diff
@@ -108,6 +108,7 @@ RUN MOUNT_S3_URL=$(if [ "$TARGETARCH" = "arm64" ]; then echo $MOUNT_S3_URL_ARM64
     mv mount-s3/bin/mount-s3 /usr/local/bin/ms3 && chmod +x /usr/local/bin/ms3 && \
     rm -rf mount-s3 $TARBALL
 
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
```
nickpetrovic, line 111: I checked the script and my assumption is that this url points to the latest version of uv. Can we find a way to pin the version?

Makefile
```diff
@@ -60,4 +67,4 @@ verify-protocol:
 	./bin/verify_proto.sh
 
 test-pkg:
-	go test -v ./pkg/... -bench=./pkg/..
+	go test -v ./pkg/... -bench=./pkg/...
```
nickpetrovic, line 70: Whats this for?

pkg/abstractions/image/build.go
nickpetrovic: Just cause I'm curious - why do this in the first place? Was there a noticeable speed improvement? If yes, what was the duration differences between uv and uv+command grouping. If there's no particular reason, then this comment can be ignored.

dleviminzi: There is a noticeable speed improvement. But also, to clarify this still maintains the ordering. That is why I store the pip index and put an empty command in the command array. It is a bit fragile because if you have two different `add_python_packages` my assumptions fall apart. But as it is right now, instead of getting 

```
command 
command 
command
pip
pip
pip
``` 
you get 
```
command 
command 
command
pip
``` 

If the `add_python_packages` was put above the `add_commands` the order above would be reversed. The goal is not to do all commands and all pip together, but to do those listed together at once.

dleviminzi: The difference for uv was 4 seconds less when grouped. Not enormous, but not nothing

nickpetrovic: So something like this

```
pip
command 
command 
command
pip
pip
```

would still do this?

```
pip
command 
command 
command
pip (grouped)
```

dleviminzi: Yeah exactly. Right now, my implementation is broken for the multiple pip case because I was just curious what performance would be like. But I think it might be worth doing it for real

docker/Dockerfile.worker
dleviminzi: I think we can probably just pin to a version in their [releases](https://github.com/astral-sh/uv/releases)

Makefile
dleviminzi: typo

docker/Dockerfile.worker
dleviminzi: Oh they actually have this ready to go 

curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.4.26/uv-installer.sh | sh

nickpetrovic: Yeah that would work.

pkg/abstractions/image/build.go
dleviminzi: It looks like this:
```
=> Building image
Waiting for build container to start...
Python 3.10.15
6
Using Python 3.10.15 environment at /usr
Resolved 35 packages in 434ms
Prepared 28 packages in 3.17s

Saving image, this may take a few minutes...
Installed 28 packages in 238ms
 + certifi==2024.8.30
 + contourpy==1.3.0
 + cycler==0.12.1
 + filelock==3.16.1
 + fonttools==4.54.1
 + fsspec==2024.10.0
 + huggingface-hub==0.26.1
 + idna==3.10
 + joblib==1.4.2
 + kiwisolver==1.4.7
 + matplotlib==3.9.2
 + numpy==2.1.2
 + opencv-python==4.10.0.84
 + pandas==2.2.3
 + pillow==11.0.0
 + pytz==2024.2
 + pyyaml==6.0.2
 + regex==2024.9.11
 + safetensors==0.4.5
 + scikit-learn==1.5.2
 + scipy==1.14.1
 + seaborn==0.13.2
 + threadpoolctl==3.5.0
 + tokenizers==0.20.1
 + tqdm==4.66.5
 + transformers==4.46.0
 + tzdata==2024.2
 + urllib3==1.26.20
=> Build complete 🎉
=> Syncing files
```

nickpetrovic: Oh great - its still readable. I haven't run uv yet, was mostly curious.

pkg/abstractions/image/build.go
nickpetrovic: Okay perfect. Yeah I think we'd want this only if we can support using multiple calls to `add_python_packages`.

dleviminzi: Yeah I have to fix this in my conda PR actually. So I'll work out a way of supporting multiple calls.

PR #660, opened 2024-10-25
Feat: Detect and log OOM kills in container logs

nickpetrovic:
- Start goroutine of watching for oom events right after runc container pid created
- Send oom message to output chan
- Move delete container instance from processStopContainerEvents to stopContainer and only delete if kill=true
- Move "container still running" log message after detecting for orphaned container

Resolve BE-1972


pkg/worker/lifecycle.go
```diff
@@ -507,3 +517,19 @@ func (s *Worker) createOverlay(request *types.ContainerRequest, bundlePath strin
...
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			if c, ok := s.containerInstances.Get(containerId); ok && c.isOOMKilled() {
+				log.Printf("<%s> - [WARNING] A process in the container was killed due to out-of-memory conditions.\n", containerId)
```
luke-lombardi, line 531: I think this needs to use the logger on the container instance, otherwise it won't show up for users

pkg/worker/worker.go
```diff
@@ -76,6 +78,63 @@ type ContainerInstance struct {
...
+	memoryEvents MemoryEvents
+}
+
+func (i *ContainerInstance) isOOMKilled() bool {
+	podUID := os.Getenv("POD_UID")
+	if podUID == "" {
+		return false
+	}
+
+	memoryEventsPath := fmt.Sprintf("/sys/fs/cgroup/kubepods/besteffort/pod%s/%s/memory.events", podUID, i.Id)
```
luke-lombardi, line 91: this ties us into the kubepods namespace pretty tightly. is there a way to make this more generic i.e. detect where this path is dynamically?

pkg/worker/worker.go
```diff
@@ -76,6 +78,63 @@ type ContainerInstance struct {
 	Port         int
 	OutputWriter *common.OutputWriter
 	LogBuffer    *common.LogBuffer
+
+	memoryEvents MemoryEvents
+}
+
+func (i *ContainerInstance) isOOMKilled() bool {
```
luke-lombardi, line 85: if this is on container instance - maybe let's move container instance to another file

luke-lombardi (changes requested):

pkg/worker/worker.go
nickpetrovic: Moved to the lifecycle.go file.

pkg/worker/worker.go
nickpetrovic: I updated the cgroup path set on the runc spec to `/sys/fs/cgroups/<worker-id>/<container-id>`. Its unclear to me if this will work in k3s since processes are managed by systemd. I'll need to test a custom worker build in staging.

pkg/worker/lifecycle.go
nickpetrovic: Replaced the print with sending to the output channel.

pkg/worker/lifecycle.go
```diff
@@ -511,3 +520,47 @@ func (s *Worker) createOverlay(request *types.ContainerRequest, bundlePath strin
 func (s *Worker) isBuildRequest(request *types.ContainerRequest) bool {
 	return request.SourceImage != nil
 }
+
+func (s *Worker) watchOOMEvents(ctx context.Context, containerId string, output chan common.OutputMsg) {
+	time.Sleep(time.Second)
```
luke-lombardi, line 525: This feels like it could be a race condition, is there a way we could make this event based / only start watching when we're sure the container is up?

luke-lombardi (changes requested):

pkg/worker/lifecycle.go
nickpetrovic: We wait for the pid in the spawn function. For some reason, if we don't sleep or retry making the events channel, the channel can be closed right after we just made it. 

I updated the logic to where we don't have random sleeps and can retry making the events channel up to 5 times before bailing.

PR #663, opened 2024-10-27
Chore: Cleanup logic for gpu count and added a helper to check GPU requirements

jsun-m:
Small changes relating to cleanup and adding of a helper function for validating gpu requirements.

Also moved `container.scheduled` events after request.Gpu is updated

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -374,7 +374,7 @@ const maxScheduleRetryCount = 3
 const maxScheduleRetryDuration = 10 * time.Minute
 
 func (s *Scheduler) addRequestToBacklog(request *types.ContainerRequest) error {
-	if request.RequiresGPU() {
+	if request.RequiresGPU() && request.GpuCount <= 0 {
 		request.GpuCount = 1
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -372,7 +372,7 @@ const maxScheduleRetryCount = 3
 const maxScheduleRetryDuration = 10 * time.Minute
 
 func (s *Scheduler) addRequestToBacklog(request *types.ContainerRequest) error {
-	if request.Gpu != "" && request.GpuCount <= 0 {
+	if len(request.GpuRequest) > 0 && request.GpuCount <= 0 {
```
luke-lombardi, line 375: how do you feel about adding a struct function to ContainerRequest like RequiresGPU()
which inspects r.GpuRequest, for example:

```
func (r *ContainerRequest) RequiresGPU() bool {
   return len(r.GpuRequest) > 0
}
```

luke-lombardi: could clean up quite a few places i think?

jsun-m: yea i think thats a good idea

luke-lombardi (approved):

PR #665, opened 2024-10-29
Feat: VLLM abstraction

dleviminzi:
Resolve BE-1783

This PR adds a VLLM abstraction. Every argument you can provide to `vllm serve ...` is supported and can be passed into the vllm class or as part of the `EngineConfig`. The defaults are the same as the ones in vllm's repo. 

Here is a simple example of how it used: 
```python
from beta9 import integrations

e = integrations.EngineConfig()
e.device = "cpu"

vllm_app = integrations.VLLM(name="vllm-abstraction-1", engine_config=e)
```

And we deploy with:
```bash
beta9 deploy app.py:vllm_app
```

Serve with:
```bash 
beta9 serve app.py:vllm_app
```

Below is an example of serving the above on my local machine. 

In order to get deploying/serving a class to work, I introduced a `set_handler` method. The deploy and serve code now checks if that message is present and uses it to set the user provided `app:obj`. Without this, we end up inspecting the sdk to generate the module:obj and that does not work. 

Screenshot 2024-10-29 at 17 24 47

### DIFF
--- a/sdk/src/beta9/abstractions/integrations/vllm.py
+++ b/sdk/src/beta9/abstractions/integrations/vllm.py
@@ -157,8 +157,11 @@ class VLLM(ASGI):
         chat_template: Optional[str] = None,
         return_tokens_as_token_ids: bool = False,
         enable_auto_tools: bool = False,
+        enable_auto_tool_choice: bool = False,
         tool_call_parser: Optional[str] = None,
         disable_log_stats: bool = False,
+        disable_log_requests: bool = False,
+        max_log_len: Optional[int] = None,
     ):
         # ASGI initialization
         super().__init__(
@@ -179,11 +182,15 @@ class VLLM(ASGI):
         self.engine_config = engine_config
         self.vllm_args = SimpleNamespace(
             model=engine_config.model,
+            served_model_name=engine_config.served_model_name,
+            disable_log_requests=disable_log_requests,
+            max_log_len=max_log_len,
             response_role=response_role,
             lora_modules=lora_modules,
             prompt_adapters=prompt_adapters,
             chat_template=chat_template,
             return_tokens_as_token_ids=return_tokens_as_token_ids,
+            enable_auto_tool_choice=enable_auto_tool_choice,
             enable_auto_tools=enable_auto_tools,
             tool_call_parser=tool_call_parser,
             disable_log_stats=disable_log_stats,
@@ -206,6 +213,7 @@ class VLLM(ASGI):
         from vllm.engine.async_llm_engine import AsyncLLMEngine
         from vllm.entrypoints.openai.api_server import (
             init_app_state,
+            router,
         )
         from vllm.usage.usage_lib import UsageContext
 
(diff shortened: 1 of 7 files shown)


sdk/src/beta9/abstractions/integrations/vllm.py
```diff
@@ -0,0 +1,334 @@
...
+where app.py is just a simple file with the instance of the vllm app
+
+# app.py
+from beta9 import vllm
+
+app = vllm(stub_args, vllm_args)
+
+Then on deploy, we set this up as an ASGI app by having some getter that returns the FASTAPI instance.
+
+app.get_fastapi_app()
```
dleviminzi, line 13: I gotta fix this

mernit:
the `VLLM` class should be lowercase IMO

sdk/src/beta9/abstractions/integrations/vllm.py
```diff
@@ -0,0 +1,390 @@
...
+    StopEndpointServeRequest,
+)
+from ...clients.gateway import DeployStubRequest, DeployStubResponse
+from ...config import ConfigContext
+from ...type import GpuType, GpuTypeAlias
+
+
+# vllm/engine/arg_utils.py:EngineArgs
+@dataclass
+class EngineConfig:
```
luke-lombardi, line 26: should we call this VLLMEngineConfig?

sdk/src/beta9/cli/deployment.py
```diff
@@ -109,26 +109,31 @@ def create_deployment(service: ServiceClient, name: str, entrypoint: str, url_ty
...
+    user_obj: Optional[DeployableMixin] = getattr(module, obj_name, None)
+    if user_obj is None:
         terminal.error(
-            f"Invalid handler function specified. Make sure '{module_path}' contains the function: '{func_name}'"
+            f"Invalid handler function specified. Make sure '{module_path}' contains the function: '{obj_name}'"
         )
 
-    if not user_func.deploy(name=name, context=service._config, url_type=url_type):  # type: ignore
+    # Class based integrations should have a set_handler method so they don't end up inspecting the sdk when
+    # trying to get the handler
```
luke-lombardi, line 132: can you clean up this comment

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/integrations/vllm.py
```diff
@@ -0,0 +1,390 @@
...
+    otlp_traces_endpoint: Optional[str] = None
+    collect_detailed_traces: Optional[str] = None
+    disable_async_output_proc: bool = False
+    override_neuron_config: Optional[Dict[str, Any]] = None
+    mm_processor_kwargs: Optional[Dict[str, Any]] = None
+    scheduling_policy: Literal["fcfs", "priority"] = "fcfs"
+    disable_log_requests: bool = False
+
+
+class vllm(ASGI):
```
luke-lombardi, line 135: @mernit I actually think this should be Vllm or VLLM to distinguish it from the vllm module. Also it's not a decorator anyway (it's a class) and in python that's the standard way of formatting classes

luke-lombardi (changes requested):
One last comment

PR #666, opened 2024-10-30
Feat: Display config path after adding new context to config

nickpetrovic:
Resolve BE-1996

### DIFF
--- a/sdk/src/beta9/cli/config.py
+++ b/sdk/src/beta9/cli/config.py
@@ -146,8 +146,7 @@ def create_context(config_path: Path, **kwargs):
     contexts[name] = context
     save_config(contexts=contexts, path=config_path)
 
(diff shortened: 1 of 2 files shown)


sdk/src/beta9/cli/config.py
```diff
@@ -146,7 +146,8 @@ def create_context(config_path: Path, **kwargs):
     contexts[name] = context
     save_config(contexts=contexts, path=config_path)
 
-    terminal.success("Added new context 🎉!")
+    rel_path = str(config_path).replace(str(Path.home()), "~")
```
dleviminzi, line 149: I don't think `~` will work for users on windows machines

nickpetrovic: Updated this to just use the path.

dleviminzi (approved):

PR #671, opened 2024-10-31
fix: bubble up exceptions invoking functions

luke-lombardi:
- fix a bug where exceptions raised while loading a handler are not visible to users in logs

### DIFF
--- a/sdk/src/beta9/runner/function.py
+++ b/sdk/src/beta9/runner/function.py
@@ -136,7 +136,7 @@ def _monitor_task(
 
 
 @json_output_interceptor(task_id=config.task_id)
-@handle_error(print_traceback=False)
+@handle_error(print_traceback=True)
 @pass_channel
 def main(channel: Channel):
     function_stub: FunctionServiceStub = FunctionServiceStub(channel)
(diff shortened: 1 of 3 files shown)


mernit (approved):

sdk/src/beta9/runner/function.py
```diff
@@ -227,6 +227,7 @@ def invoke_function(
             callback_url=callback_url,
         )
     except BaseException as e:
+        print(f"{traceback.format_exc()}")
```
nickpetrovic, line 230: This already happens here https://github.com/beam-cloud/beta9/pull/671/files#diff-1739866d6bf74331c67757941a077b334836efd70a493b62ba971a863e5604cbL177

nickpetrovic (changes requested):

sdk/src/beta9/runner/function.py
luke-lombardi: or it doesn't seem to. it's not working on my end.

nickpetrovic (approved):
Okay this makes sense. Just didn't want this to double print exceptions. The issue was that we kept the same behavior of printing the exception in the main function, so changes to the handle_error decorator were made to avoid printing it twice. 

So instead, what we've done here is what it should've been. Just don't print in the main function and remove the print_traceback parameter.

PR #673, opened 2024-10-31
Feat: Add worker id to container list when token is cluster admin

nickpetrovic:
Resolve BE-2021


pkg/gateway/services/container.go
```diff
@@ -13,26 +13,58 @@ import (
...
-			Ok:       false,
-			ErrorMsg: "Unable to list containers",
-		}, nil
+	type containerDetails struct {
+		WorkerId  string
+		MachineId string
 	}
 
-	containers := make([]*pb.Container, 0, len(containerStates))
+	var (
```
luke-lombardi, line 21: Could we move this admin specific stuff out to another helper function or something?

luke-lombardi (changes requested):

pkg/gateway/services/container.go
nickpetrovic: We can. Just updated see if it makes better sense.

dleviminzi (approved):

pkg/gateway/services/container.go
luke-lombardi: Yeah I think this is a little cleaner

luke-lombardi (approved):

sdk/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "beta9"
-version = "0.1.100"
+version = "0.1.101"
```
luke-lombardi, line 3: I think we just release 101, need a bump here

luke-lombardi (changes requested):

sdk/pyproject.toml
nickpetrovic: fixed

dleviminzi: just did 102, my bad

PR #678, opened 2024-10-31
Fix: exit more loops when context is cancelled

nickpetrovic:
Check if context is done in
- Provider's Reconcile func
- LocalKubernetesWorkerPoolController's deleteStalePendingWorkerJobs func
- RunCFunctionService's FunctionMonitor func
- WorkerPoolSizer's Start func
- Add exportable context on worker pool controllers

### DIFF
--- a/pkg/scheduler/pool.go
+++ b/pkg/scheduler/pool.go
@@ -1,6 +1,7 @@
 package scheduler
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strconv"
@@ -35,6 +36,7 @@ type WorkerPoolController interface {
 	AddWorkerToMachine(cpu int64, memory int64, gpuType string, gpuCount uint32, machineId string) (*types.Worker, error)
 	Name() string
 	FreeCapacity() (*WorkerPoolCapacity, error)
+	Context() context.Context
 }
 
 type WorkerPoolConfig struct {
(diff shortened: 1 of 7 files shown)


pkg/scheduler/pool.go
```diff
@@ -35,6 +36,7 @@ type WorkerPoolController interface {
 	AddWorkerToMachine(cpu int64, memory int64, gpuType string, gpuCount uint32, machineId string) (*types.Worker, error)
 	Name() string
 	FreeCapacity() (*WorkerPoolCapacity, error)
+	Context() context.Context
```
luke-lombardi, line 39: why are we exposing this? might be missing something but I don't see where it's used

luke-lombardi (changes requested):

pkg/scheduler/pool.go
nickpetrovic: Its being used here - https://github.com/beam-cloud/beta9/pull/678/files#diff-655f521cd80c31efe08ad17bdf5340cc513ffb2710fa3e12aa845660f24c6b88R40 - in the worker pool sizer start function.

nickpetrovic: WorkerPoolSizer takes a controller interface hence the addition of this function.

luke-lombardi (approved):

PR #681, opened 2024-11-01
Fix: Correct build command ordering and add build tests

dleviminzi:
Resolve BE-2024

The append was still wrong and I've added a script to run against stage before going to prod. 

Running the script locally will likely not work for custom images on macs due to architecture. That said, they can be run against stage.


nickpetrovic (approved):

build_tests/app.py
```diff
@@ -0,0 +1,66 @@
...
+    Image(
+        python_version="python3.10",
+        base_image="nvidia/cuda:12.2.2-cudnn8-runtime-ubuntu22.04",
+    )
+    .add_python_packages(["torch"])
+    .add_commands([f"echo 1-{uuid.uuid4()}"])
+)
+rapids_img = Image(base_image="rapidsai/rapidsai").add_commands(
+    [f"echo {uuid.uuid4()}"]
+)
```
nickpetrovic, line 27: Would be nice to do this on a matrix of python versions too, but I can see that not being desirable for cost of storage reasons.

dleviminzi: Yeah it would be nice to cover a wider variety of bases. I think I'll add those and then flag them as local. That way you could do `make build-test local` and that will run extra tests and exclude ones that will break on dev machines due to architecture.

PR #682, opened 2024-11-01
Fix: Secrets, autoscaler, and chat templates for vLLM abstraction

dleviminzi:
This PR makes two important corrections to the VLLM abstraction. 

First, it adds the secrets parameters. This is critical as it allows us to pass `HF_TOKEN` as a secret, which is required to download many of the models. 

Second, it adds a chat_template_url parameter. If set, we download this template and override the chat_template arg.

### DIFF
--- a/sdk/src/beta9/abstractions/integrations/vllm.py
+++ b/sdk/src/beta9/abstractions/integrations/vllm.py
@@ -18,7 +18,7 @@ from ...clients.endpoint import (
 )
 from ...clients.gateway import DeployStubRequest, DeployStubResponse
 from ...config import ConfigContext
-from ...type import GpuType, GpuTypeAlias
+from ...type import Autoscaler, GpuType, GpuTypeAlias, QueueDepthAutoscaler
 
 DEFAULT_VLLM_CACHE_DIR = "./vllm-cache"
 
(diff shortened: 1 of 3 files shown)


sdk/src/beta9/abstractions/integrations/vllm.py
```diff
@@ -279,24 +294,36 @@ def func(self, *args: Any, **kwargs: Any):
...
-            init_app_state,
-            router,
-        )
         from vllm.usage.usage_lib import UsageContext
 
+        if self.chat_template_url:
+            import requests
+
+            response = requests.get(self.chat_template_url)
+            with open("./vllm-cache/chat_template.jinja", "wb") as file:
```
luke-lombardi, line 307: should this overwrite the chat template like this? what if you had two separate vllm deployments using the same cache path?

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/integrations/vllm.py
dleviminzi: I changed the logic so that it writes to a file with the name set to whatever the last part of the provided url was. And, it only does so if that file does not already exist.

I added an additional change that appends the default cache volume only if the default download directory is pointing to it and downloads the templates to the download directory. This way, if a user points to a different download directory and mounts that volume we send everything there and don't bother adding the default mount. Also, since its not set as a default anymore but instead appended, there is no risk of a user accidentally removing the default.

luke-lombardi (approved):

PR #683, opened 2024-11-02
Feat: Add new filter fields to task list command

nickpetrovic:
We had some of these filters before. I think the functionality was lost when we moved to using Squirrel.

This re-adds `stub-id`, `stub-name`, and `stub-type`, but in this iteration, all of these can now accept multiple values (comma delimited). And this adds new filter fields `container-id` and `task-id`. Lastly, all filter fields have a combination of names that are single, plural, hyphened, and underscored. 

Resolve BE-2026

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ verify-protocol:
 test-pkg:
 	go test -v ./pkg/... -bench=./pkg/..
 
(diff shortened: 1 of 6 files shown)


mernit:
Nice. Will add these to the docs.

luke-lombardi (approved):

PR #685, opened 2024-11-04
Fix: Improve shutdown logic, add more context done checks

nickpetrovic:
- Improve shutdown logic by running server (gRPC/HTTP) shutdowns and global context cancellation in parallel
- Merge global and local contexts to avoid modifying downstream code that doesn't need to know about the global context
- In the runc_server.go RunCStreamLogs func, make sure we can cancel the loop if the gateway is being shutdown 
- Cancel RequestBuffer heartBeat func if gateway is shutting down

### DIFF
--- a/pkg/abstractions/common/logs.go
+++ b/pkg/abstractions/common/logs.go
@@ -102,7 +102,7 @@ _stream:
 			}
 
 		case <-ctx.Done():
-			// This ensures when the sdk exists, the message printed is
+			// This ensures when the sdk exits, the message printed is
 			// that the container timed out.
 			if err := l.exitCallback(0); err != nil {
 				break _stream
--- a/pkg/abstractions/common/logs.go
+++ b/pkg/abstractions/common/logs.go
@@ -102,6 +102,11 @@ _stream:
 			}
 
 		case <-ctx.Done():
+			// This ensures when the sdk exists, the message printed is
+			// that the container timed out.
+			if err := l.exitCallback(0); err != nil {
+				break _stream
+			}
 			return nil
 		}
 	}
--- a/pkg/common/contexts.go
+++ b/pkg/common/contexts.go
@@ -1,0 +1,30 @@
+package common
+
+import (
+	"context"
+	"sync"
+)
+
+func MergeContexts(ctxs ...context.Context) (context.Context, context.CancelFunc) {
+	ctx, cancel := context.WithCancel(context.Background())
+	var wg sync.WaitGroup
+
+	for _, c := range ctxs {
+		wg.Add(1)
+		go func(c context.Context) {
+			defer wg.Done()
+			select {
+			case <-c.Done():
+				cancel() // Cancel the derived context if any parent context is done
+			case <-ctx.Done():
+			}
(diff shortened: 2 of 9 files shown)


nickpetrovic:
From my testing, these changes allows the gateway to shutdown. Its worth testing this in staging with a remote gpu node to observe how the gateways (multiple) will respond when they are all `rollout restarted`.

pkg/abstractions/common/logs.go
```diff
@@ -102,6 +102,11 @@ _stream:
 			}
 
 		case <-ctx.Done():
+			// This ensures when the sdk exists, the message printed is
```
luke-lombardi, line 105: typo in exists

pkg/common/contexts.go
```diff
@@ -0,0 +1,30 @@
+package common
+
+import (
+	"context"
+	"sync"
+)
+
+func MergeContexts(ctxs ...context.Context) (context.Context, context.CancelFunc) {
```
luke-lombardi, line 8: are we sure we _always_ want child contexts to be cancelled immediately if the parents is cancelled? Just want to make sure we don't cancel a DB transaction or something.

luke-lombardi (changes requested):

pkg/common/contexts.go
```diff
@@ -0,0 +1,30 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	var wg sync.WaitGroup
+
+	for _, c := range ctxs {
```
luke-lombardi, line 12: also are there no built-ins for this kind of functionality? feels like overkill to add a separate utility for it

luke-lombardi: nvm not sure that one does what we want here

nickpetrovic: I don't think there is a built-in for this. There are 3rd party packages, but many of them seem to manage state with a struct and locks. To me it seems like overkill to manage state for this.

pkg/common/contexts.go
nickpetrovic: We shouldn't use this for every time theres two contexts. This is mostly a helper for when merging two contexts is okay. For example, we're merging contexts before we call the LogStream.Stream function. In particular, we can break out of the for loop in handleStream() with just a single/wrapped context vs adding a global context to LogStream and then adding a second case statement in handleStream. 

We're also using the merged context in goroutines where we don't need to add a second case statement for a global context - at least I don't think we do. 1. because they're goroutines and errors aren't handled there and 2. I didn't see a reason to have different logic if the global context was cancelled.

Hopefully that makes sense. LMK if there's something that you thinking of specifically.

pkg/abstractions/common/logs.go
nickpetrovic: Fixed.

pkg/common/contexts.go
luke-lombardi: Yep, that makes sense to me.

luke-lombardi (approved):

luke-lombardi (approved):

PR #687, opened 2024-11-04
Add pre-emptible flag to worker pool config

luke-lombardi:
- Adds an optional (default=false) preemptible flag
- Allows filtering of workers by request flags
- Allows filtering of worker pool controllers by request flags

**NOTE**: behavior of the schedule should not change unless we modify container requests going into the schedule with the `Preemptible=true` flag and add preemptible worker pools. By default they are all nonpremptible.

### DIFF
--- a/pkg/scheduler/pool_external.go
+++ b/pkg/scheduler/pool_external.go
@@ -410,6 +410,10 @@ func (wpc *ExternalWorkerPoolController) getWorkerEnvironment(workerId, machineI
 			Name:  "NETWORK_PREFIX",
 			Value: machineId,
 		},
+		{
+			Name:  "PREEMPTABLE",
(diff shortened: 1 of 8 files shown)


pkg/scheduler/pool_external.go
```diff
@@ -410,6 +415,10 @@ func (wpc *ExternalWorkerPoolController) getWorkerEnvironment(workerId, machineI
 			Name:  "NETWORK_PREFIX",
 			Value: machineId,
 		},
+		{
```
luke-lombardi, line 418: This may or may not be required, but I thought having the env var could useful worker side. It's also saved on the worker state itself, so this may be redundant.

nickpetrovic (approved):

PR #688, opened 2024-11-04
Manual login for hf token

dleviminzi:
Check if hugging face token is present and use it to login if it is. 

A warning is logged from vLLM claiming that this login is redundant, but without it gated downloads don't work...

### DIFF
--- a/sdk/src/beta9/abstractions/integrations/vllm.py
+++ b/sdk/src/beta9/abstractions/integrations/vllm.py
@@ -243,7 +243,9 @@ class VLLM(ASGI):
         cpu: Union[int, float, str] = 1.0,
         memory: Union[int, str] = 128,
         gpu: Union[GpuTypeAlias, List[GpuTypeAlias]] = GpuType.NoGPU,
-        image: Image = Image(python_version="python3.11").add_python_packages(["fastapi", "vllm"]),
+        image: Image = Image(python_version="python3.11").add_python_packages(
+            ["fastapi", "vllm", "huggingface_hub"]
+        ),
         workers: int = 1,
         concurrent_requests: int = 1,
         keep_warm_seconds: int = 60,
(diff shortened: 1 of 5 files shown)


sdk/src/beta9/abstractions/integrations/vllm.py
```diff
@@ -330,6 +330,12 @@ def __call__(self, *args: Any, **kwargs: Any):
                 f"{self.engine_config.download_dir}/{chat_template_filename}"
             )
 
+        if "HF_TOKEN" in os.environ:
+            hf_token = os.environ["HF_TOKEN"]
+            import huggingface_hub
```
nickpetrovic, line 335: Should we check if huggingface_hub is installed e.g. do a try/except?

dleviminzi: I think we should be ok because it is a dependency of vllm. I can explicitly add it to the image requirements to make sure though

nickpetrovic: Just saw your change. Can't that be overwritten? I think if we're not going to try/cache the import, then we need to make sure huggingface is installed during `__init__`.

nickpetrovic: Installed meaning, calling add_python_packages in the init function.

dleviminzi: hmm yeah, I will change the logic to be similar to how we handle the volume mounts.

nickpetrovic: Okay cool. Looks good. We just want to make sure the package is available or if its not, handle the potential exception.

nickpetrovic (approved):

sdk/src/beta9/abstractions/integrations/vllm.py
dleviminzi: Yeah makes sense. When I thought about it, I concluded that we should make sure vllm is there no matter what.

PR #694, opened 2024-11-06
Fix: worker hangs on shutdown

nickpetrovic:
- Don't delete container state or instance in stopContainer func
- Don't stop all containers in shutdown func
- Call stopContainer in clearContainer func right before delete container state is called only if the container is still running. This allows the container to terminate when the global context is cancelled and finish executing the finalizeContainer and clearContainer functions. Also, this is needed when there is a sigterm, but not when the container stops normally.
- Don't use time.Sleep to delete container state and instance. Instead use a select statement that waits for either the graceful timeout or for the global context to cancel.
- Enable runc debug mode when global debug is enabled
- Revert delete worker changes

Resolve BE-2038

### DIFF
--- a/pkg/worker/lifecycle.go
+++ b/pkg/worker/lifecycle.go
@@ -56,7 +56,8 @@ func (s *Worker) handleStopContainerEvent(event *common.Event) bool {
 func (s *Worker) stopContainer(containerId string, kill bool) error {
 	log.Printf("<%s> - stopping container.\n", containerId)
 
-	if _, exists := s.containerInstances.Get(containerId); !exists {
+	_, exists := s.containerInstances.Get(containerId)
+	if !exists {
 		log.Printf("<%s> - container not found.\n", containerId)
 		return nil
 	}
@@ -64,8 +65,6 @@ func (s *Worker) stopContainer(containerId string, kill bool) error {
 	signal := int(syscall.SIGTERM)
 	if kill {
 		signal = int(syscall.SIGKILL)
-		s.containerRepo.DeleteContainerState(containerId)
-		defer s.containerInstances.Delete(containerId)
 	}
 
 	err := s.runcHandle.Kill(context.Background(), containerId, signal, &runc.KillOpts{All: true})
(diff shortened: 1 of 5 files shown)


pkg/worker/worker.go
```diff
@@ -440,14 +440,6 @@ func (s *Worker) shutdown() error {
 	// Stops goroutines
 	s.cancel()
 
-	// Stop all running containers forcefully
-	s.containerInstances.Range(func(containerId string, _ *ContainerInstance) bool {
-		if err := s.stopContainer(containerId, true); err != nil {
-			log.Printf("Failed to stop container: %v\n", err)
-		}
-		return true // continue iterating
-	})
```
nickpetrovic, line 449: We don't need to stop containers here. Let them stop themselves when the global context is cancelled.

pkg/worker/lifecycle.go
```diff
@@ -128,7 +127,14 @@ func (s *Worker) clearContainer(containerId string, request *types.ContainerRequ
 	go func() {
 		// Allow for some time to pass before clearing the container. This way we can handle some last
 		// minute logs or events or if the user wants to inspect the container before it's cleared.
-		time.Sleep(delay)
+		select {
+		case <-time.After(time.Duration(s.config.Worker.TerminationGracePeriod) * time.Second):
+		case <-s.ctx.Done():
+		}
+
+		if err := s.stopContainer(containerId, false); err != nil {
```
luke-lombardi, line 135: I'm confused, in the normal case (i.e. no sigterm), wouldn't the container already be stopped here? Why do we need to call stopcontainer

luke-lombardi (changes requested):

pkg/worker/lifecycle.go
nickpetrovic: That's right. I just updated this to check if the container is still running.

pkg/worker/lifecycle.go
```diff
@@ -128,13 +127,26 @@ func (s *Worker) clearContainer(containerId string, request *types.ContainerRequ
...
 		// minute logs or events or if the user wants to inspect the container before it's cleared.
-		time.Sleep(delay)
+		select {
+		case <-time.After(time.Duration(s.config.Worker.TerminationGracePeriod) * time.Second):
+		case <-s.ctx.Done():
+		}
+
+		// If the container is still running, stop it. This happens when a sigterm is detected.
+		container, err := s.runcHandle.State(context.TODO(), containerId)
+		if err == nil && container.Status == "running" {
```
luke-lombardi, line 137: Anyway we can get this constant "running" from runc types or something?

luke-lombardi (changes requested):
Left one comment otherwise lgtm

pkg/worker/lifecycle.go
nickpetrovic: Yeah we can. Should I update the other place this isn't using a const or type?

https://github.com/beam-cloud/beta9/blob/main/pkg/worker/runc_server.go#L128

luke-lombardi: yeah I think we should

pkg/worker/worker.go
luke-lombardi: 👍

pkg/worker/lifecycle.go
nickpetrovic: Updated.

luke-lombardi (approved):

PR #699, opened 2024-11-11
Feat: Use python-build-standalone over apt packages

nickpetrovic:
- Use python-build-standalone as a drop-in replacement for custom images without python
- Template the install script, python versions, and a few other runtime variables

Resolve BE-1974

### DIFF
--- a/pkg/abstractions/image/build.go
+++ b/pkg/abstractions/image/build.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log"
 	"regexp"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


nickpetrovic:
Looking for suggestions on whether or not we should make these values more configurable, if we should store the builds in object storage, and/or if we should replace software-properties-common with build-essential.

pkg/abstractions/image/build.go
```diff
@@ -493,6 +494,65 @@ func getPythonInstallCommand(pythonVersion string) string {
...
+		venderAndOs = "apple-darwin"
+	default:
+		log.Println("Unsupported OS")
+	}
+
+	// TODO: Make this configurable?
+	var pySemVer string
+	switch pythonVersion[6:] {
+	case "3.9":
+		pySemVer = "3.9.20"
```
luke-lombardi, line 522: Where are these values from?

nickpetrovic: Updated the code to get these from the config.

luke-lombardi (approved):
👍

PR #709, opened 2024-11-12
Fix: Update default values for python-build-standalone

nickpetrovic:
Uses the versions available for a particular release.

Resolve BE-1974


nickpetrovic:
Use a slightly less newer version since Python 3.8 is EOL and no longer supported in https://github.com/indygreg/python-build-standalone.

luke-lombardi (approved):

PR #712, opened 2024-11-13
Fix: On worker shutdown, handle worker not found error

nickpetrovic:
Resolve BE-2046

### DIFF
--- a/bin/delete_workers.sh
+++ b/bin/delete_workers.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+echo "Deleting worker jobs..."
+kubectl delete job -l run.beam.cloud/role=worker
+
 echo "Deleting redis keys..."
 if kubectl get sts redis-master &> /dev/null; then
   replicas=$(kubectl get sts redis-master -o jsonpath='{.spec.replicas}')
@@ -12,10 +15,5 @@ if kubectl get sts redis-master &> /dev/null; then
     kubectl exec redis-master-$i -- bash -c 'for k in $(redis-cli keys worker:*); do redis-cli -c del $k; done' &
   done
 fi
-
-# Workers now handle deleting themselves from redis on shutdown. This will happen faster than when
-# the redis keys are deleted.
-echo "Deleting worker jobs..."
-kubectl delete job -l run.beam.cloud/role=worker
 
 wait
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -2,27 +2,27 @@ package main
 
 import (
 	"log"
-	"os"
 
 	"github.com/beam-cloud/beta9/pkg/types"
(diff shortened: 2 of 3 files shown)


luke-lombardi (changes requested):
I think this error also occurs on startup as well. When it loads the worker -- and that also leads to crash looping.

nickpetrovic:
> I think this error also occurs on startup as well. When it loads the worker -- and that also leads to crash looping.

I believe we fixed that already. See [this](https://github.com/beam-cloud/beta9/blob/main/pkg/worker/worker.go#L374-L378).

nickpetrovic:
> > I think this error also occurs on startup as well. When it loads the worker -- and that also leads to crash looping.
> 
> I believe we fixed that already. See [this](https://github.com/beam-cloud/beta9/blob/main/pkg/worker/worker.go#L374-L378).
> 
> Before we were not cancelling the global context, now we are. And everything else (or most things) in the worker now check the global context and will return if done, including the main loop which checks if it should shutdown using this:
> 
> ```go
> func (s *Worker) shouldShutDown(lastContainerRequest time.Time) bool {
> 	select {
> 	case <-s.ctx.Done():
> 		return true
> 	default:
> 		if (time.Since(lastContainerRequest).Seconds() > defaultWorkerSpindownTimeS) && s.containerInstances.Len() == 0 {
> 			return true
> 		}
> 		return false
> 	}
> }
> ```

Scratch all that. Looks like the worker not found error can be raised in the network manager. That might be where its happening. I'll address this.

bin/delete_workers.sh
```diff
@@ -13,9 +16,4 @@ if kubectl get sts redis-master &> /dev/null; then
   done
 fi
 
-# Workers now handle deleting themselves from redis on shutdown. This will happen faster than when
-# the redis keys are deleted.
-echo "Deleting worker jobs..."
-kubectl delete job -l run.beam.cloud/role=worker
-
 wait
```
nickpetrovic, line 19: This was a changed before because of the worker not found error. Its being reverted now.

cmd/worker/main.go
```diff
@@ -2,18 +2,27 @@ package main
 
 import (
 	"log"
+	"os"
 
+	"github.com/beam-cloud/beta9/pkg/types"
 	"github.com/beam-cloud/beta9/pkg/worker"
 )
 
+func handleWorkerError(err error, context string) {
```
luke-lombardi, line 11: hm do we really need a function for this?

nickpetrovic: No not at all. I was on the fence on a copy/paste job or a func haha. I can just check the error twice.

nickpetrovic: Updated.

luke-lombardi (approved):

PR #714, opened 2024-11-14
Fix: Race condition when adding workers on remote nodes

nickpetrovic:
- Add lock by pool name when adding workers in occupyAvailableMachines func. This prevents multiple instances of gateway from adding workers to the same pool at the same time

Resolve BE-2045

### DIFF
--- a/pkg/common/keys.go
+++ b/pkg/common/keys.go
@@ -36,7 +36,7 @@ var (
 	workerNetworkLock            string = "worker:network:%s:lock"
 	workerNetworkIpIndex         string = "worker:network:%s:ip_index"
 	workerNetworkContainerIp     string = "worker:network:%s:container_ip:%s"
-	workerPoolSizerLock          string = "worker:poolsizer:%s:lock"
+	workerPoolSizerLock          string = "worker:pool_sizer:%s:lock"
 )
 
 var (
--- a/pkg/common/keys.go
+++ b/pkg/common/keys.go
@@ -36,6 +36,7 @@ var (
 	workerNetworkLock            string = "worker:network:%s:lock"
 	workerNetworkIpIndex         string = "worker:network:%s:ip_index"
 	workerNetworkContainerIp     string = "worker:network:%s:container_ip:%s"
+	workerPoolSizerLock          string = "worker:poolsizer:%s:lock"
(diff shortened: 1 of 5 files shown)


pkg/common/keys.go
```diff
@@ -242,12 +238,8 @@ func (rl *redisKeys) WorkspaceAuthorizedToken(token string) string {
 }
 
 // WorkerPool keys
-func (rk *redisKeys) WorkerPoolLock(poolName string) string {
```
luke-lombardi, line 245: was this originally for loading the pools or something?

pkg/common/keys.go
```diff
@@ -36,6 +36,7 @@ var (
 	workerNetworkLock            string = "worker:network:%s:lock"
 	workerNetworkIpIndex         string = "worker:network:%s:ip_index"
 	workerNetworkContainerIp     string = "worker:network:%s:container_ip:%s"
+	workerPoolSizerLock          string = "worker:poolsizer:%s:lock"
```
luke-lombardi, line 39: should this be pool_sizer to conform to other key names?

pkg/common/keys.go
nickpetrovic: I believe so - when we shared the worker pool configs between the gateways in v1's agent setup.

pkg/common/keys.go
luke-lombardi: worker:pool_sizer:%s:lock

pkg/scheduler/pool_sizing.go
```diff
@@ -75,7 +75,14 @@ func (s *WorkerPoolSizer) Start() {
 }
 
 // occupyAvailableMachines ensures that all manually provisioned machines always have workers occupying them
+// This only adds one worker per machine, so if a machine has more capacity, it will not be fully utilized unless
```
luke-lombardi, line 78: this makes sense 👍

luke-lombardi (changes requested):

pkg/common/keys.go
luke-lombardi: 👍

luke-lombardi (approved):

PR #715, opened 2024-11-14
Fix: Prevent logging lock not obtained messages

nickpetrovic:
These log messages just mean that another gateway or goroutine is working on this. They're emitted every second x N workers running so it creates a bunch of noise.

Resolve BE-2045

### DIFF
--- a/pkg/scheduler/pool_sizing.go
+++ b/pkg/scheduler/pool_sizing.go
@@ -1,12 +1,13 @@
 package scheduler
 
 import (
+	"errors"
 	"log"
-	"strings"
 	"time"
 
 	"github.com/beam-cloud/beta9/pkg/repository"
 	"github.com/beam-cloud/beta9/pkg/types"
+	"github.com/bsm/redislock"
 )
 
 const poolMonitoringInterval = 1 * time.Second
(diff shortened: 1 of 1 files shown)


luke-lombardi (approved):

pkg/scheduler/pool_sizing.go
```diff
@@ -66,7 +67,7 @@ func (s *WorkerPoolSizer) Start() {
 			// Handle case where we want to make sure all available manually provisioned nodes have available workers
 			if s.workerPoolConfig.Mode == types.PoolModeExternal {
 				err := s.occupyAvailableMachines()
-				if err != nil {
+				if err != nil && !strings.Contains(err.Error(), "redislock: not obtained") {
```
luke-lombardi, line 70: Approved, but I think you could also check the error type from redislock which might be cleaner instead of hardcoding the string

nickpetrovic: Gonna use that instead.

luke-lombardi (approved):

pkg/scheduler/pool_sizing.go
luke-lombardi: 👍

PR #721, opened 2024-11-19
Feat: Add docker hub support for private images

nickpetrovic:
- Add ability to use docker hub credentials to pull private images
- Improve typing for image credentials

Resolve BE-2061

### DIFF
--- a/pkg/abstractions/image/credentials.go
+++ b/pkg/abstractions/image/credentials.go
@@ -23,8 +23,12 @@ func GetRegistryToken(opts *BuildOpts) (string, error) {
 		}
 	}
 
-	// Add additional registry checks here as needed.
-	return "", fmt.Errorf("unsupported registry or missing implementation")
+	// Default to Docker Hub
+	if token, err := GetDockerHubToken(opts); err == nil {
+		return token, nil
+	} else {
+		return "", err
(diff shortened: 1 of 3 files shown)


pkg/abstractions/image/credentials.go
```diff
@@ -23,8 +23,12 @@ func GetRegistryToken(opts *BuildOpts) (string, error) {
 		}
 	}
 
-	// Add additional registry checks here as needed.
-	return "", fmt.Errorf("unsupported registry or missing implementation")
+	// Default to Docker Hub
+	if token, err := GetDockerHubToken(opts); err == nil {
```
luke-lombardi, line 27: Seems reasonable, would it be helpful to have a unit test here? 

Also, looks like python sdk tests are failing

luke-lombardi (changes requested):

pkg/abstractions/image/credentials.go
nickpetrovic: Yeah probably. Just added tests and fixed the python linting.

luke-lombardi: 👍

luke-lombardi (approved):

PR #722, opened 2024-11-19
Feat: Create fly.io template to launch beta9 in fly.io machines

jsun-m:
Add templates and setup file to spin up an entire beta9 cluster in [fly.io](https://fly.io/). 

This includes templates for:
- control plane redis
- juicefs redis
- blobcache redis
- control plane postgres
- control plane storage
- image storage

Currently, there is no built in provider to automatically provision nodes. You would have to use the `beta9 agent` to manually add a node to the `beta9` cluster.


sdk/src/beta9/cli/machine.py
```diff
@@ -177,7 +177,8 @@ def create_machine(service: ServiceClient, pool: str):
               --tailscale-url "{res.machine.tailscale_url}" \\
               --tailscale-auth "{res.machine.tailscale_auth}" \\
               --pool-name "{res.machine.pool_name}" \\
-              --provider-name "{res.machine.provider_name}"
+              --provider-name "{res.machine.provider_name}" \\
+              --gateway-url "https://{service._config.gateway_host}"
```
jsun-m, line 181: This would need to conditionally added since in `beam` the grpc host and the http host are on different subdomains

pkg/scheduler/pool_external.go
```diff
@@ -274,7 +274,8 @@ func (wpc *ExternalWorkerPoolController) createWorkerJob(workerId, machineId str
 	workerImage := fmt.Sprintf("%s/%s:%s",
 		wpc.config.Worker.ImageRegistry,
 		wpc.config.Worker.ImageName,
-		wpc.config.Worker.ImageTag,
+		// wpc.config.Worker.ImageTag,
```
luke-lombardi, line 277: revert this

pkg/scheduler/pool_external.go
```diff
@@ -370,6 +371,11 @@ func (wpc *ExternalWorkerPoolController) getWorkerEnvironment(workerId, machineI
 		podHostname = fmt.Sprintf("machine-%s.%s.%s", machineId, wpc.config.Tailscale.User, wpc.config.Tailscale.HostName)
 	}
 
+	grpcPort := 443
+	if wpc.config.GatewayService.GRPC.ExternalPort != 0 {
```
luke-lombardi, line 375: why not just set this in the config to 443 and remove the if clause?

pkg/types/config.go
```diff
@@ -289,6 +291,16 @@ type EC2ProviderConfig struct {
 	Agent        ProviderAgentConfig `key:"agent" json:"agent"`
 }
 
+type AgentRedisConfig struct {
+	Hostname string `key:"hostname" json:"hostname"`
+}
+
+type RootAgentConfig struct {
```
luke-lombardi, line 298: can we rename this?

pkg/types/config.go
```diff
@@ -289,6 +291,16 @@ type EC2ProviderConfig struct {
...
+	ControlPlaneRedis AgentRedisConfig `key:"controlPlaneRedis" json:"control_plane_redis"`
+	JuicefsRedis      AgentRedisConfig `key:"juicefsRedis" json:"juicefs_redis"`
```
luke-lombardi, line 300: Not sure I like this config structure. We're basically only using it to override the URLs, feels like there's a more elegant approach here

pkg/worker/lifecycle.go
```diff
@@ -364,6 +364,7 @@ func (s *Worker) getContainerEnvironment(request *types.ContainerRequest, option
 		fmt.Sprintf("CONTAINER_ID=%s", request.ContainerId),
 		fmt.Sprintf("BETA9_GATEWAY_HOST=%s", os.Getenv("BETA9_GATEWAY_HOST")),
 		fmt.Sprintf("BETA9_GATEWAY_PORT=%s", os.Getenv("BETA9_GATEWAY_PORT")),
+		fmt.Sprintf("BETA9_GATEWAY_TLS=%s", os.Getenv("BETA9_GATEWAY_TLS")),
```
luke-lombardi, line 367: I think we can just detect if a host has TLS enabled programatically, something like this: 

```
def is_tls_enabled(url: str) -> bool:
    """Check if a given URL supports TLS.

    Args:
        url (str): The URL to check.

    Returns:
        bool: True if TLS is enabled, False otherwise.
    """
    try:
        # Extract hostname and default to port 443 for HTTPS
        hostname = url.split("//")[-1].split("/")[0]
        port = 443

        # Create SSL context and test the connection
        context = ssl.create_default_context()
        with socket.create_connection((hostname, port), timeout=5) as sock:
            with context.wrap_socket(sock, server_hostname=hostname):
                return True
    except (ssl.SSLError, socket.error):
        return False
```

luke-lombardi (changes requested):

pkg/worker/lifecycle.go
jsun-m: this makes sense. the code here works

pkg/types/config.go
jsun-m: i'm not a fan either. especially since we don't pass this to the agent.

we do this
`remoteConfig.Database.Redis.Addrs[0] = fmt.Sprintf("%s:%d", redisHostname, 6379)`
`remoteConfig.Storage.JuiceFS.RedisURI = fmt.Sprintf("%s://:%s@%s", scheme, juicefsRedisPassword, juiceFsRedisHostname)`
`remoteConfig.BlobCache.BlobFs.Sources[idx].JuiceFS.RedisURI = remoteConfig.Storage.JuiceFS.RedisURI`

jsun-m: I added a new variable config and removed the juicefs redis, controlplane redis and blobcache urls. 
The new veriable is just called 
`DynamicServiceHosts`

that encapsulate all of them

pkg/scheduler/pool_external.go
luke-lombardi: @jsun-m

sdk/src/beta9/config.py
```diff
@@ -176,6 +180,7 @@ def prompt_for_config_context(
     prompt_gateway_port = functools.partial(
         terminal.prompt, text="Gateway Port", default=gateway_port or settings.gateway_port
     )
+    prompt_tls = functools.partial(terminal.prompt, text="TLS", default=False)
```
luke-lombardi, line 183: why is this prompted? shouldn't it be using that function below?

is_tls_enabled

luke-lombardi (changes requested):

sdk/src/beta9/config.py
jsun-m: removed

PR #731, opened 2024-11-22
Feat: Add optional retry_for argument for task queue

dleviminzi:
The `retry_for` argument can be used to provide a list of exceptions. If any of those exceptions are encountered, the task will be retried. 

It can be used like this: 
```python
from beta9 import task_queue

@task_queue(retries=2, retry_for=[AttributeError])
def hello():
    import random
    import time
    time.sleep(2)

    raise AttributeError("This is a test exception")
```

The resulting output will look something like this: 
```bash
Starting task worker[0]
Worker[0] ready
Running task <8b96b5c9-3614-4121-aaec-f0b08cc4e108>
Retrying task <8b96b5c9-3614-4121-aaec-f0b08cc4e108> after AttributeError exception
Running task <8b96b5c9-3614-4121-aaec-f0b08cc4e108>
Retrying task <8b96b5c9-3614-4121-aaec-f0b08cc4e108> after AttributeError exception
Running task <8b96b5c9-3614-4121-aaec-f0b08cc4e108>
Retry limit of 2 exceeded for task <8b96b5c9-3614-4121-aaec-f0b08cc4e108>
```


pkg/gateway/services/stub.go
```diff
@@ -18,6 +19,7 @@ import (
 )
 
 func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCreateStubRequest) (*pb.GetOrCreateStubResponse, error) {
+	log.Printf("Getting or creating stub: %v", in)
```
luke-lombardi, line 22: remove log

sdk/src/beta9/runner/taskqueue.py
```diff
@@ -304,14 +304,20 @@ def process_tasks(self, channel: Channel) -> None:
...
                     try:
                         args = task.args or []
                         kwargs = task.kwargs or {}
 
                         result = handler(context, *args, **kwargs)
-                    except BaseException:
-                        print(traceback.format_exc())
-                        task_status = TaskStatus.Error
+                    except BaseException as e:
+                        if type(e) in handler.handler.parent.retry_for:
```
luke-lombardi, line 315: it would be nice to improve the handler class so we could reference this in a way that's not so nested

sdk/src/beta9/runner/taskqueue.py
```diff
@@ -304,14 +304,20 @@ def process_tasks(self, channel: Channel) -> None:
                     result = None
                     duration = None
 
+                    encountered_exception = ""
```
luke-lombardi, line 307: caught_exception?

luke-lombardi (changes requested):

pkg/abstractions/taskqueue/keys.go
```diff
@@ -0,0 +1,56 @@
+package taskqueue
```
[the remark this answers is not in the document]
luke-lombardi: 👍

sdk/src/beta9/runner/taskqueue.py
```diff
@@ -333,15 +339,25 @@ def process_tasks(self, channel: Channel) -> None:
                             if not complete_task_response.ok:
                                 raise RunnerException("Unable to end task")
 
-                            print(f"Task completed <{task.id}>, took {duration}s")
+                            if task_status != TaskStatus.Retry:
```
luke-lombardi, line 342: This feels weird. Retry is not the common path, and you're first checking if status is retry, but only for a log. You can alternatively check for retry and if retry, just return right away instead of the else block.

sdk/src/beta9/runner/common.py
```diff
@@ -93,6 +93,30 @@ def load_from_env(cls) -> "Config":
     config: Config = Config.load_from_env()
 
 
+class ConfigProxy:
```
luke-lombardi, line 96: Not sure ConfigProxy is the right name for this

luke-lombardi (changes requested):

sdk/src/beta9/runner/common.py
dleviminzi: I changed it to parent abstraction proxy. Bit more verbose but more accurate

luke-lombardi (approved):

PR #735, opened 2024-11-25
Feat: Optionally store task result

luke-lombardi:
- Allows users to optionally store task results (value returned from a task queue handler) for up to 1 day. Can be used like so:

```
from beta9 import task_queue

@task_queue(save_result=True)
def handler():
    d = {"a": 1, "b": 2}
    return d
```

This value (`d`) will be available for up to 1 day in responses from the task API.


pkg/repository/task_redis.go
```diff
@@ -72,6 +72,16 @@ func (r *TaskRedisRepository) IsClaimed(ctx context.Context, workspaceName, stub
 	return exists > 0, nil
 }
 
+const taskResultExpiration = 3600 * 24 // 1 day
+
+func (r *TaskRedisRepository) SetTaskResult(ctx context.Context, workspaceName, stubId, taskId, result string) error {
```
luke-lombardi, line 77: We should probably limit the result size here. I can also limit that on the GRPC size as well though.

PR #742, opened 2024-11-26
Fix: Add context guard to opening runc events channel

dleviminzi:
Resolve BE-2052

This small change guards against trying to open the channel with canceled context.


nickpetrovic (approved):
I thought we already did this?

PR #745, opened 2024-11-27
fix bug with retry for causing taskqueue to spin down 

dleviminzi:
The early return caused `monitor_task.cancel()` not to be called which would snowball into the taskqueue shutting down early and the task never getting retried.

### DIFF
--- a/sdk/pyproject.toml
+++ b/sdk/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "beta9"
-version = "0.1.126"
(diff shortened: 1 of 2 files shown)


luke-lombardi (approved):

luke-lombardi (changes requested):
I'm not a fan of the way this logic is structured. The default/most common case is that the task is not retried, so it feels weird that we check for that status and then send the callback if not. Can we invert that logic somehow?

luke-lombardi (approved):

PR #750, opened 2024-12-02
feat: configure criu per worker pool

nickpetrovic:
Resolve BE-2090

### DIFF
--- a/pkg/common/config.default.yaml
+++ b/pkg/common/config.default.yaml
@@ -106,6 +106,7 @@ worker:
       priority: 1
       jobSpec:
         nodeSelector: {}
+      criuEnabled: true
       poolSizing:
         defaultWorkerCpu: 1000m
         defaultWorkerGpuType: ""
@@ -160,8 +161,7 @@ worker:
   terminationGracePeriod: 30
   addWorkerTimeout: 10m
   blobCacheEnabled: false
-  checkpointing:
-    enabled: false
+  criu:
     storage:
       mountPath: /data/checkpoints
       mode: local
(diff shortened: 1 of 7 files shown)


pkg/worker/worker.go
```diff
@@ -251,6 +250,24 @@ func NewWorker() (*Worker, error) {
 	}, nil
 }
 
+func (s *Worker) IsCRIUAvailable() bool {
```
luke-lombardi, line 253: can we move this to the cr.go file (that's where the rest of the checkpoint restore stuff is, for the most part)

pkg/worker/worker.go
```diff
@@ -251,6 +250,24 @@ func NewWorker() (*Worker, error) {
...
+	if s.cedanaClient == nil {
+		return false
+	}
+
+	poolName := os.Getenv("WORKER_POOL_NAME")
```
luke-lombardi, line 258: are we just passing the whole pool cause in the future it might be more convenient than adding another extra env var just for criu enabled?

luke-lombardi (changes requested):
looks good, a couple small things

pkg/worker/worker.go
nickpetrovic: Updated.

pkg/common/config.default.yaml
```diff
@@ -106,6 +106,7 @@ worker:
       priority: 1
       jobSpec:
         nodeSelector: {}
+      criuEnabled: true
```
luke-lombardi, line 109: should this be off by default?

luke-lombardi (changes requested):

pkg/common/config.default.yaml
nickpetrovic: Yep good catch. Updated.

luke-lombardi (approved):

PR #751, opened 2024-12-03
Feat: Handle generator functions in realtime decorator

jsun-m:


sdk/src/beta9/abstractions/endpoint.py
```diff
@@ -439,18 +440,25 @@ async def _heartbeat():
 
                         internal_asgi_app.input_queue.put(data)
 
+                        async def _handle_output(output):
```
luke-lombardi, line 443: Were you able to test this with all these cases?

luke-lombardi (changes requested):
I think it looks good just want to make sure we tested every type of output

sdk/src/beta9/abstractions/endpoint.py
jsun-m: Yep. str and dict/list works for both returnable functions and generators

luke-lombardi (approved):

PR #758, opened 2024-12-04
Fix: Dont coalesce pip install commands if they contain flags

dleviminzi:
If we have an image like the following, we should not combing the python package installs because of the `--no-deps` flag. It will break the prior packages install unintentionally. This PR flushes any commands if a flag command is encountered and flushes flag commands immediately. 

```
image = (
    Image(python_version="python3.11")
    .micromamba()
    .add_micromamba_packages(["pytorch-cuda=12.1", "pytorch", "cudatoolkit", "xformers", "numpy"], ["pytorch", "nvidia", "xformers", "conda-forge"])
    .add_python_packages(["unsloth[colab-new] @ git+https://github.com/unslothai/unsloth.git", "--no-deps trl peft accelerate bitsandbytes"])
)
```

### DIFF
--- a/pkg/abstractions/image/build.go
+++ b/pkg/abstractions/image/build.go
@@ -622,12 +622,27 @@ func parseBuildSteps(buildSteps []BuildStep, pythonVersion string) []string {
 			commands = append(commands, step.Command)
 		}
 
+		flagCmd := containsFlag(step.Command)
+
+		// Flush any pending pip or mamba groups
+		if pipStart != -1 && (step.Type != pipCommandType || flagCmd) {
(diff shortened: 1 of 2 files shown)


pkg/abstractions/image/build.go
```diff
@@ -622,12 +622,27 @@ func parseBuildSteps(buildSteps []BuildStep, pythonVersion string) []string {
 			commands = append(commands, step.Command)
 		}
 
+		flagCmd := containsFlag(step.Command)
```
luke-lombardi, line 625: can we add a unit test for this?

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #762, opened 2024-12-05
Feat: Multi-gpu workloads

luke-lombardi:
- Enable multi-gpu workloads (on multi-gpu workers)


pkg/repository/backend_postgres_migrations/019_add_workspace_multi_gpu.go
```diff
@@ -0,0 +1,21 @@
+package backend_postgres_migrations
+
+import (
+	"database/sql"
+
+	"github.com/pressly/goose/v3"
+)
+
+func init() {
+	goose.AddMigration(upAddFieldWorkspaceMultiGpuEnabled, downDropFieldWorkspaceMultiGpuEnabled)
```
nickpetrovic, line 10: I think this is deprecated. Can we use the AddMigrationContext function instead?

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -132,6 +134,9 @@ def __init__(
         self.checkpoint_enabled = checkpoint_enabled
         self.extra: dict = {}
 
+        if (self.gpu != "" or len(self.gpu) > 0) and self.gpu_count == 0:
```
nickpetrovic, line 137: Have we tested this when `gpu` is a List?

luke-lombardi: Will confirm

PR #767, opened 2024-12-06
Fix: Dont init multiprocessing Value unless it will be used

dleviminzi:
It seems `Value` is responsible for leaking the semaphore. Even though it is not used in the serve, the value is initialized on import because it is global. This tweak fixes the leak. 

Before
```
=> Stopping serve container
Goodbye 👋
/opt/homebrew/Cellar/python@3.12/3.12.7_1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```
After
```
=> Stopping serve container
Goodbye 👋
```

### DIFF
--- a/sdk/src/beta9/runner/common.py
+++ b/sdk/src/beta9/runner/common.py
@@ -163,12 +163,8 @@ class FunctionContext:
 
 
 workers_ready = None
-
-
-def initialize_workers_ready():
-    global workers_ready
-    if workers_ready is None:
-        workers_ready = Value("i", 0)
+if is_remote():
+    workers_ready = Value("i", 0)
 
 
 class FunctionHandler:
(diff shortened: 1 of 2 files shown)


sdk/src/beta9/runner/common.py
```diff
@@ -391,6 +397,8 @@ def __exit__(self, *_, **__):
 
 
 def wait_for_checkpoint():
+    initialize_workers_ready()
```
luke-lombardi, line 400: I'm not sure we should do it like this, because then we're creating the Value in the subprocess vs before it launches and I don't think it will have the same effect. We could just check if we're in the remote environment and only initialize it if so maybe

luke-lombardi (changes requested):

sdk/src/beta9/runner/common.py
dleviminzi: I can make that change as well. However, it doesn't seem like the Value is used anywhere else besides `wait_for_checkpoint`. Does it even need to be global (not sure I totally follow its usage)?

luke-lombardi: It's used to synchronize worker processes as they are spinning up before auto checkpoint. I think it has to either be global or passed into the worker processes when they start (either would work, global was simpler)

dleviminzi: Ahh got it. Updated to use `is_remote` check instead of the current method

luke-lombardi (approved):

PR #771, opened 2024-12-10
Feat: Add auxiliary state to help check the runtime status of a stub

jsun-m:
Added a new redis value
`scheduler:stub:unhealthy_state:[stub_id]`

That stores the state of a given deployment/stub. Currently state can be `degraded` or `warning`.
A nonexistent value for this variable indicates that the deployment currently has no runtime issues and is healthy.

`degraded` occurs once the `max container failure` threshold is reached
`warning` occurs when at least 1 failed container occurs in the past 5 minutes (5 minutes before the container exit code value expires)
`healthy` is the state when 0 containers failed in the past 5 minutes. !!!Caveat: That means even if the containers spin down after failing and there are 0 running containers, it is considered `healthy`


pkg/abstractions/common/instance.go
```diff
@@ -287,3 +298,72 @@ func (i *AutoscaledInstance) State() (*AutoscaledInstanceState, error) {
 	state.FailedContainers = failedContainers
 	return &state, nil
 }
+
+func (i *AutoscaledInstance) HandleDeploymentDegraded(stubId string, containers []string) {
```
luke-lombardi: It feels weird / redundant that we have an additional state, and an event. I would either remove the redis key or remove the event

pkg/abstractions/common/instance.go
```diff
@@ -287,3 +298,72 @@ func (i *AutoscaledInstance) State() (*AutoscaledInstanceState, error) {
...
+func (i *AutoscaledInstance) HandleDeploymentWarning(stubId string, containers []string) {
+	var state string
+	state, err := i.ContainerRepo.GetStubUnhealthyState(stubId)
+	if err != nil {
+		if err != redis.Nil {
+			log.Printf("<%s> failed to get unhealthy state\n", i.Name)
+			return
+		}
+
+		state = "healthy"
```
luke-lombardi: I'm not sure I would have a new "state", I almost feel like the client can derive the state implicitly, if a certain number of failed containers has been met.

luke-lombardi (changes requested):

pkg/abstractions/common/instance.go
```diff
@@ -23,7 +24,9 @@ type AutoscaledInstanceState struct {
 	RunningContainers  int
 	PendingContainers  int
 	StoppingContainers int
-	FailedContainers   int
+
+	// TODO: We can potentially store the strings for RUNNING, PENDING, STOPPING, containers as well
+	FailedContainers []string
```
luke-lombardi, line 29: we already have the ability to GetActiveContainers by stub right? this comment may be redundant, although I see your point

pkg/common/keys.go
```diff
@@ -21,6 +21,7 @@ var (
 	schedulerContainerLock           string = "scheduler:container:lock:%s"
 	schedulerContainerExitCode       string = "scheduler:container:exit_code:%s"
 	schedulerCheckpointState         string = "scheduler:checkpoint_state:%s:%s"
+	schedulerStubUnhealthyState      string = "scheduler:stub:unhealthy_state:%s"
```
luke-lombardi, line 24: i feel like this should be scheduleStubState (and it could contain healthy contain count, and failed containers), that way client can infer what the state means. to me that sounds cleaner than "unhealthy" state

luke-lombardi (changes requested):

pkg/abstractions/common/instance.go
jsun-m: i had to add the GetActiveContainersByStubId. It replaces the GetActiveContainer**Count**ByStubId function

pkg/common/keys.go
jsun-m: that makes sense. However i was only planning on having this state if the state was non consider healthy. 
If this doesn't exist, the state should be healthy

luke-lombardi: yeah the name just feels wrong imo

pkg/repository/base.go
```diff
@@ -180,6 +183,8 @@ type EventRepository interface {
 	PushRunStubEvent(workspaceId string, stub *types.Stub)
 	PushTaskUpdatedEvent(task *types.TaskWithRelated)
 	PushTaskCreatedEvent(task *types.TaskWithRelated)
+	PushStubStateNotHealthy(workspaceId string, stubId string, currentState, previousState string, reason string, failedContainers []string)
```
luke-lombardi, line 186: can we maybe call this PushStubStateDegraded?

luke-lombardi (changes requested):

pkg/abstractions/common/instance.go
```diff
@@ -287,3 +292,47 @@ func (i *AutoscaledInstance) State() (*AutoscaledInstanceState, error) {
 	state.FailedContainers = failedContainers
 	return &state, nil
 }
+
+func (i *AutoscaledInstance) handleUnhealthyDeploymentEvents(failedContainers []string) {
```
luke-lombardi, line 296: since this is also sending events like i.HandleDeploymentHealthy(i.Stub.ExternalId), should it just be called handledeploymentevents?

pkg/abstractions/common/instance.go
```diff
@@ -287,3 +292,47 @@ func (i *AutoscaledInstance) State() (*AutoscaledInstanceState, error) {
...
+	if len(failedContainers) >= i.FailedContainerThreshold {
+		i.HandleDeploymentUnhealthy(i.Stub.ExternalId, types.StubStateDegraded, "failed container threshold", failedContainers)
+	} else if len(failedContainers) > 0 {
+		i.HandleDeploymentUnhealthy(i.Stub.ExternalId, types.StubStateWarning, "one or more containers failed", failedContainers)
+	} else if len(failedContainers) == 0 {
+		i.HandleDeploymentHealthy(i.Stub.ExternalId)
+	}
+}
+
+func (i *AutoscaledInstance) HandleDeploymentUnhealthy(stubId, currentState, reason string, containers []string) {
```
luke-lombardi, line 306: i don't think this method is exported, so should be lowercase

luke-lombardi (changes requested):

pkg/abstractions/common/instance.go
jsun-m: yea that makes sense

luke-lombardi (approved):

PR #773, opened 2024-12-10
Feat: Allow users to bring their own Dockerfile

dleviminzi:
Resolve BE-2084

This feature adds a new operator to the `Image` abstraction that allows users to specify a path to a Dockerfile that they wish to use as the base of their Image. We use `buildah` and `umoci` to build their Dockerfile and gather the bundle needed to run it. 

It can be used as follows: 
```python
from beta9 import Image, endpoint

image = Image().from_dockerfile("./Dockerfile").add_python_packages(["numpy"])


@endpoint(image=image, name="test_dockerfile")
def handler():
    return "pass"
```
Where `./Dockerfile` is: 
```
FROM ubuntu:22.04
RUN echo hello
ENV FOO=bar
```

Note: when testing locally, you might need to build gateway again to get `make start` to work because the go version is bumped.


pkg/abstractions/image/build.go
```diff
@@ -156,11 +202,23 @@ func (i *BaseImage) String() string {
...
+		// Need to set base image registry, name, and tag to something that will be unique
+		opts.BaseImageRegistry = authInfo.Workspace.ExternalId
+		opts.BaseImageName = opts.Dockerfile
+		opts.BaseImageTag = "latest"
+		opts.addPythonRequirements()
+		dockerfile = &opts.Dockerfile
+	case opts.ExistingImageUri != "":
 		err := b.handleCustomBaseImage(opts, outputChan)
 		if err != nil {
+			// FIXME: Why do we pass another message here? Should we just remove this since we already pass the error in the handleCustomBaseImage method?
```
dleviminzi, line 221: Highlighting this, which I'm still not sure about.

pkg/abstractions/image/build.go
```diff
@@ -121,6 +167,12 @@ func (b *Builder) GetImageId(opts *BuildOpts) (string, error) {
 			fmt.Fprint(h, envVar)
 		}
 	}
+	if opts.Dockerfile != "" {
+		fmt.Fprint(h, opts.Dockerfile)
+	}
+	if opts.BuildCtxObject != "" {
+		fmt.Fprint(h, opts.BuildCtxObject)
+	}
```
dleviminzi, line 175: This feels pretty hacky to me, but I don't think we can change hash body.

pkg/worker/image.go
```diff
@@ -256,6 +256,75 @@ func (c *ImageClient) InspectAndVerifyImage(ctx context.Context, sourceImage str
...
+
+	for _, dir := range requiredContainerDirectories {
+		fullPath := filepath.Join(tempBundlePath, "rootfs", dir)
+		err := os.MkdirAll(fullPath, 0755)
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("creating /%s directory", dir))
+		}
+	}
+
+	defer os.RemoveAll(tempBundlePath)
```
nickpetrovic, line 319: Why is this defer here and not right after we create tempBundlePath?

dleviminzi: Moved it up

docker/Dockerfile.worker
```diff
@@ -93,6 +116,8 @@ RUN apt-get install -y --no-install-recommends nvidia-container-toolkit-base nvi
 
 RUN apt-get update && apt-get install -y fuse3 libfuse2 libfuse3-dev libfuse-dev bash-completion
 
+RUN apt-get install -y libbtrfs-dev libgpgme11-dev podman
```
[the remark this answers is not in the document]
dleviminzi: Hmm this actually doesn't seem to work. 

```
worker-build-cbb72112-zhzfj worker time="2024-12-11T19:09:41Z" level=error msg="Failed to built-in GetDriver graph overlay /tmp/866741800/image"
worker-build-cbb72112-zhzfj worker time="2024-12-11T19:09:41Z" level=error msg="Failed to built-in GetDriver graph btrfs /tmp/866741800/image"
worker-build-cbb72112-zhzfj worker STEP 1/4: FROM ubuntu:22.04
worker-build-cbb72112-zhzfj worker Error: creating build container: short-name "ubuntu:22.04" did not resolve to an alias and no containers-registries.conf(5) was found
```

Working on alternative ways to at least remove podman as a dependency

docker/Dockerfile.gateway
```diff
@@ -25,7 +25,7 @@ FROM base AS build
 
 WORKDIR /workspace
 
-RUN apt-get install -y libfuse3-dev && \
+RUN apt-get install -y libfuse3-dev libbtrfs-dev libgpgme-dev && \
```
luke-lombardi, line 28: I think there may have been an issue with libgpg + cedana, we should look into if it causes any issues with C/R

docker/Dockerfile.worker
```diff
@@ -93,6 +115,11 @@ RUN apt-get install -y --no-install-recommends nvidia-container-toolkit-base nvi
 
 RUN apt-get update && apt-get install -y fuse3 libfuse2 libfuse3-dev libfuse-dev bash-completion
 
+RUN apt-get install -y libbtrfs-dev libgpgme11-dev
```
luke-lombardi, line 118: same comment as above

pkg/abstractions/image/build.go
```diff
@@ -72,17 +74,61 @@ func (o *BuildOpts) String() string {
 	fmt.Fprintf(&b, "  \"BaseImageName\": %q,", o.BaseImageName)
 	fmt.Fprintf(&b, "  \"BaseImageTag\": %q,", o.BaseImageTag)
 	fmt.Fprintf(&b, "  \"BaseImageCreds\": %q,", o.BaseImageCreds)
+	fmt.Fprintf(&b, "  \"ExistingImageUri\": %q,", o.ExistingImageUri)
+	fmt.Fprintf(&b, "  \"ExistingImageCreds\": %#v,", o.ExistingImageCreds)
+	fmt.Fprintf(&b, "  \"Dockerfile\": %q,", o.Dockerfile)
+	fmt.Fprintf(&b, "  \"BuildCtxObject\": %q,", o.BuildCtxObject)
```
luke-lombardi, line 80: What exactly is in this var?

pkg/abstractions/image/build.go
```diff
@@ -201,6 +263,8 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
 		ImageId:          baseImageId,
 		SourceImage:      &sourceImage,
 		SourceImageCreds: opts.BaseImageCreds,
+		Dockerfile:       dockerfile,
+		BuildCtxObject:   &opts.BuildCtxObject,
```
luke-lombardi, line 267: I really don't like that this is top level on the ContainerRequest. It might seem out of scope but I think we should nest these under a BuildOptions struct on the container request

luke-lombardi (changes requested):

pkg/abstractions/image/build.go
dleviminzi: The object ID of the synced build context directory

pkg/types/scheduler.go
```diff
@@ -101,6 +111,16 @@ func (c *ContainerRequest) RequiresGPU() bool {
 	return len(c.GpuRequest) > 0 || c.Gpu != ""
 }
 
+func (c *ContainerRequest) ImageSourceType() ImageSourceType {
+	if c.Dockerfile != nil {
+		return ImageSourceTypeBuild
```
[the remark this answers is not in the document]
dleviminzi: I just removed this entirely as I don't think it was really needed

pkg/worker/lifecycle.go
```diff
@@ -151,18 +151,31 @@ func (s *Worker) RunContainer(request *types.ContainerRequest) error {
...
-	if err != nil && request.SourceImage != nil {
-		log.Printf("<%s> - lazy-pull failed, pulling source image: %s\n", containerId, *request.SourceImage)
-		err = s.imageClient.PullAndArchiveImage(context.TODO(), *request.SourceImage, request.ImageId, request.SourceImageCreds)
-		if err == nil {
-			err = s.imageClient.PullLazy(request)
-		}
-	}
 	if err != nil {
-		return err
+		switch request.ImageSourceType() {
```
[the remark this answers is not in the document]
dleviminzi: Done

pkg/abstractions/image/build.go
dleviminzi: Removed it.

pkg/abstractions/image/build.go
dleviminzi: Bumped it out

nickpetrovic (approved):

pkg/types/scheduler.go
```diff
@@ -79,7 +79,17 @@ type ContainerState struct {
 	Memory      int64           `redis:"memory" json:"memory"`
 }
 
+type ImageSourceType string
```
luke-lombardi, line 82: is this being used ?

pkg/worker/lifecycle.go
```diff
@@ -211,6 +222,28 @@ func (s *Worker) RunContainer(request *types.ContainerRequest) error {
 	return nil
 }
 
+func (s *Worker) buildOrPullImage(request *types.ContainerRequest, containerId string, outputChan chan common.OutputMsg) error {
+	switch {
+	case request.Dockerfile != nil:
+		log.Printf("<%s> - lazy-pull failed, building image from Dockerfile\n", containerId)
```
luke-lombardi, line 228: can you clean up the spacing a bit here?

pkg/worker/lifecycle.go
```diff
@@ -676,3 +695,20 @@ func (s *Worker) watchOOMEvents(ctx context.Context, containerId string, output
 		}
 	}
 }
+
+func getBuildContext(request *types.ContainerRequest) (string, error) {
+	buildCtxPath := "."
+	if request.BuildCtxObject != nil {
+		err := common.ExtractObjectFile(context.TODO(), *request.BuildCtxObject, request.Workspace.Name)
```
luke-lombardi, line 702: should this be a context.TODO?

pkg/worker/lifecycle.go
```diff
@@ -211,6 +222,28 @@ func (s *Worker) RunContainer(request *types.ContainerRequest) error {
...
 
+func (s *Worker) buildOrPullImage(request *types.ContainerRequest, containerId string, outputChan chan common.OutputMsg) error {
+	switch {
+	case request.Dockerfile != nil:
+		log.Printf("<%s> - lazy-pull failed, building image from Dockerfile\n", containerId)
+		buildCtxPath, err := getBuildContext(request)
+		if err != nil {
+			return err
+		}
+		if err := s.imageClient.BuildAndArchiveImage(context.TODO(), outputChan, *request.Dockerfile, request.ImageId, buildCtxPath); err != nil {
```
luke-lombardi, line 233: I don't think request should have dockerfile on it directly. Shouldn't it be request.buildoptions.dockerfile?

pkg/worker/lifecycle.go
```diff
@@ -211,6 +222,28 @@ func (s *Worker) RunContainer(request *types.ContainerRequest) error {
...
+			return err
+		}
+	case request.SourceImage != nil:
```
luke-lombardi, line 236: same as above, does this build...?

pkg/worker/lifecycle.go
```diff
@@ -676,3 +695,20 @@ func (s *Worker) watchOOMEvents(ctx context.Context, containerId string, output
...
+			return "", err
+		}
+		buildCtxPath = filepath.Join(types.DefaultExtractedObjectPath, request.Workspace.Name, *request.BuildCtxObject)
+	}
+	return buildCtxPath, nil
+}
+
+// isBuildRequest checks if the sourceImage or Dockerfile field is not-nil, which means the container request is for a build container
+func isBuildRequest(request *types.ContainerRequest) bool {
+	return request.SourceImage != nil || request.Dockerfile != nil
```
luke-lombardi, line 713: same as above, this should match the nested build options struct

pkg/worker/runc_console.go
```diff
@@ -51,7 +51,7 @@ func NewConsoleWriter(writer io.Writer) (*ConsoleWriter, error) {
 		defer socket.Close()
 
 		// Get the master file descriptor from runC.
-		master, err := utils.RecvFd(socket)
+		master, err := utils.RecvFile(socket)
```
luke-lombardi, line 54: why did this change?

luke-lombardi (changes requested):
left some final comments

pkg/worker/runc_console.go
dleviminzi: The old function no longer exists. It was just renamed to this.

pkg/worker/lifecycle.go
dleviminzi: It can be. I embedded the build options struct so that its members could be directly accessed. We can name it so that the access path is as you describe though.

dleviminzi: Named it

pkg/types/scheduler.go
dleviminzi: removed

pkg/worker/lifecycle.go
dleviminzi: switched to worker `s.ctx`

pkg/worker/lifecycle.go
```diff
@@ -136,19 +136,30 @@ func (s *Worker) clearContainer(containerId string, request *types.ContainerRequ
...
+		log.Printf("<%s> - failed to remove container state: %v\n", containerId, err)
+	}
+}
+
 // Spawn a single container and stream output to stdout/stderr
-func (s *Worker) RunContainer(request *types.ContainerRequest) error {
+func (s *Worker) RunContainer(request *types.ContainerRequest) (err error) {
 	containerId := request.ContainerId
+
+	defer func() {
```
luke-lombardi, line 157: I don't think you need this, because there is already a call to clearcontainer upstream of this:

```
err := s.RunContainer(request)
				if err != nil {
					log.Printf("Unable to run container <%s>: %v\n", containerId, err)

					// Set a non-zero exit code for the container (both in memory, and in repo)
					exitCode := 1

					serr, ok := err.(*types.ExitCodeError)
					if ok {
						exitCode = serr.ExitCode
					}

					err := s.containerRepo.SetContainerExitCode(containerId, exitCode)
					if err != nil {
						log.Printf("<%s> - failed to set exit code: %v\n", containerId, err)
					}

					s.containerLock.Unlock()
					s.clearContainer(containerId, request, exitCode)
					continue
				}
```

luke-lombardi (changes requested):

pkg/worker/lifecycle.go
dleviminzi: ahh I see. removed

pkg/types/scheduler.go
```diff
@@ -79,7 +79,15 @@ type ContainerState struct {
...
 
+type BuildOptions struct {
+	SourceImage      *string `json:"source_image"`
+	Dockerfile       *string `json:"dockerfile"`
+	BuildCtxObject   *string `json:"build_context"`
+	SourceImageCreds string  `json:"source_image_creds"`
+}
+
 type ContainerRequest struct {
+	BuildOptions      BuildOptions    `json:"build_options"`
```
luke-lombardi, line 90: nit pick but can move this down in the struct

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #775, opened 2024-12-11
Fix: Add bzip for Mamba and remove distutils for Python 3.12

dleviminzi:
Resolve BE-2125

This is a small change to the runner dockerfiles to fix Mamba builds by adding the bzip dependency and to fix the Python 3.12 build by removing disutils (see below). 

> Of note, the distutils package has been removed from the standard library.

https://docs.python.org/3/whatsnew/3.12.html


nickpetrovic (approved):
lgtm as long as non-mamba builds also work.

dleviminzi:
> lgtm as long as non-mamba builds also work.

This change was only additive in terms of supporting Mamba (just added bzip). The removal of distutils was just fixing an issue I noticed for the normal 3.12 build while looking into the Mamba builds.

PR #777, opened 2024-12-11
Fix: Use heartbeats to correct for request token drift

luke-lombardi:
- Uses key events + request heartbeats to correct for token drift after a particular gateway crashed unexpectedly

The consequence of this is that if a gateway that was actively handling a request crashes or is forcibly terminated, there will be a 30 delay before any active containers that were handling requests have the token count incremented. This should fix "bricked" containers that had an inaccurately low token count.

### DIFF
--- a/pkg/abstractions/endpoint/task.go
+++ b/pkg/abstractions/endpoint/task.go
@@ -74,10 +74,15 @@ func (t *EndpointTask) Cancel(ctx context.Context, reason types.TaskCancellation
 }
 
 func (t *EndpointTask) HeartBeat(ctx context.Context) (bool, error) {
-	heartbeatKey := Keys.endpointRequestHeartbeat(t.msg.WorkspaceName, t.msg.StubId, t.msg.TaskId)
-	exists, err := t.es.rdb.Exists(ctx, heartbeatKey).Result()
+	task, err := t.es.backendRepo.GetTask(ctx, t.msg.TaskId)
 	if err != nil {
-		return false, fmt.Errorf("failed to retrieve endpoint heartbeat key <%v>: %w", heartbeatKey, err)
+		return false, err
+	}
+
+	key := Keys.endpointRequestHeartbeat(t.msg.WorkspaceName, t.msg.StubId, t.msg.TaskId, task.ContainerId)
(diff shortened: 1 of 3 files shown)


pkg/abstractions/endpoint/task.go
```diff
@@ -74,10 +74,15 @@ func (t *EndpointTask) Cancel(ctx context.Context, reason types.TaskCancellation
...
 
 func (t *EndpointTask) HeartBeat(ctx context.Context) (bool, error) {
-	heartbeatKey := Keys.endpointRequestHeartbeat(t.msg.WorkspaceName, t.msg.StubId, t.msg.TaskId)
-	exists, err := t.es.rdb.Exists(ctx, heartbeatKey).Result()
+	task, err := t.es.backendRepo.GetTask(ctx, t.msg.TaskId)
+	if err != nil {
+		return false, err
+	}
+
+	key := Keys.endpointRequestHeartbeat(t.msg.WorkspaceName, t.msg.StubId, t.msg.TaskId, task.ContainerId)
```
luke-lombardi, line 82: It should be safe to do this, because heartbeat is only called if the task is claimed (which means the task has an assigned container ID): ```
				heartbeat, err := task.HeartBeat(ctx)
				if err != nil {
					continue
				}
```

File: pkg/abstractions/endpoint/task.go [hunk 2]
```diff
@@ -74,7 +74,12 @@ func (t *EndpointTask) Cancel(ctx context.Context, reason types.TaskCancellation
 }
 
 func (t *EndpointTask) HeartBeat(ctx context.Context) (bool, error) {
-	heartbeatKey := Keys.endpointRequestHeartbeat(t.msg.WorkspaceName, t.msg.StubId, t.msg.TaskId)
+	task, err := t.es.backendRepo.GetTask(ctx, t.msg.TaskId)
```
luke-lombardi, line 77: Downside is that we'll have to retrieve the task, which is not great, but the heartbeat check only happens every 5 seconds right now so if requests are faster than that they'll never hit these check

nickpetrovic (approved):

PR #790, opened 2024-12-17
Chore: Clean up gateway config

jsun-m:
Change 

```
gateway:
  externalHost: gateway.stage.beam.cloud
  externalURL: http://localhost:1994
  ....
  grpc:
  ....
  http:
  ....
```
to 
```
gateway:
  host: beta9-gateway.beta9
  ....
  grpc:
    externalHost: localhost
    externalPort: 1993
    tls: false
    ....
  http:
    externalHost: localhost
    externalPort: 1994
    tls: false
    ....
```

The ExternalURL value was less flexible

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,7 +10,5 @@
     "[proto3]": {
         "editor.defaultFormatter": "xaver.clang-format",
         "editor.formatOnSave": true
-    },
-    "python.defaultInterpreterPath": "sdk/.venv/bin/python",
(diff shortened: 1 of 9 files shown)


nickpetrovic (approved):
We need to make sure the config in stage and prod are updated before this is rolled out, otherwise, this looks good.

jsun-m:
> We need to make sure the config in stage and prod are updated before this is rolled out, otherwise, this looks good.

I'm going to retain both the old config and new config temporarily

PR #792, opened 2024-12-18
Feat: Build secrets

dleviminzi:
Resolve BE-2149

This PR adds support for referencing secrets during builds. 

First, create a secret. 
```bash
beta9 secret create TEST TEST
```

Then, use `with_secrets` to flag that you would like it to be available during the build. 
```python
from beta9 import Image, endpoint

image = (
    Image()
    .with_secrets(["TEST"])
    .add_commands(
        [
            "echo $TEST",
        ]
    )
)


@endpoint(image=image, name="secret-test")
def handler():
    return "pass"
```

The build logs will look like this: 
```
=> Building image 
Waiting for build container to start...
Python 3.10.16
TEST
...
```


pkg/worker/runc_server.go
```diff
@@ -111,6 +111,8 @@ func (s *RunCServer) RunCExec(ctx context.Context, in *pb.RunCExecRequest) (*pb.
 	if !exists {
 		return &pb.RunCExecResponse{Ok: false}, nil
 	}
+	process.Env = append(process.Env, instance.Spec.Process.Env...)
```
dleviminzi, line 114: I added this env here as well. It is not necessary, but I thought it was strange that env included using `with_env` or `env_vars` was not available.

pkg/abstractions/image/image.go
```diff
@@ -149,6 +169,21 @@ func (is *RuncImageService) BuildImage(in *pb.BuildImageRequest, stream pb.Image
 	return nil
 }
 
+func (is *RuncImageService) parseBuildSecrets(ctx context.Context, secrets []string, authInfo *auth.AuthInfo) ([]string, error) {
```
jsun-m, line 172: nitpick but can this be renamed to `retrieveBuildSecrets`?

dleviminzi: done!

PR #793, opened 2024-12-18
Feat: Export workspace gateway config for configuration in dashboard

jsun-m:
```
❯ beta9 config export-config
{
    "gateway_http_url": "localhost",
    "gateway_http_port": 1994,
    "gateway_grpc_url": "localhost",
    "gateway_grpc_port": 1993,
    "workspace_id": "d78b6cfa-1357-45e8-b61f-07edef33f71e",
    "token": "********************"
}
```

This is the simplified way to configure a self-hosted beam cluster into the management dashboard


sdk/src/beta9/cli/config.py
```diff
@@ -175,3 +179,38 @@ def select_context(name: str, config_path: Path):
     save_config(contexts=contexts, path=config_path)
 
     terminal.success(f"Default context updated with '{name}'.")
+
+
+@management.command(
+    name="export-config",
```
nickpetrovic, line 185: Can we name this just "export"? I think this command being under the config subcommand is enough to say what kind of export this is. e.g. `beam config export`

sdk/src/beta9/cli/config.py
```diff
@@ -175,3 +179,38 @@ def select_context(name: str, config_path: Path):
...
+    help="Export the current workspace configuration",
+)
+@click.option(
+    "--name",
```
nickpetrovic, line 189: Can we just do a positional argument instead? This is simple enough to just need that.

sdk/src/beta9/cli/config.py
```diff
@@ -175,3 +179,38 @@ def select_context(name: str, config_path: Path):
...
+    name: str,
+):
+    contexts = load_config()
+    context = contexts.get(name)
+
+    if not context:
+        terminal.error(f"Context '{name}' does not exist.")
+
+    resp = service.gateway.export_workspace_config(ExportWorkspaceConfigRequest())
+    config = {
```
nickpetrovic, line 207: Can you look at using `resp.to_json(indent=2, casing=Casing.SNAKE)` instead of the json.dumps and print at the end?

jsun-m: i need to add in the `token` variable before dumpnig it

nickpetrovic: Oh right, so maybe to_dict, then json.dump it?

jsun-m: done

nickpetrovic (approved):

PR #797, opened 2024-12-19
Fix: Write response headers from copy of internal response's headers

dleviminzi:
Instead of directly writing the headers into the final response, make a copy and then write from the copy. 

> fatal error: concurrent map iteration and map write


luke-lombardi (approved):
@dleviminzi can you test in stage to confirm that headers returned from an asgi app etc are still propagated properly.

PR #799, opened 2024-12-19
Fix: Handle cancelled requests correctly (superfluous write header issue)

dleviminzi:
There are two primary changes: 
* only use buffered response writing when necessary (sse, multipart, etc) otherwise just use io.Copy
* check ctx before writing other types of responses

To reproduce the superfluous write header issue locally do the following:

1. Deploy the following App.
```python
from beta9 import QueueDepthAutoscaler, endpoint


@endpoint(
    name="endpoint-test-app",
    cpu=2,
    memory=2048,
    autoscaler=QueueDepthAutoscaler(max_containers=3, tasks_per_container=10),
)
def handler():
    import random
    import time

    time.sleep(random.randint(0, 2))
    return "ok"
```
2. Create a vegeta urls.txt
```
POST http://localhost:1994/endpoint/endpoint-test-app/v1
Authorization: Bearer TOKEN
Content-Type: application/json
```
3. Attack with vegeta (with a timeout of 1s)
```bash
vegeta attack -timeout=1s -connections 5 -rate 5 -duration 1m -targets urls.txt
```

You will see logs like this 
```bash
7:20PM INF request method=POST status=200 uri=/endpoint/endpoint-test-app/v1
2024/12/19 19:20:18 http: superfluous response.WriteHeader call from github.com/labstack/echo/v4.(*Response).WriteHeader (response.go:65)
7:20PM INF request method=POST status=200 uri=/endpoint/endpoint-test-app/v1
```

With the changes in this PR, the issue is fixed. 

You can confirm that streamed responses for asgi apps still work as well.
1. Deploy the following asgi app
```python
from beta9 import Image, asgi


@asgi(name="sse-example-app", image=Image().add_python_packages(["sse_starlette"]))
def handler():
    import asyncio
    from datetime import datetime

    from fastapi import FastAPI
    from sse_starlette.sse import EventSourceResponse

    app = FastAPI()

    @app.get("/")
    async def root():
        return {"message": "Welcome to SSE Example"}

    @app.get("/events")
    async def events():
        async def event_generator():
            while True:
                # Yield a message every 2 seconds
                current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
                yield {"event": "update", "data": f"Server Time: {current_time}"}
                await asyncio.sleep(2)

        return EventSourceResponse(event_generator())

    return app
```

2. Make a request to `/events`
```bash
curl -X GET -N 'http://localhost:1994/asgi/sse-example-app/v2/events' \
-H 'Authorization: Bearer TOKEN'
```

The streamed response will look like this:
```bash
event: update
data: Server Time: 2024-12-19 19:54:09

event: update
data: Server Time: 2024-12-19 19:54:11

event: update
data: Server Time: 2024-12-19 19:54:13
```

### DIFF
--- a/pkg/abstractions/endpoint/buffer.go
+++ b/pkg/abstractions/endpoint/buffer.go
@@ -1,6 +1,7 @@
 package endpoint
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
@@ -17,6 +18,7 @@ import (
 	"github.com/gorilla/websocket"
 	"github.com/labstack/echo/v4"
 	"github.com/redis/go-redis/v9"
+	"github.com/rs/zerolog/log"
 
 	abstractions "github.com/beam-cloud/beta9/pkg/abstractions/common"
 	"github.com/beam-cloud/beta9/pkg/common"
@@ -414,7 +416,10 @@ func (rb *RequestBuffer) handleWSRequest(req *request, c container) {
 }
 
 func (rb *RequestBuffer) handleHttpRequest(req *request, c container) {
+	ctx := req.ctx.Request().Context()
+
 	request := req.ctx.Request()
+	response := req.ctx.Response()
 
 	requestBody := request.Body
 	if !rb.isASGI {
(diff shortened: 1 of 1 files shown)


luke-lombardi:
@dleviminzi I handled this in a separate PR (in a similar way) -- going to close this.

PR #800, opened 2024-12-19
Fix: add no wait in endpoint to skip graceful timeout

jsun-m:


sdk/src/beta9/runner/endpoint.py
```diff
@@ -277,4 +277,7 @@ def shutdown(self, signum=None, frame=None):
         "timeout": cfg.timeout,
     }
 
+    if os.environ.get("no_wait") == "true":
```
luke-lombardi, line 280: can you name this something more descriptive e.g. GUNICORN_NO_WAIT

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #808, opened 2024-12-21
Fix: fix memory leak in endpoint buffer

luke-lombardi:
NOTE: Still need to reproduce this locally before merging


dleviminzi (approved):

pkg/abstractions/endpoint/task.go
```diff
@@ -30,7 +32,7 @@ func (t *EndpointTask) Execute(ctx context.Context, options ...interface{}) erro
 		return err
 	}
 
-	return instance.buffer.ForwardRequest(echoCtx, t)
+	return instance.buffer.ForwardRequest(echoCtx, t, payload)
```
luke-lombardi, line 35: This is a little hacky, but it makes sure that endpoint request payloads don't get stored in redis along with the rest of the task message. Since endpoint tasks don't support retries, this isn't necessary and wastes redis memory.

pkg/abstractions/endpoint/buffer.go
```diff
@@ -176,6 +173,11 @@ func (rb *RequestBuffer) processRequests() {
 				continue
 			}
 
+			if req.ctx.Request().Context().Err() != nil {
```
luke-lombardi, line 176: If the client hung up by the time we're handling this request (error is not nil even if the context is just cancelled), then we should just mark this task as canceled and move on to the next item in the request buffer.

PR #809, opened 2024-12-23
feat: add multipart uploads and range downloads for volumes

nickpetrovic:
- Add multipart volume uploads
- Add http range volume downloads
- Add fly.io juicefs gateway deployment
- Sort common commands by where they are defined
- Sort management commands alphabetically

Resolve BE-2133


sdk/src/beta9/cli/main.py
```diff
@@ -25,6 +25,8 @@
 
 click.formatting.FORCED_WIDTH = shutil.get_terminal_size().columns
 
+os.environ["GRPC_VERBOSITY"] = os.getenv("GRPC_VERBOSITY") or "ERROR"
```
nickpetrovic, line 28: This should stop prevent the gRPC library logs from being emitted between builds and when thread/process pools are being used.

sdk/src/beta9/channel.py
```diff
@@ -18,27 +18,58 @@
...
+        options.append(("grpc.secondary_user_agent", get_user_agent()))
+
         if credentials is not None:
             channel = grpc.secure_channel(addr, credentials)
         elif addr.endswith("443"):
             channel = grpc.secure_channel(addr, grpc.ssl_channel_credentials())
         else:
             channel = grpc.insecure_channel(addr)
 
+        channel.subscribe(*retry)
```
nickpetrovic, line 71: Setting this to True (the default) lets the channel reconnect on disconnects. It works really well from my testing.

sdk/src/beta9/channel.py
```diff
@@ -18,27 +18,58 @@
...
+        retry: Tuple[Callable[[grpc.ChannelConnectivity], None], bool] = (
+            channel_reconnect_event,
+            True,
+        ),
     ):
+        if options is None:
+            options = []
+
+        options = [opt for opt in options if "grpc.secondary_user_agent" not in opt[0]]
+        options.append(("grpc.secondary_user_agent", get_user_agent()))
```
nickpetrovic, line 62: The primary header is the grpclib and its version, the secondary is our CLI and its version.

pkg/common/config.default.yaml
```diff
@@ -56,6 +56,11 @@ gateway:
     memory: 32768
     maxReplicas: 10
     maxGpuCount: 2
+fileService:
+  endpointUrl: https://data-stage-beam-cloud.fly.dev
+  bucketName: beta9-fs
+  accessKey:
+  secretKey:
```
nickpetrovic, line 63: TODO: Add region as an option so its not hardcoded to an empty string.

pkg/abstractions/volume/multipart.go
```diff
@@ -0,0 +1,269 @@
...
+func (s *GlobalVolumeService) getS3Client() *s3.Client {
+	return s3.New(s3.Options{
+		Credentials: credentials.NewStaticCredentialsProvider(
+			s.config.AccessKey,
+			s.config.SecretKey,
+			"",
+		),
+		BaseEndpoint: aws.String(s.config.EndpointURL),
+		UsePathStyle: true,
+		Region:       "-",
```
nickpetrovic, line 34: TODO: make region configurable

deploy/fly-io/fly.toml
```diff
@@ -0,0 +1,47 @@
...
+    type = 'http'
+    interval = '15s'
+    timeout = '5s'
+    grace_period = '30s'
+    method = 'get'
+    path = '/minio/health/live'
+
+[[vm]]
+  cpus = 2
+  cpu_kind = 'shared'
```
nickpetrovic, line 46: Did not see a difference between "performance" and "shared" in my testing. Could be different in production though.

deploy/fly-io/fly.toml
nickpetrovic, line 1: In general, will probably want there to be a production version of this config so it has a rolling deployment strategy and different vm requirements.

deploy/fly-io/fly.toml
nickpetrovic: An observation - localstack struggles hard when downloading files from the juicefs s3 gateway.

sdk/src/beta9/cli/volume.py
```diff
@@ -189,6 +147,118 @@ def read_with_progress(
...
+    required=True,
+)
+@click.option(
+    "--version",
+    "-v",
+    is_eager=True,
+    type=click.STRING,
+    help="Version of the copy API.",
+    default="v2",
+)
```
nickpetrovic, line 242: Need to fallback to v1 if a file service is not configured.

deploy/fly-io/fly.toml
```diff
@@ -0,0 +1,47 @@
+app = 'data-stage-beam-cloud'
```
luke-lombardi, line 1: wondering if this should be passed in as an arg or something? since this is our real staging deployment, not sure

deploy/fly-io/start.sh
```diff
@@ -0,0 +1,37 @@
...
+
+while [ -z "$(tailscale ip 2>/dev/null)" ]; do
+  echo "Waiting for Tailscale to fully establish a connection..."
+  sleep 1
+done
+
+# juicefs
+JUICEFS_ADDRESS=${JUICEFS_ADDRESS:-0.0.0.0:9000}
+JUICEFS_BLOCK_SIZE=${JUICEFS_ADDRESS:-4096}
+JUICEFS_BUFFER_SIZE=${JUICEFS_BUFFER_SIZE:-300}
```
luke-lombardi, line 15: where are these configs pulled from?

pkg/abstractions/volume/multipart.go
```diff
@@ -0,0 +1,274 @@
...
+		Expires: aws.Time(time.Now().Add(time.Minute)),
+	})
+	if err != nil {
+		return &pb.CreateMultipartUploadResponse{
+			Ok:     false,
+			ErrMsg: err.Error(),
+		}, nil
+	}
+
+	total_parts := math.Ceil(float64(in.FileSize) / float64(in.ChunkSize))
```
luke-lombardi, line 161: should fix the casing here and below `totalParts`, `uploadParts`

luke-lombardi (changes requested):
This is looking great! Left a couple minor nitpicks.

pkg/abstractions/volume/multipart.go
nickpetrovic: Updated!

deploy/fly-io/start.sh
nickpetrovic: Good call out. They're from fly's secrets. Let me add a README to this directory. There's a few steps that need to take place before you can call `fly deploy`.

deploy/fly-io/fly.toml
nickpetrovic: Good idea. That's probably possible, along with some of the VM configs. I'll look into this.

nickpetrovic: Alright I modified the name to be "my-deployment". The reason is because fly will write the name of your deployment to this file even if you provide it as an argument to the CLI. I mentioned this in the README.

deploy/fly-io/start.sh
nickpetrovic: Added a README. Let me know what you think.

sdk/src/beta9/cli/volume.py
```diff
@@ -189,6 +148,144 @@ def read_with_progress(
     yield b""
 
 
+@common.command(
+    help="Copy contents to and from a volume.",
```
mernit, line 152: maybe update copy to include download here i.e. copy and download contents

nickpetrovic: Updated to this:

> Upload or download contents to or from a volume.

What do you think?

deploy/fly-io/fly.toml
luke-lombardi: LGTM

luke-lombardi (approved):

luke-lombardi (approved):

PR #812, opened 2024-12-26
Feat: Add warm up endpoint to both endpoints and taskqueues

jsun-m:
Added a new variable `NoOp` to the `TaskMessage` data model which only exists in Redis. If this variable is detected to be true in a task, the idea is that the task handler will detect it and run it as a fake task just so it can complete and start up the timer for the keep warm.

### DIFF
--- a/pkg/abstractions/endpoint/buffer.go
+++ b/pkg/abstractions/endpoint/buffer.go
@@ -434,6 +434,9 @@ func (rb *RequestBuffer) handleHttpRequest(req *request, c container) {
 	}
 
 	containerUrl := fmt.Sprintf("http://%s/%s", c.address, req.ctx.Param("subPath"))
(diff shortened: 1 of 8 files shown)


pkg/abstractions/endpoint/buffer.go
```diff
@@ -434,6 +434,9 @@ func (rb *RequestBuffer) handleHttpRequest(req *request, c container) {
 	}
 
 	containerUrl := fmt.Sprintf("http://%s/%s", c.address, req.ctx.Param("subPath"))
+	if req.task.msg.NoOp {
```
luke-lombardi, line 437: Feels like there a simpler way to do this that doesn't "trick" the system into warming up the task queue. Why not just have the warm up endpoint launch a container and set a keep warm on it so the autoscaler can't shut it down?

PR #813, opened 2024-12-26
Feat: Add warm up endpoint to both endpoints and taskqueues

jsun-m:
Add endpoint to warmup a deployment


pkg/abstractions/endpoint/http.go
```diff
@@ -17,17 +15,22 @@ type endpointGroup struct {
 func registerEndpointRoutes(g *echo.Group, es *HttpEndpointService) *endpointGroup {
 	group := &endpointGroup{routeGroup: g, es: es}
 
-	g.POST("/id/:stubId", auth.WithAuth(group.endpointRequest))
-	g.POST("/:deploymentName", auth.WithAuth(group.endpointRequest))
-	g.POST("/:deploymentName/latest", auth.WithAuth(group.endpointRequest))
-	g.POST("/:deploymentName/v:version", auth.WithAuth(group.endpointRequest))
-	g.POST("/public/:stubId", auth.WithAssumedStubAuth(group.endpointRequest, group.es.isPublic))
+	g.POST("/id/:stubId", auth.WithAuth(group.EndpointRequest))
```
jsun-m, line 18: uppercased these handlers to be consistent with other http handlers

pkg/abstractions/common/deployment.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"strconv"
+
+	apiv1 "github.com/beam-cloud/beta9/pkg/api/v1"
+	"github.com/beam-cloud/beta9/pkg/auth"
+	"github.com/beam-cloud/beta9/pkg/repository"
+	"github.com/beam-cloud/beta9/pkg/types"
+)
+
+func ParseAndValidateDeploymentStubId(
```
luke-lombardi, line 13: @jsun-m once we add a test for this, I think it would be fine to add. Also not sure if this should live here or if it should live under http/apiv1? Since you're passing in the backendRepo and stuff anyway.

luke-lombardi (changes requested):

pkg/abstractions/common/deployment.go
jsun-m: I'm not sure where it would live in apiv1. The deployments.go file is completely diff

luke-lombardi (approved):

PR #815, opened 2024-12-26
Chore: Prune more ContainerRequest data with new SanitizedContainerRequest type

jsun-m:


luke-lombardi (approved):

luke-lombardi (changes requested):
@jsun-m I think we talked about doing this with struct tags instead of adding an additional struct? Should we close this PR - or just do that in this PR?

PR #816, opened 2024-12-27
Fix: endpoint throughput improvements

luke-lombardi:
- Move endpoint task serialization to the point right before we actually run the task
- Migrate json marshal/unmarshal to faster library


jsun-m (approved):

pkg/abstractions/endpoint/endpoint.go
```diff
@@ -2,12 +2,13 @@ package endpoint
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"time"
 
+	jsoniter "github.com/json-iterator/go"
```
jsun-m, line 10: Why can't we name this `json` instead of `jsoniter` instead? Looks like it worked fine over here https://github.com/beam-cloud/beta9/pull/816/files#diff-3094a48851e225b16303320d7a7a55a493ecd14c499dec5418585eff10d1debbR17

luke-lombardi: See comment in slack

PR #817, opened 2024-12-31
Feat: beam shell

luke-lombardi:
Adds a shell command to let users connect to containers with their stub config. It can be used with any decorator that inherits from the `DeployableMixin`.

Screenshot 2025-01-01 at 6 52 13 PM


pkg/worker/base_runc_config.json
```diff
@@ -24,49 +24,65 @@
...
 				"CAP_CHOWN",
 				"CAP_DAC_OVERRIDE",
 				"CAP_FSETID",
 				"CAP_FOWNER",
 				"CAP_SETGID",
 				"CAP_SETUID",
-				"CAP_SETFCAP"
+				"CAP_SETFCAP",
+				"CAP_MKNOD",
+				"CAP_SYS_CHROOT"
```
luke-lombardi, line 37: I think the only one of these I actually need is `CAP_SYS_CHROOT`. Will test and revert the other changes if so.

sdk/src/beta9/abstractions/mixins.py
```diff
@@ -60,3 +65,81 @@ def deploy(
...
+            transport.auth_password("runc", ssh_token)
+            session = transport.open_session()
+
+            with SSHShell(
+                channel=session,
+            ) as _:
+                pass
+
+        except paramiko.SSHException as e:
+            print(f"SSH error: {e}")
```
luke-lombardi, line 136: TODO: convert to better errors and add automatic reconnect logic

pkg/abstractions/shell/shell.go
```diff
@@ -0,0 +1,253 @@
...
+		ContainerId: containerId,
+		Token:       token,
+	}, nil
+}
+
+func (ss *SSHShellService) genContainerId(stubId string) string {
+	return fmt.Sprintf("%s-%s-%s", shellContainerPrefix, stubId, uuid.New().String()[:8])
+}
+
+func (ss *SSHShellService) waitForContainerRunning(ctx context.Context, containerId string, delay time.Duration) error {
```
luke-lombardi, line 222: TODO: clean this up and make timeout configurable via param

pkg/abstractions/shell/shell.go
```diff
@@ -0,0 +1,253 @@
...
+	RouteGroup    *echo.Group
+	Tailscale     *network.Tailscale
+	EventRepo     repository.EventRepository
+}
+
+func NewSSHShellService(
+	ctx context.Context,
+	opts ShellServiceOpts,
+) (ShellService, error) {
+	keyEventManager, err := common.NewKeyEventManager(opts.RedisClient)
```
luke-lombardi, line 67: TODO: use key event manage to detect existing (running) shell containers on start

pkg/abstractions/shell/http.go
```diff
@@ -0,0 +1,103 @@
...
+	// TODO: confirm disconnects happen when python client exits
+	defer log.Println("disconnected")
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	// Copy from client -> container
+	go func() {
+		defer wg.Done()
+		buf := make([]byte, 32*1024) // 32KB buffer
```
luke-lombardi, line 90: TODO: move buffer size to const

pkg/abstractions/shell/http.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// Dial ssh server in the container
+	containerConn, err := net.Dial("tcp", containerAddress)
+	if err != nil {
+		return ctx.String(http.StatusBadGateway, "Failed to connect to container")
+	}
+	defer containerConn.Close()
+
+	// TODO: confirm disconnects happen when python client exits
+	defer log.Println("disconnected")
```
luke-lombardi, line 82: TODO: remove log once disconnect behavior is validated

pkg/gateway/services/stub.go
```diff
@@ -255,13 +255,19 @@ func (gws *GatewayService) GetURL(ctx context.Context, in *pb.GetURLRequest) (*p
...
 
-	// Get URL for Serves
+	// Get URL for Serves or Shells
 	if stub.Type.IsServe() {
 		invokeUrl := common.BuildServeURL(gws.appConfig.GatewayService.HTTP.GetExternalURL(), in.UrlType, stub)
 		return &pb.GetURLResponse{
 			Ok:  true,
 			Url: invokeUrl,
 		}, nil
+	} else if stub.Type.Kind() == types.StubTypeShell {
```
dleviminzi, line 265: Should we have `IsShell` like `IsServe`?

pkg/abstractions/shell/http.go
```diff
@@ -0,0 +1,101 @@
...
+	done := make(chan struct{})
+	var once sync.Once
+
+	go g.ss.keepAlive(ctx.Request().Context(), containerId, done)
+
+	// Send a 200 OK before hijacking
+	ctx.Response().WriteHeader(http.StatusOK)
+	ctx.Response().Flush()
+
+	// Hijack the connection
```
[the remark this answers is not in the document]
luke-lombardi: Hm, not sure I agree. It's tied to the http request.

pkg/abstractions/shell/shell.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+func (ss *SSHShellService) CreateShell(ctx context.Context, in *pb.CreateShellRequest) (*pb.CreateShellResponse, error) {
+	authInfo, _ := auth.AuthInfoFromContext(ctx)
+
+	stub, err := ss.backendRepo.GetStubByExternalId(ctx, in.StubId)
+	if err != nil {
+		return &pb.CreateShellResponse{
+			Ok: false,
+		}, nil
```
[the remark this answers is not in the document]
luke-lombardi: Rather be explicit imo

pkg/gateway/services/stub.go
luke-lombardi: I don't think so, since there's only one stub type for it

pkg/abstractions/shell/shell.go
```diff
@@ -0,0 +1,359 @@
...
+	containerId := ss.genContainerId(stub.ExternalId)
+
+	// Don't allow negative and 0-valued compute requests
+	if stubConfig.Runtime.Cpu <= 0 {
+		stubConfig.Runtime.Cpu = defaultContainerCpu
+	}
+
+	if stubConfig.Runtime.Memory <= 0 {
+		stubConfig.Runtime.Memory = defaultContainerMemory
+	}
```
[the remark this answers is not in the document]
luke-lombardi: that makes sense. I copied this from one of the other places we spawn containers

pkg/abstractions/shell/shell.go
```diff
@@ -0,0 +1,359 @@
...
+
+	go ss.eventRepo.PushRunStubEvent(authInfo.Workspace.ExternalId, &stub.Stub)
+
+	var stubConfig types.StubConfigV1 = types.StubConfigV1{}
+	err = json.Unmarshal([]byte(stub.Config), &stubConfig)
+	if err != nil {
+		return &pb.CreateShellResponse{
+			Ok: false,
+		}, nil
+	}
```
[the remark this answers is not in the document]
luke-lombardi: yeah agreed. I can put up a separate PR for this and the comment below.

pkg/abstractions/shell/shell.go
```diff
@@ -0,0 +1,359 @@
...
+	shellRoutePrefix              string        = "/shell"
+	shellContainerPrefix          string        = "shell"
+	shellContainerTtlS            int           = 30 // 30 seconds
+	shellProxyBufferSizeKb        int           = 32 * 1024
+	defaultContainerCpu           int64         = 100
+	defaultContainerMemory        int64         = 128
+	containerDialTimeoutDurationS time.Duration = 300 * time.Second
+	containerWaitTimeoutDurationS time.Duration = 5 * time.Minute
+	containerWaitPollIntervalS    time.Duration = 1 * time.Second
+	containerKeepAliveIntervalS   time.Duration = 5 * time.Second
```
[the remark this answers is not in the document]
luke-lombardi: I'm not sure these are things we'll need to change much -- config is kind of busy enough as it is

pkg/abstractions/shell/http.go
```diff
@@ -0,0 +1,101 @@
...
+		buf := make([]byte, shellProxyBufferSizeKb)
+		_, _ = io.CopyBuffer(containerConn, conn, buf)
+		once.Do(func() { close(done) })
+	}()
+
+	go func() {
+		buf := make([]byte, shellProxyBufferSizeKb)
+		_, _ = io.CopyBuffer(conn, containerConn, buf)
+		once.Do(func() { close(done) })
+	}()
```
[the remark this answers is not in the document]
luke-lombardi: Maybe yeah. I think in this case if it exits, then there's not much we can do since we just shut the other connection. There's  a reconnect on the client side automatically if it was some unexpected error

pkg/abstractions/shell/http.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	defer containerConn.Close()
+
+	// Create a context that will be canceled when the client disconnects
+	clientCtx, clientCancel := context.WithCancel(ctx.Request().Context())
+	defer clientCancel()
+
+	go func() {
+		buf := make([]byte, shellProxyBufferSizeKb)
+		_, _ = io.CopyBuffer(containerConn, conn, buf)
```
jsun-m, line 83: Closing 1 of these goroutines won't always close the other immediately.
Say if the copyBuffer is trying to read from `containerConn` but its blocking because containerConn has not read any messages yet, and then at the same time user's `conn` closes, I don't think the copyBuffer command will know this until the `containerConn` unblocks

jsun-m: func forwardWSConn(src net.Conn, dst net.Conn) {
	defer func() {
		src.Close()
		dst.Close()
	}()

	_, err := io.Copy(src, dst)
	if err != nil {
		return
	}
}

I remember having to forcefully close the other connection just in case here

pkg/abstractions/shell/http.go
```diff
@@ -0,0 +1,101 @@
...
+	go func() {
+		buf := make([]byte, shellProxyBufferSizeKb)
+		_, _ = io.CopyBuffer(conn, containerConn, buf)
+		once.Do(func() { close(done) })
+	}()
+
+	// Wait for either connection to close
+	select {
+	case <-done:
+		clientCancel()
```
jsun-m, line 96: do we need this? We already defer the clientCancel

pkg/abstractions/shell/shell.go
```diff
@@ -0,0 +1,359 @@
...
+	return fmt.Sprintf("%s-%s-%s", shellContainerPrefix, stubId, uuid.New().String()[:8])
+}
+
+func (ss *SSHShellService) keepAlive(ctx context.Context, containerId string, done <-chan struct{}) {
+	ticker := time.NewTicker(containerKeepAliveIntervalS)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
```
nickpetrovic, line 304: Can we also have a case for the global/gateway context? This'll let the gateway shutdown during a deployment.

pkg/abstractions/shell/http.go
luke-lombardi: kk I can do that

pkg/abstractions/shell/http.go
luke-lombardi: removed

pkg/gateway/services/stub.go
luke-lombardi: updates this logic

pkg/abstractions/shell/http.go
```diff
@@ -0,0 +1,100 @@
...
+		once.Do(func() { close(done) })
+	}()
+
+	// Wait for either connection to close
+	select {
+	case <-done:
+		return nil
+	case <-clientCtx.Done():
+		return nil
+	}
```
nickpetrovic, line 104: Can we wait for the global/gateway context here too?

pkg/abstractions/shell/http.go
luke-lombardi: updated

pkg/abstractions/shell/shell.go
```diff
@@ -0,0 +1,359 @@
...
+
+func (ss *SSHShellService) waitForContainer(ctx context.Context, containerId string, timeout time.Duration) error {
+	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	for {
+		select {
+		case <-timeoutCtx.Done():
+			return fmt.Errorf("timed out waiting for container to be available")
+		default:
```
nickpetrovic, line 322: Should add a case for the global context too.

pkg/abstractions/shell/shell.go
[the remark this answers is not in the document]
jsun-m: I second this

nickpetrovic (approved):
Left a few comments about the global contexts.

pkg/abstractions/shell/shell.go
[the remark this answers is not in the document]
jsun-m: I think a step up from this is to also have a function to configure `mounts, secrets, cpu, memory, and gpu stuff`

pkg/abstractions/shell/shell.go
luke-lombardi: done

pkg/abstractions/shell/http.go
luke-lombardi: done

pkg/abstractions/shell/shell.go
luke-lombardi: done

pkg/abstractions/shell/shell.go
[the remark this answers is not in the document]
luke-lombardi: I agree, I just think it should be a separate PR

PR #818, opened 2024-12-31
Fix: Use correct ignore file for beta9 and beam cli commands

dleviminzi:
Previously, commands would always get the default sdk settings. This was caused because importing `sync.py` caused the default settings to be loaded no matter what. Now, we don't set `_settings` in `sync.py` and instead access it indirectly so that importing sync doesn't initialize settings with defaults. When "beam" is included in the modules, we load the beam config instead of the default beta9 config. 

Old behavior:
```bash
~/Dev/beam/bs/endpoint ❯ beam deploy app.py:handler                                        
=> Building image 
=> Using cached image 
=> Syncing files 
Reading .beta9ignore file  # WRONG
```

New behavior: 
```bash
~/Dev/beam/bs/endpoint ❯ beam deploy app.py:handler                                     
=> Building image 
=> Using cached image 
=> Syncing files 
Reading .beamignore file  # CORRECT
```


sdk/src/beta9/config.py
```diff
@@ -73,10 +76,26 @@ def is_valid(self) -> bool:
...
+            s = Settings(
+                name="Beam",
+                api_host=os.getenv("API_HOST", "https://app.beam.cloud"),
+                gateway_host=os.getenv("GATEWAY_HOST", "gateway.beam.cloud"),
+                gateway_port=int(os.getenv("GATEWAY_PORT", 443)),
+                config_path=Path("~/.beam/config.ini").expanduser(),
+                use_defaults_in_prompt=True,
+            )
+
     _SETTINGS = s
```
nickpetrovic, line 99: When is _SETTINGS set if this is remote? Does it work properly?

dleviminzi: The logic here is the same as it was when remote. If called without an explicitly passed in settings it will just load the default. I didn't want to change the remote behavior, which is why I included the local check.

dleviminzi: I found that the import of `sync.py` was causing the default init of the settings and changed that instead of doing the module checking.

nickpetrovic (approved):

jsun-m (approved):

PR #820, opened 2025-01-02
Fix: Return error and exception when endpoint app raises an exception

dleviminzi:
Resolve BE-2148

Once the warning message has been circulated and users have had time to update any handling logic we can merge this. 

It is the exact same thing as https://github.com/beam-cloud/beta9/pull/791, but we need to wait to release

### DIFF
--- a/sdk/src/beta9/runner/endpoint.py
+++ b/sdk/src/beta9/runner/endpoint.py
@@ -5,7 +5,7 @@ import signal
 import traceback
 from contextlib import asynccontextmanager
 from http import HTTPStatus
-from typing import Any, Dict, Optional, Tuple, Union
+from typing import Any, Dict, Optional, Union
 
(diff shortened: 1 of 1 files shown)


luke-lombardi (approved):
LGTM - cc @mernit to communicate the change in behavior to users. Basically now if your endpoint throws an error you'll get a 500 instead of a 200 as you do now.

luke-lombardi:
> LGTM - cc @mernit to communicate the change in behavior to users. Basically now if your endpoint throws an error you'll get a 500 instead of a 200 as you do now.

@mernit  For more context, @dleviminzi's original message:

`Previously, if the handler raised an exception it would be returned with status 200 and the task would be marked complete. With this change, the exception is returned with status 500 and the task is marked as errored.`

mernit:
@luke-lombardi we already communicated this in Slack, I told them we'd release it Thursday

PR #830, opened 2025-01-07
Fix: clear worker address when deleting container state

luke-lombardi:


pkg/repository/base.go
```diff
@@ -50,8 +50,8 @@ type ContainerRepository interface {
 	UpdateAssignedContainerGPU(string, string) error
 	DeleteContainerState(containerId string) error
 	SetWorkerAddress(containerId string, addr string) error
-	SetContainerStateWithConcurrencyLimit(quota *types.ConcurrencyLimit, request *types.ContainerRequest) error
 	GetWorkerAddress(ctx context.Context, containerId string) (string, error)
+	SetContainerStateWithConcurrencyLimit(quota *types.ConcurrencyLimit, request *types.ContainerRequest) error
```
luke-lombardi, line 54: This just moved because it made more sense for the worker address related funcs to be next to each other

nickpetrovic (approved):

PR #832, opened 2025-01-08
Increase tmp size limit

dleviminzi:
Now that we copy image data into this directory, we need to increase its size for large images.

### DIFF
--- a/pkg/scheduler/pool_external.go
+++ b/pkg/scheduler/pool_external.go
@@ -450,9 +450,18 @@ func (wpc *ExternalWorkerPoolController) getWorkerVolumes(workerMemory int64) []
 	hostPathType := corev1.HostPathDirectoryOrCreate
 	sharedMemoryLimit := calculateMemoryQuantity(wpc.workerPool.PoolSizing.SharedMemoryLimitPct, workerMemory)
 
-	tmpSizeLimit := resource.MustParse(wpc.config.Worker.TmpSizeLimit)
-	if wpc.workerPool.TmpSizeLimit != "" {
-		tmpSizeLimit = resource.MustParse(wpc.workerPool.TmpSizeLimit)
+	defaultLimit := resource.MustParse("30Gi")
+
+	// First try worker pool specific limit, then fall back to global config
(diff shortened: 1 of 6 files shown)


luke-lombardi (approved):

pkg/scheduler/pool_external.go
```diff
@@ -450,7 +450,11 @@ func (wpc *ExternalWorkerPoolController) getWorkerVolumes(workerMemory int64) []
 	hostPathType := corev1.HostPathDirectoryOrCreate
 	sharedMemoryLimit := calculateMemoryQuantity(wpc.workerPool.PoolSizing.SharedMemoryLimitPct, workerMemory)
 
-	tmpSizeLimit := resource.MustParse("30Gi")
+	tmpSizeLimit := resource.MustParse(wpc.config.Worker.TmpSizeLimit)
+	if wpc.workerPool.TmpSizeLimit != "" {
```
luke-lombardi, line 454: can you validate the size as well here after parsing it? i could also imagine this panicking if the user specified some nonsense value.

luke-lombardi (changes requested):

pkg/scheduler/pool_local.go
```diff
@@ -252,7 +252,20 @@ func (wpc *LocalKubernetesWorkerPoolController) getWorkerVolumes(workerMemory in
 	hostPathType := corev1.HostPathDirectoryOrCreate
 	sharedMemoryLimit := calculateMemoryQuantity(wpc.workerPool.PoolSizing.SharedMemoryLimitPct, workerMemory)
 
-	tmpSizeLimit := resource.MustParse("30Gi")
+	defaultLimit := resource.MustParse("30Gi")
```
luke-lombardi, line 255: should this be a helper function since the logic is the same in both places? (pool external / pool local)

luke-lombardi (changes requested):

pkg/scheduler/pool_local.go
dleviminzi: yeah, bumped it into utils

luke-lombardi (approved):

PR #833, opened 2025-01-08
Fix: Patch issue with image build process not spinning down and undo revert for killing expired build containers

jsun-m:


luke-lombardi:
@jsun-m converting this to draft until it's ready

luke-lombardi (approved):

PR #836, opened 2025-01-09
Fix: Stuck build containers

dleviminzi:
Resolve BE-2203

This PR contains to work to fix one of the likely causes of stuck build workers. Basically, if a user cancels a build (Ctrl-C) before the build container has started, the worker will get stuck. As things are, the entire build will carry on as if nothing happened and then the container will just sit there forever. 

To prevent this from happening we can use redis pub/sub. When the context is cancelled in the gateway, we publish a stop-build event to a channel containing the container id (`stop-build-{container_id}`). Meanwhile, when the worker claims the container request, it subscribes to that same channel. If an event comes through it cancels the context that has been passed down.


pkg/scheduler/scheduler.go
```diff
@@ -164,6 +164,21 @@ func (s *Scheduler) Stop(stopArgs *types.StopContainerArgs) error {
 	return nil
 }
 
+func (s *Scheduler) StopBuild(containerId string) error {
```
luke-lombardi, line 167: We couldn't piggy back on the stop_container message? It feels wrong that this is a separate scheduler method

luke-lombardi (changes requested):

pkg/worker/worker.go
```diff
@@ -272,7 +272,21 @@ func (s *Worker) Run() error {
 			if !exists {
 				log.Info().Str("container_id", containerId).Msg("running container")
 
-				err := s.RunContainer(request)
+				ctx, cancel := context.WithCancel(context.Background())
```
luke-lombardi, line 275: should we be using the worker context here as the base context?

pkg/worker/worker.go
```diff
@@ -272,7 +272,21 @@ func (s *Worker) Run() error {
...
+				eventbus := common.NewEventBus(s.redisClient, common.EventBusSubscriber{Type: common.EventType("stop-build" + "-" + containerId), Callback: func(e *common.Event) bool {
+					log.Info().Str("container_id", containerId).Msg("received stop build event")
+					cancel()
+					return true
+				}})
+				go eventbus.ReceiveEvents(ctx)
+
+				s.containerInstances.Set(containerId, &ContainerInstance{
```
luke-lombardi, line 283: move back to where it was

luke-lombardi (changes requested):

pkg/worker/worker.go
dleviminzi: changed

pkg/abstractions/image/build.go
```diff
@@ -256,6 +258,14 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
 
 	containerId := b.genContainerId()
 
+	go func() {
+		<-ctx.Done() // If user cancels the build, send a stop-build event to the scheduler
```
jsun-m, line 262: at line 331 we do a runc client.Kill(containerId) for this exact same case. We should decide on one and remove the other

pkg/abstractions/image/build.go
```diff
@@ -327,36 +337,47 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
...
 
-		exitCode, err := b.containerRepo.GetContainerExitCode(containerId)
-		if err == nil && exitCode != 0 {
-			msg, ok := types.WorkerContainerExitCodes[exitCode]
-			if !ok {
-				msg = types.WorkerContainerExitCodes[types.WorkerContainerExitCodeUnknownError]
+		case <-time.After(100 * time.Millisecond):
+			r, err := client.Status(containerId)
+			if err != nil {
+				outputChan <- common.OutputMsg{Done: true, Success: false, Msg: "Error occurred while checking container status: " + err.Error()}
```
jsun-m, line 351: Should there be some maximum checks before sending `done:  true, success: false` back to the client? In case some networking issue happens?

jsun-m (changes requested):

pkg/abstractions/image/build.go
dleviminzi: Yeah I think this makes sense, but I decided to leave it for now because this is how it currently is. I figured we could make that tweak in a follow up.

pkg/abstractions/image/build.go
jsun-m: i think then this could replace that other piece of code then. The stop container event through the scheduler should also kill the container.

dleviminzi: There will be no container running. The stop build request is sent to stop things like the skopeo copy and image unpacking. That is distinct from stopping a running container on the worker. 

It is possible to consolidate the two functions such that it sends the stop build request and the container kill. The reason I didn't opt for that is because it allows for a user to cancel the request before the runc client is connected. That isn't the biggest deal in the world, but I figured the faster we can cancel the request the better.

dleviminzi: At first I explored piggy backing off of the scheduler stop container argument. The issue with that is that the stop container request will not actually be handled by the worker until the container is actually running. At that point, we will have finished copying, unpacking, and archiving the base image. If the user cancelled in the middle of the copy, a lot of work might continue happening after they've requested a cancellation.

jsun-m (approved):

pkg/worker/worker.go
```diff
@@ -272,7 +272,15 @@ func (s *Worker) Run() error {
 			if !exists {
 				log.Info().Str("container_id", containerId).Msg("running container")
 
-				err := s.RunContainer(request)
+				ctx, cancel := context.WithCancel(s.ctx)
```
luke-lombardi, line 275: we shouldn't be subscribing to this event unless its a build container. is there a way we can refactor this out into a separate method that we only call if its a build container?

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #843, opened 2025-01-10
Add note warning about poetry shell being moved to plugin in v2.0.0

dleviminzi:


Makefile
```diff
@@ -9,6 +9,7 @@ setup:
 	# helm install beta9 deploy/charts/beta9 --create-namespace --values deploy/charts/beta9/values.local.yaml
 	kustomize build --enable-helm manifests/kustomize/overlays/cluster-dev | kubectl apply -f-
 
+# Note: you may need https://github.com/python-poetry/poetry-plugin-shell as of poetry v2.0.0
```
nickpetrovic, line 12: What if we try doing something like `poetry self add poetry-plugin-shell || true` in this target? Assume its there?

dleviminzi: We could do that, but I'm not sure what will happen if the user has an old version still. Also, depending on how you installed poetry you might need to run something different.

nickpetrovic (approved):

PR #845, opened 2025-01-10
Feat: Build with gpu option for images

dleviminzi:
Resolve BE-2199

This PR adds a new flag to image: `build_with_gpu(type)`. If a GPU is specified, the container request for the build will go to a GPU worker. 

Note: this should not be released until the stuck build container issue is resolved.


sdk/src/beta9/abstractions/image.py
```diff
@@ -242,9 +246,18 @@ def handler():
                 pass
             ```
 
-            The context directory should contain all files referenced in your Dockerfile
-            (like files being COPYed). If no context_dir is specified, the directory
-            containing the Dockerfile will be used as the context.
+            Building on a GPU Node
```
luke-lombardi, line 249: update copy to say "building in a gpu container" or "building in a gpu environment" maybe

luke-lombardi (changes requested):

pkg/abstractions/image/build.go
```diff
@@ -429,6 +385,70 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
 	return nil
 }
 
+// generateContainerRequest generates a container request for the build container
```
luke-lombardi, line 388: Hard to see exactly what changed in here - is it mostly just the refactor to separate method?

dleviminzi: Yeah the only real change here is the addition of [these lines](https://github.com/beam-cloud/beta9/pull/845/files#diff-fff33a75503ea6a83c56e06f80aefb54535dcc3502ac57f84ff518c59984f1c7R442-R447). Besides that, I just bumped this out into another function because it was sprawling.

luke-lombardi (approved):

PR #846, opened 2025-01-10
Feat: Add min_containers to QueueDepthAutoscaler and instance controller for loading and reloading instances

jsun-m:
Add min_containers field that allows users to set a the minimum amount of containers to keep warm when there is no activity. 

Follow up:
- [ ] Release worker
- [ ] Release sdk
- [ ] Release gateway


pkg/gateway/services/stub.go
```diff
@@ -226,6 +227,22 @@ func (gws *GatewayService) DeployStub(ctx context.Context, in *pb.DeployStubRequ
 
 	go gws.eventRepo.PushDeployStubEvent(authInfo.Workspace.ExternalId, &stub.Stub)
 
+	var config types.StubConfigV1
+	if err := json.Unmarshal([]byte(stub.Config), &config); err != nil {
```
luke-lombardi, line 231: Confused about why this was added here. Also why are triggering an event reload here, that also seems wrong

pkg/repository/backend_postgres.go
```diff
@@ -1234,6 +1234,17 @@ func (c *PostgresBackendRepository) ListStubsPaginated(ctx context.Context, filt
 	return *page, nil
 }
 
+func (r *PostgresBackendRepository) ListKeepWarmDeploymentStubs(ctx context.Context, stubType []string) ([]types.Stub, error) {
```
luke-lombardi, line 1237: Why do we need this?

luke-lombardi (changes requested):

pkg/gateway/services/stub.go
jsun-m: The autoscaler needs to be running in order to make sure we have min containers accounted for. In order to get the autoscaler running the instance of the deployment need to be initialized.

pkg/repository/backend_postgres.go
luke-lombardi: i don't think this should be here, it can be pulled in the autoscaler

pkg/gateway/services/stub.go
luke-lombardi: i don't think so

pkg/abstractions/common/instance.go
```diff
@@ -175,6 +175,9 @@ func (i *AutoscaledInstance) WaitForContainer(ctx context.Context, duration time
 }
 
 func (i *AutoscaledInstance) ConsumeScaleResult(result *AutoscalerResult) {
+	if i.StubConfig.Autoscaler.MinContainers > uint(result.DesiredContainers) {
```
luke-lombardi, line 178: This makes sense, but should we use a max func here instead? I.e. whichever is larger

desired containers = 0, min containers = 1 --> 1
desired containers = 2, min containers = 1 --> 2
desired containers = 0 , min containers = 0 --> 0

luke-lombardi (changes requested):

pkg/api/v1/deployment.go
```diff
@@ -150,16 +150,16 @@ func (g *DeploymentGroup) DeleteDeployment(ctx echo.Context) error {
...
-		return HTTPInternalServerError("Failed to stop deployment")
-	}
-
 	// Delete deployment
 	if err := g.backendRepo.DeleteDeployment(ctx.Request().Context(), deploymentWithRelated.Deployment); err != nil {
 		return HTTPInternalServerError("Failed to delete deployment")
 	}
 
+	// Stop deployment
+	if err := g.stopDeployments([]types.DeploymentWithRelated{*deploymentWithRelated}, ctx); err != nil {
```
luke-lombardi, line 159: Why did this move down?

pkg/abstractions/common/instance.go
```diff
@@ -321,3 +324,102 @@ func (i *AutoscaledInstance) emitUnhealthyEvent(stubId, currentState, reason str
...
+		_, err := ic.getOrCreateInstance(ic.ctx, stub.Stub.ExternalId)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (ic *InstanceController) reloadInstance(stubId, stubType string) error {
```
luke-lombardi, line 411: let's just call this reload

pkg/abstractions/common/instance.go
```diff
@@ -321,3 +324,102 @@ func (i *AutoscaledInstance) emitUnhealthyEvent(stubId, currentState, reason str
...
+		types.DeploymentFilter{
+			StubType:         ic.StubTypes,
+			MinContainersGTE: 1,
+			Active:           ptr.To(true),
+		},
+	)
+	if err != nil {
+		return err
+	}
+	log.Printf("Found %d stubs to load", len(stubs))
```
luke-lombardi, line 399: remove log

pkg/abstractions/common/instance.go
```diff
@@ -321,3 +324,102 @@ func (i *AutoscaledInstance) emitUnhealthyEvent(stubId, currentState, reason str
 	log.Info().Str("instance_name", i.Name).Msgf("%s\n", reason)
 	go i.EventRepo.PushStubStateUnhealthy(i.Workspace.ExternalId, stubId, currentState, state, reason, containers)
 }
+
+type InstanceController struct {
```
luke-lombardi, line 328: I think I like this idea, but should it be called AutoscaledInstanceController? Also, where exactly is the separate between autoscaled instance... isn't that also a "controller" in a sense? Should it all be the same struct?

luke-lombardi: or do you mean instance controller controls instances separately? because this get we have getorcreateinstnace so its sort of like an instance manager?

luke-lombardi (changes requested):

pkg/api/v1/deployment.go
jsun-m: stop deployments includes the eventBus event that reloads the instances in the gateway. The problem is that we need to set the active field to false before sending that event else the instance reloads the data thinking that the deployment is still active.

pkg/abstractions/common/instance.go
jsun-m: The naming can be updated but the concept of this struct is that it has some functions that manage all the instances of a resource (and those functions are same amongst resources with instances)

jsun-m:
For the refactored portion of the controllers. 

I added a new controller called `InstanceController` that handles loading and reloading of the AutoscaledInstances. Here's the heirarchy
image

pkg/abstractions/common/instance.go
```diff
@@ -321,3 +324,101 @@ func (i *AutoscaledInstance) emitUnhealthyEvent(stubId, currentState, reason str
...
+			if err := ic.reload(stubId, stubType); err != nil {
+				return false
+			}
+
+			return true
+		}},
+	)
+	go eventBus.ReceiveEvents(ic.ctx)
+
+	if err := ic.loadStubInstances(); err != nil {
```
luke-lombardi, line 368: should just say "load" maybe?

pkg/abstractions/common/instance.go
```diff
@@ -321,3 +324,101 @@ func (i *AutoscaledInstance) emitUnhealthyEvent(stubId, currentState, reason str
...
+) error {
+	instance, err := ic.getOrCreateInstance(ctx, stubId)
+	if err != nil {
+		return err
+	}
+
+	return instance.HandleScalingEvent(1)
+}
+
+func (ic *InstanceController) loadStubInstances() error {
```
luke-lombardi, line 387: hm I guess I see why you called it that. trying to think of a better name here...

pkg/abstractions/common/instance.go
```diff
@@ -321,3 +324,101 @@ func (i *AutoscaledInstance) emitUnhealthyEvent(stubId, currentState, reason str
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (ic *InstanceController) reload(stubId, stubType string) error {
+	if stubType != types.StubTypeEndpointDeployment && stubType != types.StubTypeASGIDeployment {
```
luke-lombardi, line 411: can you remind me why we have this check? isn't this used for taskqueues as well?

pkg/abstractions/endpoint/endpoint.go
```diff
@@ -32,19 +32,20 @@ type EndpointService interface {
...
+	scheduler          *scheduler.Scheduler
+	backendRepo        repository.BackendRepository
+	workspaceRepo      repository.WorkspaceRepository
+	containerRepo      repository.ContainerRepository
+	eventRepo          repository.EventRepository
+	taskRepo           repository.TaskRepository
+	endpointInstances  *common.SafeMap[*endpointInstance]
+	tailscale          *network.Tailscale
+	taskDispatcher     *task.Dispatcher
+	instanceController *abstractions.InstanceController
```
luke-lombardi, line 48: can we just call it controller?

pkg/abstractions/endpoint/endpoint.go
```diff
@@ -102,34 +103,17 @@ func NewHTTPEndpointService(
...
-				return false
-			}
-
-			instance.Reload()
-
-			return true
-		}},
-	)
-	go eventBus.ReceiveEvents(ctx)
+	es.instanceController = abstractions.NewInstanceController(ctx, es.InstanceFactory, []string{types.StubTypeEndpointDeployment, types.StubTypeASGIDeployment}, es.backendRepo, es.rdb)
```
luke-lombardi, line 112: same comment as above

pkg/abstractions/taskqueue/taskqueue.go
```diff
@@ -67,11 +67,12 @@ type RedisTaskQueue struct {
...
-	keyEventManager *common.KeyEventManager
-	queueClient     *taskQueueClient
-	tailscale       *network.Tailscale
-	eventRepo       repository.EventRepository
+	queueInstances     *common.SafeMap[*taskQueueInstance]
+	keyEventManager    *common.KeyEventManager
+	queueClient        *taskQueueClient
+	tailscale          *network.Tailscale
+	eventRepo          repository.EventRepository
+	instanceController *abstractions.InstanceController
```
luke-lombardi, line 75: same comment

luke-lombardi (changes requested):

pkg/abstractions/common/instance.go
jsun-m: this is wrong. I didnt remove this piece of code when consolidating the logic.

pkg/abstractions/taskqueue/taskqueue.go
```diff
@@ -103,34 +104,18 @@ func NewRedisTaskQueueService(
...
-	eventManager.Listen(ctx)
+	eventManager.Listen()
 
-	eventBus := common.NewEventBus(
-		opts.RedisClient,
-		common.EventBusSubscriber{Type: common.EventTypeReloadInstance, Callback: func(e *common.Event) bool {
-			stubId := e.Args["stub_id"].(string)
-			stubType := e.Args["stub_type"].(string)
-
-			if stubType != types.StubTypeTaskQueueDeployment {
```
luke-lombardi, line 118: @jsun-m the place where you're doing the `correctStub` check, you need to return true I think or else it will resend the event

luke-lombardi (changes requested):

PR #847, opened 2025-01-11
fix: multipart uploads and downloads with empty files

nickpetrovic:
- When file size is 0, make just 1 presigned url upload part
- When upload is complete, call progress callback at least once to indicate it completed
- Don't send any data to juicefs s3 gateway when chunk is empty
- Fix divide by zero errors with downloads

Resolve BE-2133


mernit (approved):
Nice

PR #856, opened 2025-01-14
Feat: Use cloudpickle to run remote functions in jupyter environments

dleviminzi:
Resolve BE-2218

This PR changes the strategy used for running remote functions in notebooks. Instead of creating a temporary file built by stitching together executed cells, we use cloudpickle to capture the required function + deps. This resolves issues with re-executed cells and is generally more reliable. 

This PR also changes the way that the default python environment is selected for `Image`. It tries to see if the active python version from the local environment is a valid remote python version. If it is, it will use that version by default. I can remove this change or limit it to ipython environments only. This is nice when working in jupyter notebooks because cloudpickle dumps are dependent on the python version. If you create a dump in 3.12 and try to load it in 3.10, things will not work. 

TODO: 
- [x] Test whether different protocols improve compatibility for cloudpickle python version mixing
- [x] Alert users when they try to run different local/remote python versions in jupyter environments
- [ ] Explore ways to detect function content changes to avoid re-syncing when the cell has been re-executed but no code has changed


sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -264,24 +288,34 @@ def _parse_cpu_to_millicores(self, cpu: Union[float, str]) -> int:
...
+                pickle_name = os.path.basename(tmp_file.name)
+                module_name = f"pickled_functions/{pickle_name}"
+                self.tmp_files.append(tmp_file)
+
+                setattr(self, attr, f"{module_name}:{func.__name__}")
+            except Exception as e:
+                os.unlink(tmp_file.name)
+                raise ValueError(f"Failed to pickle function: {str(e)}")
+        elif hasattr(module, "__file__"):
+            # Normal module case - use relative path
```
dleviminzi, line 312: I inverted this check because doing so allows us to add support for marimo notebooks. I know its odd to check for the special case before the normal case, but I think the slight awkwardness is worth it in this case.

pkg/gateway/services/stub.go
```diff
@@ -15,9 +15,11 @@ import (
 	"github.com/beam-cloud/beta9/pkg/common"
 	"github.com/beam-cloud/beta9/pkg/types"
 	pb "github.com/beam-cloud/beta9/proto"
+	"github.com/rs/zerolog/log"
 )
 
 func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCreateStubRequest) (*pb.GetOrCreateStubResponse, error) {
+	log.Info().Msg("GetOrCreateStub")
```
luke-lombardi, line 22: remove log

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -77,6 +75,32 @@ def _set_stub_created_for_workspace(value: bool) -> None:
     _stub_created_for_workspace = value
 
 
+class TempFile:
```
luke-lombardi, line 78: should this live here? feels like a util?

luke-lombardi: also kind of surprised there isn't a builtin that supports this

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -264,24 +288,38 @@ def _parse_cpu_to_millicores(self, cpu: Union[float, str]) -> int:
...
+        For Jupyter notebooks, serialize everything using cloudpickle.
         """
         if getattr(self, attr):
             return
 
-        module = inspect.getmodule(func)  # Determine module / function name
-        if hasattr(module, "__file__"):
+        module = inspect.getmodule(func)
+
+        # We check for the special cases before the normal cases because marimo notebooks
```
luke-lombardi, line 298: wondering if we should refactor out the notebook logic to a separate file under base? like base/ipython.py or something? maybe not. just an idea

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -264,24 +288,38 @@ def _parse_cpu_to_millicores(self, cpu: Union[float, str]) -> int:
...
+            tmp_file = TempFile(name=f"{func.__name__}.pkl", mode="wb")
+            try:
+                cloudpickle.dump(func, tmp_file)
+                tmp_file.flush()
+                pickle_name = os.path.basename(tmp_file.name)
+                module_name = f"pickled_functions/{pickle_name}"
+                self.tmp_files.append(tmp_file)
+
+                setattr(self, attr, f"{module_name}:{func.__name__}")
+            except Exception as e:
```
luke-lombardi, line 312: what exception do you expect to occur here?

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -264,24 +288,38 @@ def _parse_cpu_to_millicores(self, cpu: Union[float, str]) -> int:
...
+        # are normal python files, but don't work with the module based invocation logic.
+        if is_ipython_env():
+            import cloudpickle
+
+            tmp_file = TempFile(name=f"{func.__name__}.pkl", mode="wb")
+            try:
+                cloudpickle.dump(func, tmp_file)
+                tmp_file.flush()
+                pickle_name = os.path.basename(tmp_file.name)
+                module_name = f"pickled_functions/{pickle_name}"
```
luke-lombardi, line 308: is this path fixed? should be a const probably

sdk/src/beta9/abstractions/image.py
```diff
@@ -80,14 +81,27 @@ class NGCCredentials(TypedDict, total=False):
...
+
 class Image(BaseAbstraction):
     """
     Defines a custom container image that your code will run in.
     """
 
     def __init__(
         self,
-        python_version: PythonVersionAlias = PythonVersion.Python310,
+        python_version: PythonVersionAlias = detected_python_version(),
```
luke-lombardi, line 104: this makes sense but I'm wondering if this will break peoples existing pipelines. since they may be deploying from a python version that doesn't match their actual deployment environment. worth considering.

sdk/src/beta9/env.py
```diff
@@ -68,3 +69,16 @@ def try_env(env: str, default: EnvValue) -> Union[EnvValue, bool]:
...
+
+def is_ipython_env() -> bool:
+    if "google.colab" or "marimo" in sys.modules:
+        return True
+
+    try:
+        from IPython import get_ipython
+
+        shell = get_ipython().__class__.__name__
+        return shell == "ZMQInteractiveShell"
```
luke-lombardi, line 82: where's this come from? is this constant? can we pull it from a dependency?

sdk/src/beta9/runner/common.py
```diff
@@ -199,14 +217,21 @@ def _load(self):
                 module, func = config.handler.split(":")
 
             with self.importing_user_code():
-                target_module = importlib.import_module(module)
-
-            self.handler = getattr(target_module, func)
-            self.signature = inspect.signature(self.handler.func)
+                if "pickled_functions" in module:
```
luke-lombardi, line 220: this is messy, gotta come up with a cleaner way of doing this

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/base/runner.py
dleviminzi: The `cloudpickle.dump` call can raise a variety of exceptions. I don't think that our code is in danger of triggering any of them, but I figured it was safer to catch in case user code does something odd to the func object.

sdk/src/beta9/abstractions/image.py
dleviminzi: Yeah it is a tricky change. I could limit it to notebook environments for now. Might be safer

sdk/src/beta9/runner/common.py
dleviminzi: Yeah I think a better way would be to detect the file type. I'll make the change

sdk/src/beta9/runner/common.py
```diff
@@ -199,14 +217,21 @@ def _load(self):
...
+            # Check if handler is a wrapped function or direct function
+            target_func = getattr(self.handler, "func", self.handler)
+            self.signature = inspect.signature(target_func)
             self.pass_context = "context" in self.signature.parameters
-            self.is_async = asyncio.iscoroutinefunction(self.handler.func)
-        except BaseException:
-            raise RunnerException(f"Error loading handler: {traceback.format_exc()}")
+            self.is_async = asyncio.iscoroutinefunction(target_func)
+        except BaseException as e:
+            raise RunnerException(f"Error loading handler: {str(e)}\n{traceback.format_exc()}")
```
luke-lombardi, line 234: shouldn't traceback.format_exc show the correct exception? why do you have the str(e) here?

sdk/src/beta9/runner/common.py
```diff
@@ -199,14 +217,21 @@ def _load(self):
...
             with self.importing_user_code():
-                target_module = importlib.import_module(module)
-
-            self.handler = getattr(target_module, func)
-            self.signature = inspect.signature(self.handler.func)
+                if Path(module).suffix == PICKLE_SUFFIX:
+                    # Handle pickled function case
+                    self.handler = self._load_pickled_function(module)
+                else:
+                    # Handle normal module case
```
luke-lombardi, line 224: can we change the phrasing of these comments. Just `Handle standard modules (i.e. .py files)` etc

luke-lombardi: Also will need to update comments in runner/base.py

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #877, opened 2025-01-21
Feat: Limit number of lines a container can log per some interval

dleviminzi:
Resolve BE-2198

This gives us config options to stop spammy applications from overwhelming our logging infrastructure. 

Update: 

I changed this to use golang.org/x/time/rate. This way tokens are replenished.


pkg/common/config.default.yaml
```diff
@@ -177,6 +177,8 @@ worker:
   terminationGracePeriod: 30
   addWorkerTimeout: 10m
   blobCacheEnabled: false
+  containerLogLineLimit: 1000
```
luke-lombardi, line 180: This seems too low as a default, and the reset interval also seems too low? maybe 1000 for every 10 minutes? Curious what we were seeing in production

pkg/worker/logger.go
```diff
@@ -72,11 +76,37 @@ func (r *ContainerLogger) CaptureLogs(containerId string, logChan chan common.Lo
...
 		msgDecoded := false
 
+		currentCount := lineCount.Load()
+		if currentCount > int64(r.containerLogLineLimit) {
+			if currentCount == int64(r.containerLogLineLimit)+1 {
+				log.Info().Str("container_id", containerId).Msg("Reached log line limit of " + strconv.Itoa(r.containerLogLineLimit) + ", stopping log capture")
+				f.WithFields(logrus.Fields{
+					"container_id": containerId,
+					"stub_id":      instance.StubId,
+				}).Info("Reached log line limit, stopping log capture")
```
luke-lombardi, line 100: won't this log spam once you reach the limit?

pkg/worker/logger.go
```diff
@@ -72,11 +76,37 @@ func (r *ContainerLogger) CaptureLogs(containerId string, logChan chan common.Lo
...
+				instance.LogBuffer.Write([]byte(fmt.Sprintf("Reached log line limit of %d, stopping log capture until timer resets\n", r.containerLogLineLimit)))
```
luke-lombardi, line 101: same comment as above

luke-lombardi (changes requested):

pkg/common/config.default.yaml
dleviminzi: Yeah I figured we would override these in each environment, but I can bump defaults.

pkg/worker/logger.go
dleviminzi: This block is only entered on the condition: 
```
if currentCount == int64(r.containerLogLineLimit)+1
```

luke-lombardi: so currentCount won't continuously evaluate to, let's say, 1000 before we send those logs? The only place I see linecount being reset is on the timer, where is it decremented?

luke-lombardi (changes requested):

pkg/worker/logger.go
dleviminzi: On the counter it is set back to zero. On [line 102](https://github.com/beam-cloud/beta9/blob/d1aa21c21c395ba413fed640667222bda1933357/pkg/worker/logger.go#L102), it is incremented once more after the above condition. Lets say the limit is 10. 

The 11th log comes in and the above logic is true. So, we log those statements _and_ we increment the counter one more time. Now the counter is at 12 so any subsequent logs will no longer fall into the condition above.

dleviminzi: I changed this to use `golang.org/x/time/rate`. This way tokens are replenished.

pkg/worker/logger.go
```diff
@@ -72,7 +74,25 @@ func (r *ContainerLogger) CaptureLogs(containerId string, logChan chan common.Lo
 		return errors.New("container not found")
 	}
 
+	limiter := rate.NewLimiter(rate.Limit(float64(r.logLinesPerHour)/3600.0), r.logLinesPerHour)
+	rateLimitMessageLogged := false
+
 	for o := range logChan {
+		if !limiter.Allow() {
+			if !rateLimitMessageLogged {
+				log.Info().Str("container_id", containerId).Msg("Rate limit exceeded, logging at reduced rate, some logs will be dropped")
```
luke-lombardi, line 83: could prolly put this into a var `rateLimitMsg` or something so you don't need to duplicate it here and below

luke-lombardi (changes requested):

pkg/types/config.go
```diff
@@ -237,6 +237,7 @@ type WorkerConfig struct {
 	BlobCacheEnabled           bool                        `key:"blobCacheEnabled" json:"blob_cache_enabled"`
 	CRIU                       CRIUConfig                  `key:"criu" json:"criu"`
 	TmpSizeLimit               string                      `key:"tmpSizeLimit" json:"tmp_size_limit"`
+	ContainerLogLinesPerHour   int                         `key:"containerLogLinesPerHour" json:"container_log_lines_per_hour"`
```
luke-lombardi, line 240: I feel like this window should be shorter, i.e. every 10 minutes. The thing I'm concerned with is if a user has spammy logs, and then their app crashes at the 45 minute mark, they'll never know what happened. I almost feel like we should make the limits lower but keep the reset interval shorter. Another thing to consider is if there is an error that crashes the app, how we could force that to propogate. Not sure that's something we need to handle in this PR though.

dleviminzi: I think the difference between an hour and 10 minutes is sort of misleading here. It is only used to set a rate below. This value `float64(r.logLinesPerHour)/3600.0` is the number of lines per second that the limiter will allow. Given that the default log lines per hour is 6000, the current default rate is 1.666 lines per second. The reason I provided the value as log lings per hour is because it makes it easier to think about the burst setting below. A rate of 1.666 says the user can log up to 6000 lines per hour and the burst says that if they want to log them all at once, they can.

dleviminzi: We could change it to 10 minutes, but if the rate is the same the effect wont really be different.

luke-lombardi (approved):

PR #883, opened 2025-01-22
Feat: Add on_deploy hook to abstractions with deployments

jsun-m:
Add an on_deploy hook to endpoint/taskqueue/container/function deployments.
This is ran right after the `prepare_runtime` of the deployable abstraction and the `stub_id` of the `on_deploy` is passed to the `deployment` object.

```
from beam import Volume, endpoint, function 


my_volume = Volume(name="weights")


@function(volumes=[my_volume])
def func():
  # download weights to a volume, or similar...


@endpoint(on_deploy=func, volumes=[my_volume])
def handler():
```


pkg/gateway/gateway.proto
```diff
@@ -276,6 +276,7 @@ message GetOrCreateStubResponse {
 message DeployStubRequest {
   string stub_id = 1;
   string name = 2;
+  string on_deploy_stub_id = 3;
```
luke-lombardi, line 279: Is there a way we can do this without the on_deploy_stub_id and the data model changes? I don't know if its necessary to track the stub ID of the run function since we are storing the source code that stub along with the deployed stub, if that makes sense.

pkg/repository/backend_postgres_migrations/021_deployment_on_deploy_stub.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/pressly/goose/v3"
+)
+
+func init() {
+	goose.AddMigrationContext(upAddOnDeployStubIdToDeployment, downRemoveOnDeployStubIdFromDeployment)
```
luke-lombardi, line 12: Same comment as above, I would do this without the data model changes. I appreciate that we're trying to track it, but since we have the source code of the deployed stub you should be able to see what you're on_deploy hook would have looked like.

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/taskqueue.py
```diff
@@ -8,6 +8,7 @@
     TASKQUEUE_DEPLOYMENT_STUB_TYPE,
     TASKQUEUE_SERVE_STUB_TYPE,
     TASKQUEUE_STUB_TYPE,
+    AbstractCallableWrapper,
```
luke-lombardi, line 11: Not sure I follow why we can't use CallableWrapper here, can you elaborate?

pkg/repository/backend_postgres_migrations/021_deployment_on_deploy_stub.go
jsun-m: removed

pkg/gateway/gateway.proto
jsun-m: I removed the db change but I will still pass in the external_id here so we can update the config

sdk/src/beta9/abstractions/taskqueue.py
jsun-m: _CallableWrapper is initialized in each abstraction and each are slightly different. Also importing any of them will result in a circular import

pkg/repository/backend_postgres.go
```diff
@@ -832,8 +832,28 @@ func (r *PostgresBackendRepository) GetStubByExternalId(ctx context.Context, ext
 	return &stub, nil
 }
 
-// Volume
+func (r *PostgresBackendRepository) UpdateStubConfig(ctx context.Context, stubId uint, config types.StubConfigV1) error {
```
dleviminzi, line 835: Just making sure I understand here. Is the primary reason we update the stub config so that the user is able to find the stub ID of the on deploy function that ran?

jsun-m: yea that is correct

jsun-m:
I updated stub_config to have both `on_deploy` and `on_deploy_id`.

`on_deploy` will just be a reference to the `module:func` like `on_start` and `handler`
`on_deploy_id` is the actual stub id of the `on_deploy` method. This is consistent with the `image_id` field that later gets inserted

pkg/gateway/services/stub.go
```diff
@@ -216,6 +217,23 @@ func (gws *GatewayService) DeployStub(ctx context.Context, in *pb.DeployStubRequ
 		version = lastestDeployment.Version + 1
 	}
 
+	if in.OnDeployStubId != "" {
```
luke-lombardi, line 220: This is looking good, but is there a way to reverse the order of operations so we have the stub ID at the time we create the deployed stub? In other words, so we do not have to update the stub config, but can put it in from the beginning? @jsun-m

luke-lombardi (approved):

PR #884, opened 2025-01-22
Feat: Set build container spin up timeout based on image size

dleviminzi:
Resolve BE-2217

I pulled the skopeo logic out of the image client and into its own struct. This makes it easy to use the inspect functionality in the gateway's builder. Aside from that, I do a simple calculation of 120 nanoseconds per byte. That evaluates to 2 minutes per GB of the base image. That is a very generous amount of time. 

TODO: 
- [ ] make time multiplier config?


pkg/abstractions/image/build.go
```diff
@@ -230,11 +232,13 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
 	case opts.Dockerfile != "":
 		opts.addPythonRequirements()
 		dockerfile = &opts.Dockerfile
+		containerSpinupTimeout = 1 * time.Hour
```
dleviminzi, line 235: I dont think its realistically possible to guess how long a dockerfile build might take. One hour might be excessive, but I do think it should be a fairly high timeout.

pkg/abstractions/image/build.go
```diff
@@ -46,6 +46,7 @@ type Builder struct {
 	tailscale     *network.Tailscale
 	eventBus      *common.EventBus
 	rdb           *common.RedisClient
+	imageCopier   *common.SkopeoCopier
```
luke-lombardi, line 49: should just this be called skopeoClient? same with the struct. Unless there's a better reason for the name skopeocopier

pkg/abstractions/image/build.go
luke-lombardi, line 235: if that's the case let's move this to a constant.

pkg/abstractions/image/build.go
```diff
@@ -830,3 +828,33 @@ func tagOrDigest(digest string, tag string) string {
 	}
 	return digest
 }
+
+func (b *Builder) estimateNewBaseImageTime(ctx context.Context, opts *BuildOpts) time.Duration {
```
luke-lombardi, line 832: not sure I like the naming here

pkg/abstractions/image/build.go
```diff
@@ -830,3 +828,33 @@ func tagOrDigest(digest string, tag string) string {
...
+		log.Error().Err(err).Msg("failed to inspect image")
+		return defaultContainerSpinupTimeout
+	}
+
+	imageSizeBytes := 0
+	for _, layer := range imageMetadata.LayersData {
+		imageSizeBytes += layer.Size
+	}
+
+	secondsPerByte := 120 * time.Nanosecond
```
luke-lombardi, line 845: I don't follow how this is defined. Where is this 120 from?

pkg/common/skopeo.go
```diff
@@ -0,0 +1,162 @@
...
+	LayersData    []struct {
+		MIMEType    string `json:"MIMEType"`
+		Digest      string `json:"Digest"`
+		Size        int    `json:"Size"`
+		Annotations any    `json:"Annotations"`
+	} `json:"LayersData"`
+	Env []string `json:"Env"`
+}
+
+func NewSkopeoCopier(config types.AppConfig) *SkopeoCopier {
```
luke-lombardi, line 51: Same comment as above, should be just `SkopeoClient`

luke-lombardi (changes requested):

pkg/worker/image.go
```diff
@@ -65,18 +60,19 @@ func getImageMountPath(workerId string) string {
...
 	imageCachePath     string
 	imageMountPath     string
 	imageBundlePath    string
-	pullCommand        string
-	pDeathSignal       syscall.Signal
 	mountedFuseServers *common.SafeMap[*fuse.Server]
-	commandTimeout     int
-	debug              bool
-	creds              string
+	imagePuller        ImageCopier
```
luke-lombardi, line 75: why not just call this skopeoclient?  that's what it is. and we shouldn't need the interface above right?

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #892, opened 2025-01-26
Feat: repository services (Pt. 1)

luke-lombardi:
Adds the go2proto package into our protobuf generation script, and automatically generates the protobuf message types. Any struct with the // @go2proto comment above them in types will be exported as protobuf messages.

This is part 1, which lays out the various structs and sets up the grpc services for the worker repo and container repo. The follow up will implement the actual repository services needed for the worker with the generated types.


pkg/gateway/gateway.proto
```diff
@@ -211,16 +212,7 @@ message StopTasksResponse {
...
-}
-
-message MountPointConfig {
-  string bucket_name = 1;
-  string access_key = 2;
-  string secret_key = 3;
-  string endpoint_url = 4;
-  string region = 5;
-  bool read_only = 6;
+  optional types.MountPointConfig config = 3;
```
luke-lombardi, line 215: You can see an example of how this is used here.

pkg/scheduler/pool_external.go
```diff
@@ -242,8 +246,13 @@ func (wpc *ExternalWorkerPoolController) createWorkerOnMachine(workerId, machine
 		return nil, err
 	}
 
+	token, err := wpc.backendRepo.CreateToken(wpc.ctx, 1, types.TokenTypeWorker, true)
```
luke-lombardi, line 249: We really need a helper to get the workspace admin here somehow...

pkg/repository/container_redis.go
```diff
@@ -115,7 +115,9 @@ func (cr *ContainerRedisRepository) GetContainerExitCode(containerId string) (in
 	return exitCode, nil
 }
 
-func (cr *ContainerRedisRepository) UpdateContainerStatus(containerId string, status types.ContainerStatus, expiry time.Duration) error {
+func (cr *ContainerRedisRepository) UpdateContainerStatus(containerId string, status types.ContainerStatus, expirySeconds int64) error {
+	expiry := time.Duration(expirySeconds) * time.Second
```
nickpetrovic, line 119: Is time.Duration not supported?

pkg/repository/services/container_repo.go
nickpetrovic, line 1: Are we unable to have these services live under pkg/gateway/services? Seems like this should go there since its where most of the core gRPC API lives.

Makefile
```diff
@@ -37,10 +37,6 @@ worker:
 	docker push localhost:5001/beta9-worker:$(workerTag)
 	bin/delete_workers.sh
 
-proxy:
-	docker build . --target build -f ./docker/Dockerfile.proxy -t localhost:5001/beta9-proxy:$(tag)
-	docker push localhost:5001/beta9-proxy:$(tag)
-
```
nickpetrovic, line 43: Can we remove the helm and kustomize stuff too?

luke-lombardi: Yeah I can, want me to do that in this PR? was a shitload of changes

pkg/repository/services/container_repo.go
luke-lombardi: Nah, I can move it. I went back and forth on this since we keep some of the other grpc stuff "close" to wherever

luke-lombardi: Updated

Makefile
nickpetrovic: Hm, maybe not then. In a follow PR up maybe.

nickpetrovic (approved):

PR #898, opened 2025-01-27
Fix: Upgrade to Cuda 12.4, use defaults for missing env vars, include more libs

dleviminzi:
Resolve BE-2270

This PR does three things 

1. It upgrades the worker to CUDA 12.4.
2. If CUDA_HOME is not set or provided anywhere, use a default location for CUDA 12.4
3. Install encode, decode, nvcc, compute, utils, etc in worker. This all ends up available through mounts so runner size/cold start will not be impacted. 

WARNING: Do not release until A10G cuda version/driver is updated

### DIFF
--- a/docker/Dockerfile.worker
+++ b/docker/Dockerfile.worker
@@ -94,7 +94,7 @@ RUN go build -o /usr/local/bin/worker ./cmd/worker/main.go
 
 # final image
 # ========================
-FROM nvidia/cuda:12.3.1-base-ubuntu22.04 AS release
+FROM nvidia/cuda:12.4.1-base-ubuntu22.04 AS release
 FROM release AS dev
 
 FROM ${BASE_STAGE} AS final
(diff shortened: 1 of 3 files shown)


docker/Dockerfile.worker
```diff
@@ -109,6 +109,8 @@ RUN apt-get update && \
     apt-get update && \
     apt-get install psmisc
 
+RUN apt-get install -y cuda-nvcc-12-4 libnvidia-compute-550 libnvidia-encode-550 libnvidia-decode-550 libnvidia-extra-550 nvidia-utils-550
```
luke-lombardi, line 112: Can you check how much this adds to the worker image? Want to make sure we're keeping track of this

luke-lombardi (approved):
@dleviminzi should we move this to draft until we get the OCI nodes updated?

docker/Dockerfile.worker
dleviminzi: 761 MB

luke-lombardi: 👍

PR #904, opened 2025-01-28
Feat: Drain & cordon by machine id

dleviminzi:
I plan on using this to cycle the A10s

### DIFF
--- a/sdk/src/beta9/cli/worker.py
+++ b/sdk/src/beta9/cli/worker.py
@@ -269,9 +269,7 @@ def machine():
     required=True,
 )
 @extraclick.pass_service_client
-@click.pass_context
 def drain_machine(
-    ctx: click.Context,
     service: ServiceClient,
     machine_id: str,
 ):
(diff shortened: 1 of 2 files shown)


sdk/src/beta9/cli/worker.py
```diff
@@ -207,3 +207,105 @@ def drain_worker(
...
+
+    terminal.info(f"Found {len(matching_workers)} workers on machine {machine_id}")
+
+    for worker_id in matching_workers:
+        res = service.gateway.drain_worker(DrainWorkerRequest(worker_id=worker_id))
+        if not res.ok:
+            text = res.err_msg.capitalize()
+            terminal.warn(text)
+        else:
+            terminal.success(f"Worker {worker_id} has been drained.")
```
nickpetrovic, line 264: I'm not sure if we can invoke across modules, but can we try it so we don't have duplicate logic?

See how to invoke in Click here.
https://github.com/beam-cloud/beta9/blob/main/sdk/src/beta9/cli/deployment.py#L69

nickpetrovic: Oh hm, you may want `ctx.invoked_subcommand(...)`

nickpetrovic (approved):

PR #905, opened 2025-01-29
Fix: Merge paths during env var injection

dleviminzi:
Resolve BE-2295

Previously we would just add more env vars with reference like this: `LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu:/usr/lib/worker/x86_64-linux-gnu:/usr/local/nvidia/lib64:/usr/local/cuda-12.3/targets/x86_64-linux/lib:$LD_LIBRARY_PATH`. The issue with this approach is that references like `$LD_LIBRARY_PATH` are never actually expanded. 

To fix this, we can inspect the existing environment and add our additional things to the tail of the existing vars or create them if they don't exist at all. 

While working on this, I made another change. Previously, we would set the cuda env vars even if the host key contains nothing. There is no point in doing that and it makes the tests more confusing, so I removed it.


pkg/worker/nvidia.go
```diff
@@ -216,72 +217,73 @@ func (c *ContainerNvidiaManager) InjectEnvVars(env []string, options *ContainerO
...
+		imageEnvVars[name] = value
 	}
 
 	cudaVersion := defaultContainerCudaVersion
-	existingCudaVersion, existingCudaFound := initialEnvVars["CUDA_VERSION"]
+	existingCudaVersion, existingCudaFound := imageEnvVars["CUDA_VERSION"]
 	if existingCudaFound {
 		splitVersion := strings.Split(existingCudaVersion, ".")
 		if len(splitVersion) >= 2 {
-			major := splitVersion[0]
```
luke-lombardi, line 241: I prefer the clarity the major/minor variables add

pkg/worker/nvidia.go
```diff
@@ -216,72 +217,73 @@ func (c *ContainerNvidiaManager) InjectEnvVars(env []string, options *ContainerO
...
-
-		if existingCudaFound {
-			if value, exists := initialEnvVars[key]; exists {
-				cudaEnvVarValue = value
-			} else {
-				continue
-			}
+		hostValue := os.Getenv(key)
+		switch {
+		case imageEnvVars[key] != "":
```
luke-lombardi, line 250: won't this panic if it's not found? Why not use the value, exists here? IMO this is worse

pkg/worker/nvidia.go
```diff
@@ -216,72 +217,73 @@ func (c *ContainerNvidiaManager) InjectEnvVars(env []string, options *ContainerO
...
+	for key, value := range imageEnvVars {
+		modifiedEnv = append(modifiedEnv, fmt.Sprintf("%s=%s", key, value))
 	}
 
-	env = append(env, cudaEnvVars...)
+	return modifiedEnv, existingCudaFound
+}
+
+func mergePaths(pathName string, initEnv map[string]string, mergeIn []string) {
+	if initEnv[pathName] == "" {
```
luke-lombardi, line 271: Same comment as above, I don't like the way we're checking this

luke-lombardi (changes requested):

pkg/worker/nvidia.go
dleviminzi: missing keys just return the zero value of the type, [they don't panic](https://goplay.tools/snippet/kEPbEtiQSzq)

pkg/worker/nvidia.go
dleviminzi: Given that we don't care whether the key exists, just whether it has any value I don't think its bad to check this way. We're just checking if there's a non-empty value, regardless of whether the key exists or not.

luke-lombardi (approved):

PR #906, opened 2025-01-29
GRPC repository services: Pt. 2

luke-lombardi:
Puts all container/worker repository methods accessed inside worker behind GRPC services.


pkg/common/config.default.yaml
```diff
@@ -35,7 +35,7 @@ gateway:
   host: beta9-gateway
   invokeURLType: path
   grpc:
-    externalHost: localhost
+    externalHost: beta9-gateway
```
luke-lombardi, line 38: @nickpetrovic is this ok?

pkg/gateway/services/repository/worker_repo.go
```diff
@@ -13,3 +16,172 @@ type WorkerRepositoryService struct {
...
+	return &pb.RemoveContainerFromWorkerResponse{Ok: true}, nil
+}
+
+func (s *WorkerRepositoryService) GetWorkerById(ctx context.Context, req *pb.GetWorkerByIdRequest) (*pb.GetWorkerByIdResponse, error) {
+	worker, err := s.workerRepo.GetWorkerById(req.WorkerId)
+	if err != nil {
+		return &pb.GetWorkerByIdResponse{Ok: false, ErrorMsg: err.Error()}, nil
+	}
+
+	return &pb.GetWorkerByIdResponse{Ok: true, Worker: &pb.Worker{
```
luke-lombardi, line 83: replace with ToProto

pkg/types/scheduler.go
```diff
@@ -122,6 +139,89 @@ func (c *ContainerRequest) IsBuildRequest() bool {
...
+				Id: uint32(c.Stub.Object.Id),
+			},
+		},
+		RetryCount:        int64(c.RetryCount),
+		PoolSelector:      c.PoolSelector,
+		Preemptable:       c.Preemptable,
+		CheckpointEnabled: c.CheckpointEnabled,
+		Timestamp:         timestamppb.New(c.Timestamp),
+		BuildOptions:      &pb.BuildOptions{
+			// SourceImage:      c.BuildOptions.SourceImage,
```
luke-lombardi, line 182: Fill in the rest of these types.

pkg/types/scheduler.go
```diff
@@ -122,6 +139,89 @@ func (c *ContainerRequest) IsBuildRequest() bool {
...
+		WorkspaceId: in.WorkspaceId,
+		Workspace: Workspace{
+			Name: in.Workspace.Name,
+		},
+		StubId:            in.StubId,
+		Timestamp:         in.Timestamp.AsTime(),
+		CheckpointEnabled: in.CheckpointEnabled,
+		Preemptable:       in.Preemptable,
+		BuildOptions: BuildOptions{
+			// SourceImage:      in.BuildOptions.SourceImage,
```
luke-lombardi, line 217: Same comment as above.

nickpetrovic (approved):

pkg/worker/worker.go
```diff
@@ -266,58 +273,86 @@ func (s *Worker) Run() error {
...
+				exitCode = serr.ExitCode
+			}
+
+			_, err = handleGRPCResponse(s.containerRepoClient.SetContainerExitCode(ctx, &pb.SetContainerExitCodeRequest{
+				ContainerId: containerId,
+				ExitCode:    int32(exitCode),
+			}))
+			if err != nil {
+				log.Error().Str("container_id", containerId).Err(err).Msg("failed to set exit code")
+			}
```
dleviminzi, line 347: Previously we were setting the exit code before releasing the lock. How come the order is swapping? Not sure if the original order was just coincidental or actually important.

pkg/worker/worker.go
```diff
@@ -266,58 +273,86 @@ func (s *Worker) Run() error {
...
+		})
+		if err != nil {
+			if err == context.Canceled || s.ctx.Err() == context.Canceled {
+				break
+			}
 
-			s.containerLock.Lock()
+			time.Sleep(containerRequestStreamInterval)
+			continue
+		}
```
dleviminzi, line 291: Bit of a nitpick, but I think using a ticker to do delayed loops like this is a bit cleaner. One advantage is that when the worker context is cancelled it will stop even if its in the delay window. Something like this:

```go
	ticker := time.NewTicker(containerRequestStreamInterval)
	defer ticker.Stop()

	handlingRequests := true
	for handlingRequests {
		select {
		case <-s.ctx.Done():
			handlingRequests = false
			continue
		case <-ticker.C:
			stream, err := s.workerRepoClient.GetNextContainerRequest(s.ctx, &pb.GetNextContainerRequestRequest{
				WorkerId: s.workerId,
			})
			if err != nil {
				if errors.Is(err, context.Canceled) || s.ctx.Err() == context.Canceled {
					handlingRequests = false
				}
				continue
			}

			lastContainerRequest := time.Now()
			for {
				response, err := stream.Recv()
				if err != nil {
					break
				}

				if response.ContainerRequest != nil {
					lastContainerRequest = time.Now()
					request := types.NewContainerRequestFromProto(response.ContainerRequest)
					s.handleContainerRequest(request)
				}

				if s.shouldShutDown(lastContainerRequest) {
					break containerRequestStream
				}
			}
		}
	}

	return s.shutdown()
```

pkg/worker/worker.go
luke-lombardi: Nah shouldn't matter either way. Once the container is dead the order of operations is less important.

pkg/worker/worker.go
luke-lombardi: Usually I would agree with that, but in this case, the ticker makes it less clear IMO. Here were only retrying / sleeping if there is a stream error, and setting up a ticker sort of implies that we're creating a new stream every `containerRequestStreamInterval`. I know effectively that might not be the case, but at first glance the ticker implies something will happen every X duration. I think the sleep is simpler and cleaner here.

PR #910, opened 2025-01-29
Fix: Add validation messages for on_deploy

jsun-m:


sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -351,6 +351,20 @@ def _parse_gpu(self, gpu: Union[GpuTypeAlias, List[GpuTypeAlias]]) -> str:
         else:
             return GpuType(gpu).value
 
+    def _validate_on_deploy(self, func: Callable):
+        if func is None:
+            return None
+
+        if not callable(func):
+            raise terminal.error(
+                "Unable to build: on_deploy must be a callable function with a function decorator"
```
luke-lombardi, line 360: since this string is the same in both cases I'd move it to a var and re-use it twice. also, I would maybe change the string to "on_deploy must be a callable function with a function decorator"

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/base/runner.py
jsun-m: I'll just reuse the same message for both.
`on_deploy must be a callable function with a function decorator` encases the other message as well

luke-lombardi (approved):

PR #912, opened 2025-01-30
Feat: Detect existing python

dleviminzi:
Resolve BE-2249

If a user does not specify a python version, we should first see if the image they are using contains one before installing 3.10. Many images do (pytorch, jax, miniconda, etc) and we currently just install a new version and mess them up. 

Note: this only works once this fix is in https://github.com/beam-cloud/beta9/pull/905


nickpetrovic (changes requested):
If I understand this correctly, this change makes it so the server defines the default python version if the user specifies python3 as its python version. 

I think we need to fix a couple of issues here if the server is going to decide the default python version.

1. Update the sdk's Image class to no longer detect the python version and instead just use python3. This might cause existing images to be rebuilt. There's a new function called detected_python_version that would need to be removed.
2. Add docstrings to the PythonVersion enum that mentions what Python3 is
3. Add docstrings in the Image class that mentions how the default the PythonVersion works
4. Use a constant and single place to define the default python version, maybe even consider a setting for it. This PR has it peppered all over the place which is definitely going to cause issues down the road.

pkg/common/config.default.yaml
```diff
@@ -181,7 +182,6 @@ worker:
   terminationGracePeriod: 30
   addWorkerTimeout: 10m
   blobCacheEnabled: false
-  containerLogLinesPerHour: 6000
```
nickpetrovic, line 184: Is this intentional?

sdk/src/beta9/type.py
```diff
@@ -33,7 +33,9 @@ class TaskExitCode:
 
 class PythonVersion(str, Enum):
     """
-    An enum that defines versions of Python.
+    An enum that defines versions of Python. The default version is `python3`. When this version
+    is used, we will use whatever version of Python 3 exists in the image. If none exists, we will
```
nickpetrovic, line 37: I might rephrase and format this to the following.

> An enum that defines versions of Python.
>
> The default version is `python3`. This defaults to `python3` already in the image. If `python3` does not exist in the image, then the default version of Python will be determined by the server (e.g. Python 3.10).

pkg/common/config.default.yaml
```diff
@@ -66,12 +66,13 @@ fileService:
 imageService:
   localCacheEnabled: true
   # 120 ns per byte is a rough estimate of the time it takes per byte to pull, unpack, and archive an image.
-  # In the worst case, we have seen 10gb images take 20 minutes. That is 1e+10 bytes taking 1200 seconds.
-  # That gives 1200 / 1e+10 = 120 ns per byte.
+	# In the worst case, we have seen 10gb images take 20 minutes. That is 1e+10 bytes taking 1200 seconds.
+	# That gives 1200 / 1e+10 = 120 ns per byte.
```
nickpetrovic, line 70: nitpick, can we just revert this? shouldn't change things that don't need to be changed.

dleviminzi: Yeah my bad. Editor is being a bit wonky when I'm ssh'd into my GPU machine. Reverting this

nickpetrovic (approved):
I think this look good. Still need to review the env changes though.

pkg/types/python_versions.go
```diff
@@ -0,0 +1,16 @@
+package types
```
luke-lombardi, line 1: I can imagine we might have other types related to python stuff. Maybe rename this to py.go or python.go?

luke-lombardi (changes requested):
one small nitpick and then g2g

luke-lombardi (approved):

PR #913, opened 2025-01-30
feat: update local setup

nickpetrovic:
- setup.sh - update binary versions
- setup.sh - pin kubectl version to stable 1.29
- setup.sh - fail on any error
- setup.sh - improve log statements
- k3d.yaml - use k3s being used in production
- manifests/k3d - remove nvidia device plugin, its already in kustomize. addresses user reported issue.
- manifests/overlays/cluster-dev - remove resources from beta9-gateway deployment. addresses user reported issue.
- manifests/overlays/remote-node-aws - not needed
- manifests/overlays/remote-node-crusoe - not needed

Resolve BE-2308


mernit (approved):

manifests/kustomize/overlays/cluster-dev/kustomization.yaml
```diff
@@ -12,6 +12,13 @@ components:
 - ../../components/localstack
 - ../../components/beta9-gateway
 patches:
+# Remove resources from beta9 components
+- target:
+    kind: Deployment
+    name: beta9-gateway
+  patch: |-
+    - op: remove
+      path: /spec/template/spec/containers/0/resources
```
dleviminzi, line 21: Can you explain a bit more what this is doing?

nickpetrovic: This removes the resource key entirely. For this deployment in particular, we request 4GiB of RAM. This patch removes that requirement. This only affects the cluster-dev overlay which is our local setup.

One of the Cedana people ran into resource issues here with this. In a way, it protects the user from their own mistakes, but also this is a dev environment and IMO it should allow the developer to figure out what cpu, memory, etc they may or may not need.

nickpetrovic: One pain point I've run into is running multiple gateways locally. They don't use 4GiB out the gate, if ever. This lets me add multiple gateways without needing 4GiB of RAM per gateway.

PR #920, opened 2025-02-03
Feat: Add  `headless` flag for remote functions

jsun-m:
Added ability to cancel task if client-gateway connection is closed prematurely.

Also modified Function's Task.Cancel function to add `endedAt` to its task info as well as only update the status if the task was canceled in an `inflight` state (pending, retry, running)


sdk/src/beta9/abstractions/function.py
```diff
@@ -101,6 +103,7 @@ def __init__(
         name: Optional[str] = None,
         task_policy: TaskPolicy = TaskPolicy(),
         on_deploy: Optional[AbstractCallableWrapper] = None,
+        headless: bool = True,
```
dleviminzi, line 106: I think the default should be false. Right @luke-lombardi? In the case of function runs, I think the expectation of a user is that stopping their program will stop the every part of the execution. That is what would happen with normal code and we are providing an abstraction that is supposed to act like normal code.

jsun-m: That reasoning makes sense

pkg/types/backend.go
```diff
@@ -170,6 +170,12 @@ const (
 	TaskStatusRetry     TaskStatus = "RETRY"
 )
 
+var InflightTaskStatuses = map[TaskStatus]bool{
```
luke-lombardi, line 173: what exactly is this needed for?

luke-lombardi (changes requested):

pkg/abstractions/function/task.go
```diff
@@ -179,15 +180,22 @@ func (t *FunctionTask) Cancel(ctx context.Context, reason types.TaskCancellation
 		return err
 	}
 
+	if !types.InflightTaskStatuses[task.Status] {
+		return nil
+	}
```
dleviminzi, line 185: Bit of a nitpick, but I think this could be a function with a clearer name like: 

```go 
func taskIsInflight(status string) {
    return status == ... || status == ...
}
```

luke-lombardi:
@jsun-m can you take another look at this PR?

pkg/types/backend.go
jsun-m: This is used in the `Cancel` function of the task in the cases where `Cancel` is called after a task already has been `Completed`.

pkg/abstractions/function/task.go
jsun-m: How about now?

dleviminzi (approved):

pkg/abstractions/function/task.go
dleviminzi: noice

PR #924, opened 2025-02-04
feat: add 25% more memory to container hard limit

nickpetrovic:
- Memory soft limit is what the user specifies
- Memory hard limit is what the user specifies + 25% of what the user specifies
- Kill all processes in a runc container if it OOMs

Resolve BE-2326

### DIFF
--- a/sdk/pyproject.toml
+++ b/sdk/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "beta9"
-version = "0.1.167"
+version = "0.1.168"
 description = ""
 authors = ["beam.cloud <support@beam.cloud>"]
 packages = [
--- a/pkg/worker/lifecycle.go
+++ b/pkg/worker/lifecycle.go
@@ -36,6 +36,10 @@ const (
 //go:embed base_runc_config.json
 var baseRuncConfigRaw string
 
+var cgroupV2Parameters map[string]string = map[string]string{
+	"memory.oom.group": "1", // Kill all processes in the cgroup when an OOM occurs
+}
+
 // handleStopContainerEvent used by the event bus to stop a container.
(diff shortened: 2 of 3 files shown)


sdk/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "beta9"
-version = "0.1.167"
+version = "0.1.168"
```
nickpetrovic, line 3: Bumping version for an earlier build.

dleviminzi (approved):

luke-lombardi (changes requested):
This looks good, but I have a concern about task queue deployments. In the case where there is a memory spike, but only one of the processes fail (and the steady state is under the limit), it may force the entire deployment to shut down (if I'm understanding this correctly). Same thing with endpoints / any multi worker deployments.

nickpetrovic:
> This looks good, but I have a concern about task queue deployments. In the case where there is a memory spike, but only one of the processes fail (and the steady state is under the limit), it may force the entire deployment to shut down (if I'm understanding this correctly). Same thing with endpoints / any multi worker deployments.

If the total amount of memory plus the overhead is completely used up, then yes, the entire runc container will be sigkilled. If this happens, the end user should increase their memory limit. It's pretty much the same with resource limits in k8s. 

The 25% buffer should help with this too (unlike k8s). Is there another number we should use? Or do you have another suggestion?

nickpetrovic:
Made a couple of more changes.

- Moved the cgroup v2 params (memory.oom.group) to only be enabled when `worker.runcResourcesEnforced` is `true` so we can toggle this behavior
- Updated messaging to say the container is killed vs a process within the container is killed
- Updated comments

PR #925, opened 2025-02-04
feat: improve oom message with serves

nickpetrovic:
- Detect OOMKill in watchOOMEvents
- Handle OOMKill signal in serve logic
- Move serve final message output to server side
- Add OOM error message to server side as a response when OOMKill signal is detected

Resolve BE-2325


nickpetrovic:
Example of what this looks like.

**OOM**

Screenshot 2025-02-03 at 9 03 34 PM

**Regular timeout, beam container stop, or UI stop container**

Screenshot 2025-02-04 at 9 41 42 AM

mernit (approved):

pkg/abstractions/endpoint/serve.go
```diff
@@ -65,10 +66,19 @@ func (es *HttpEndpointService) StartEndpointServe(in *pb.StartEndpointServeReque
 	}
 
 	exitCallback := func(exitCode int32) error {
-		if err := stream.Send(&pb.StartEndpointServeResponse{Done: true, ExitCode: int32(exitCode)}); err != nil {
-			return err
+		output := "Container was stopped."
+		if exitCode != 0 {
+			output = fmt.Sprintf("Container failed with exit code %d", exitCode)
+			if exitCode == 137 {
```
luke-lombardi, line 72: Can we move this to a constant ? I feel like we had sigterm constants somewhere

pkg/abstractions/taskqueue/serve.go
```diff
@@ -56,10 +57,19 @@ func (tq *RedisTaskQueue) StartTaskQueueServe(in *pb.StartTaskQueueServeRequest,
 	}
 
 	exitCallback := func(exitCode int32) error {
-		if err := stream.Send(&pb.StartTaskQueueServeResponse{Done: true, ExitCode: int32(exitCode)}); err != nil {
-			return err
+		output := "Container was stopped."
+		if exitCode != 0 {
+			output = fmt.Sprintf("Container failed with exit code %d", exitCode)
+			if exitCode == 137 {
```
luke-lombardi, line 63: Same comment as above

pkg/worker/cr.go
```diff
@@ -234,6 +236,9 @@ func (s *Worker) waitForRestoredContainer(ctx context.Context, containerId strin
 			}
 
 			if state.Status != types.RuncContainerStatusRunning && state.Status != types.RuncContainerStatusPaused {
+				if isOOMKilled {
```
luke-lombardi, line 239: Hm, we can consolidate this into a single call that jsut takes the exit code? I.e. override the exit code and have one call to cleanup

luke-lombardi (changes requested):

pkg/abstractions/endpoint/serve.go
nickpetrovic: Yep, updated.

pkg/abstractions/taskqueue/serve.go
nickpetrovic: Updated.

sdk/src/beta9/abstractions/taskqueue.py
```diff
@@ -235,23 +234,28 @@ def notify(*_, **__):
...
 
-        if r is None or not r.done or r.exit_code != 0:
-            terminal.error("Serve container failed ❌")
+        if r is None:
+            return terminal.error("Serve failed ❌")
+
+        if not r.done or r.exit_code != 0:
+            return terminal.error(f"{r.output} ❌")
 
-        terminal.warn("Taskqueue serve timed out. Container has been stopped.")
```
luke-lombardi, line 254: Do we still have the timeout message? that's still a valid exit reason

pkg/worker/cr.go
nickpetrovic: Updated.

sdk/src/beta9/abstractions/taskqueue.py
nickpetrovic: So I removed the "timed out" part of the messaging because when you stop the container from the CLI or the frontend, you see the timed out message. We don't distinguish between timing out and a regular stop. 

I can add the "timed out" messaging back if you think it makes sense?

luke-lombardi (approved):

Issue #921, opened 2025-02-03
Kubernetes install is not working


NguyenIconAI:
## Bug description

Hi, I'm following this doc to setup a local cluster https://docs.beta9.beam.cloud/installation/kubernetes-install but the first command is not working:
```
$ helm repo add beta9 oci://public.ecr.aws/n4e0e1y0/beta9-chart:0.1.58

Error: looks like "oci://public.ecr.aws/n4e0e1y0/beta9-chart:0.1.58" is not a valid chart repository or cannot be reached: invalid_reference: invalid tag
```

## How to reproduce

1. Install k8s
2. Install helm latest version (helm 3.14)
3. helm repo add beta9 oci://public.ecr.aws/n4e0e1y0/beta9-chart:0.1.58

## Environment

- Beam Self-Hosted

## Additional context

I tried an alternative by:
- Install localstack first:
```
$ kubectl apply -f https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/local-path-storage.yaml
$ helm install localstack localstack/localstack -n localstack --values=- <<EOF
extraEnvVars:
- name: SERVICES
  value: "s3"
enableStartupScripts: true
startupScriptContent: |
  #!/bin/bash
  awslocal s3 mb s3://juicefs
  awslocal s3 mb s3://logs
persistence:
  enabled: true
  storageClass: local-path
  accessModes:
  - ReadWriteOnce
  size: 50Gi
EOF
```
- Install beta9 helm chart:
```
$ helm install beta9 oci://public.ecr.aws/n4e0e1y0/beta9-chart \
  --version 0.1.300 \
  --namespace beta9 --create-namespace \
  --values https://raw.githubusercontent.com/beam-cloud/beta9/main/deploy/charts/beta9/values.local.yaml
```
=> Error
```
Pulled: public.ecr.aws/n4e0e1y0/beta9-chart:0.1.300
Digest: sha256:c302ed98a2a84f519fdd29b8dedf675ddfd6e6034dcc3bbe1f838ab48a1e72db
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "localstack" namespace: "" from "": no matches for kind "HelmChart" in version "helm.cattle.io/v1
```

mernit:
Hi @NguyenIconAI the documentation you linked is incorrect. These are the correct docs: https://docs.beam.cloud/v2/self-hosting/local-machine

[...]

PR #926, opened 2025-02-04
feat: log err when worker fails to start

nickpetrovic:
- worker - Log err when worker fails to start
- setup.sh - Detect aarch64 when on Linux


nickpetrovic:
Needed to help with issue https://github.com/beam-cloud/beta9/issues/921 in order to see why the worker is unable to run.

mernit (approved):

PR #929, opened 2025-02-04
fix: display correct local timezone on deploy

nickpetrovic:
Resolve BE-2220


nickpetrovic:
Testing in a Ubuntu VM with America/Denver for its timezone.

image

PR #934, opened 2025-02-05
Feat: Worker Pool failover and health metrics

luke-lombardi:
Adds a feature to worker pools which allows us to monitor their health. In the event of an outage, it will automatically cordon those workers and workloads will start being scheduled on another pool. The settings to trigger these conditions are exposed in config like so:

```
worker:
  failover:
    enabled: true
    maxPendingWorkers: 10
    maxSchedulingLatencyMs: 300000 # 5 minutes
    minMachinesAvailable: 1 # minMachinesAvailable only applies to external pools
```


pkg/repository/worker_pool_redis.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func NewWorkerPoolRedisRepository(rdb *common.RedisClient) WorkerPoolRepository {
+	return &WorkerPoolRedisRepository{rdb: rdb, lock: common.NewRedisLock(rdb)}
+}
+
+// GetWorkerPoolState retrieves a collection of health metrics for a worker pool
+func (r *WorkerPoolRedisRepository) GetWorkerPoolState(ctx context.Context, poolName string) (*types.WorkerPoolState, error) {
+	stateKey := common.RedisKeys.WorkerPoolState(poolName)
+	res, err := r.rdb.HGetAll(context.TODO(), stateKey).Result()
```
jsun-m, line 24: context.Todo

pkg/scheduler/pool_health.go
```diff
@@ -0,0 +1,234 @@
...
+	if p.workerConfig.Failover.Enabled {
+		// If failover is enabled and status is degraded, we need to cordon all workers in the pool
+		if previousState.Status != status && nextState.Status == types.WorkerPoolStatusDegraded {
+			p.eventRepo.PushWorkerPoolDegradedEvent(p.wpc.Name(), failoverReasons)
+
+			log.Warn().Str("pool_name", p.wpc.Name()).Msg("pool is degraded, cordoning all workers")
+
+			err = p.workerRepo.CordonAllWorkersInPool(p.wpc.Name())
+			if err != nil {
+				log.Error().Err(err).Msg("failed to cordon all workers in pool")
```
jsun-m, line 225: Do we really want to cordon all the workers? What about the workers that are succeeding?

pkg/types/types.proto
```diff
@@ -69,11 +70,12 @@ message ContainerState {
   string stub_id = 2;
   string status = 3;
   int64 scheduled_at = 4;
-  string workspace_id = 5;
```
jsun-m, line 72: I would be careful with this in case there is a client/server grpc proto mismatch

jsun-m (changes requested):

pkg/types/types.proto
luke-lombardi: Good call, I can move it to the end of the struct so the order is preserved

pkg/scheduler/pool_health.go
luke-lombardi: Yeah I think so. It doesn't kill the workloads running on them or anything, just says don't schedule anymore there. I can _only_ cordon pending workers if you think that makes more sense. What do you think?

pkg/types/types.proto
luke-lombardi: Updated

pkg/repository/worker_pool_redis.go
luke-lombardi: Updated

nickpetrovic (approved):
lgtm. Lets see how it behaves in staging.

PR #936, opened 2025-02-06
fix: juicefs does not exit when the worker is oomkilled

nickpetrovic:
When resource limits/requests are configured, the worker gets a oom_score_adj of -997 (or not -1000), but all juicefs processes get -1000 which means they are not oom-killable. 

This is mentioned in the [oom_score_adj](https://man7.org/linux/man-pages/man5/proc_pid_oom_score_adj.5.html) man page. 

> The value of oom_score_adj is added to the badness score
> before it is used to determine which task to kill.
> Acceptable values range from -1000 (OOM_SCORE_ADJ_MIN) to
> +1000 (OOM_SCORE_ADJ_MAX).  This allows user space to
> control the preference for OOM-killing, ranging from always
> preferring a certain task or completely disabling it from
> OOM-killing.  The lowest possible value, -1000, is
> equivalent to disabling OOM-killing entirely for that task,
> since it will always report a badness score of 0.

Changes to fix this:

- Match parent's oom_score_adj for all juicefs processes. This value is dynamically adjusted based on a pod's QoS which is based on resources limits/requests. This ensures juicefs gets the same oom_score_adj as the worker.
- Don't use daemon mode for juicefs because it immediately creates a defunct process since it's started in a goroutine anyway

Resolve BE-2334

### DIFF
--- a/pkg/common/proc.go
+++ b/pkg/common/proc.go
@@ -1,0 +1,114 @@
+package common
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/shirou/gopsutil/v4/process"
+)
+
+var (
+	ErrInvalidScore = errors.New("oom score must be between -1000 and 1000")
+)
+
+type PID interface {
+	int | int32 | uint32
+}
+
+func MatchParentOOMScoreAdj[T PID](pid T) error {
+	score, err := GetOOMScoreAdj(os.Getpid())
+	if err != nil {
+		return err
+	}
+
+	procs, err := FindProcesses(pid)
+	if err != nil {
+		return err
+	}
+
+	var errs []error
+	for _, p := range procs {
+		if err := SetOOMScoreAdj(int(p.Pid), score); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
+	return nil
+}
+
+func GetOOMScoreAdj[T PID](pid T) (int, error) {
+	if pid <= 0 {
+		return 0, fmt.Errorf("invalid pid: %d", pid)
+	}
+
+	path := fmt.Sprintf("/proc/%d/oom_score_adj", pid)
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return 0, fmt.Errorf("failed to read oom_score_adj: %v", err)
+	}
+
+	score, err := strconv.Atoi(strings.TrimSpace(string(data)))
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse oom_score_adj: %v", err)
+	}
+
+	return score, nil
+}
+
+func SetOOMScoreAdj(pid, score int) error {
+	if pid <= 0 {
+		return fmt.Errorf("invalid pid: %d", pid)
+	}
+	if score < -1000 || score > 1000 {
+		return ErrInvalidScore
+	}
+
+	path := fmt.Sprintf("/proc/%d/oom_score_adj", pid)
+	if err := os.WriteFile(path, []byte(strconv.Itoa(score)), 0644); err != nil {
+		return fmt.Errorf("failed to set oom_score_adj: %v", err)
+	}
+
+	return nil
+}
+
+func FindProcesses[T PID](pid T) ([]*process.Process, error) {
+	processes, err := process.Processes()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, p := range processes {
+		if p.Pid == int32(pid) {
+			return append(FindChildProcesses(p), p), nil
+		}
+	}
+
+	return nil, fmt.Errorf("failed to find processes for pid %v", pid)
+}
+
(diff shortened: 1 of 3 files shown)


luke-lombardi (approved):
This is such a crazy find 🙌

dleviminzi:
Just making sure I understand here. This change no longer runs the mount command detached so that it remains a child. Then if the parent OOMs it overrides the OOM priority of juicefs children so that they get removed too. I guess the priority is initially set to -1000 because it's a filesystem mount?

nickpetrovic:
> This is such a crazy find 🙌

> Just making sure I understand here. This change no longer runs the mount command detached so that it remains a child. Then if the parent OOMs it overrides the OOM priority of juicefs children so that they get removed too. I guess the priority is initially set to -1000 because it's a filesystem mount?

Definitely had to dig deep into OOM territory today. But here's what it looks like today in prod. The defunct juicefs process (because of the -d flag) is -997. This process becomes useless. The important thing is it ends up forking and those processes are -1000, even though the worker is -997. 

```
ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  4 Feb05 ?        00:00:56 /usr/local/bin/worker
root          42       1  0 Feb05 ?        00:00:00 [juicefs] <defunct>
root          55       1  0 Feb05 ?        00:00:00 juicefs mount ...
root          67      55  1 Feb05 ?        00:00:19 /usr/local/bin/juicefs mount ...
root          83       1  0 Feb05 ?        00:00:00 /usr/bin/cedana ...
root         114       1  0 Feb05 ?        00:00:00 ms3 ...
root        4697       1  1 00:15 pts/0    00:00:00 python3.10 -m beta9.runner.taskqueue
root        4711       1  0 00:15 ?        00:00:00 runc events ...
root        4717    4697  0 00:15 pts/0    00:00:00 /usr/bin/python3.10 -c ...
root        4718    4697 91 00:15 pts/0    00:01:04 /usr/bin/python3.10 -c ...
root        4915       0  0 00:17 pts/0    00:00:00 bash
root        4923    4915  0 00:17 pts/0    00:00:00 ps -ef
root@ip-10-110-48-89:/workspace# cat /proc/42/oom_score_adj
-997
root@ip-10-110-48-89:/workspace# cat /proc/55/oom_score_adj
-1000
root@ip-10-110-48-89:/workspace# cat /proc/67/oom_score_adj
-1000
cat /proc/1/oom_score_adj
-997
```

@dleviminzi I'm not sure why juicefs's forked mount processes are -1000 even though its parent is not -1000. As far as I can tell, this is not configurable. If this is because its a mount, I'd like to see the docs for that.

As for detached mode, I don't believe it technically fixes anything here, but it avoids having a defunct juicefs parent process. Since we already have a goroutine to start juicefs the process, might as well make use of it. This was just a small win to clean up the process tree.

dleviminzi:
If we're not longer running detached, shouldn't the mount be stopped when the parent OOMs already? Just by nature of the go routine that started the mount being stopped.

nickpetrovic:
I'm not 100% sure. Should it? Even though we run detached, there are still two more juicefs processes. For example, this does not seem to be the case for runc containers or python processes that the worker spawns. If you're asking me to test this, I can though.

dleviminzi:
> I'm not 100% sure. Should it? Even though we run detached, there are still two more juicefs processes. For example, this does not seem to be the case for runc containers or python processes that the worker spawns - even before running runc containers in detached mode. If you're asking me to test this, I can though.

I was only curious because that's just what I would've assumed would be the case. If we've already observed it behaving differently elsewhere then I don't think it's worth testing.

nickpetrovic:
> I was only curious because that's just what I would've assumed would be the case. If we've already observed it behaving differently elsewhere then I don't think it's worth testing.

I've tested not setting the oom_score_adj and keeping juicefs in the foreground (without -d). The oom_score_adj is still -1000 for both juicefs processes.

```
ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  0 01:00 ?        00:00:00 /usr/local/bin/worker
root          25       1  1 01:00 ?        00:00:00 [rosetta] /usr/local/bin/juicefs juicefs mount ...
root          36      25  2 01:00 ?        00:00:00 [rosetta] /usr/local/bin/juicefs /usr/local/bin/juicefs mount ...
root          55       0  0 01:00 pts/0    00:00:00 bash
root          63      55  0 01:00 pts/0    00:00:00 ps -ef
root@worker-default-245e883c-npkl4:/workspace# cat /proc/25/oom_score_adj
-1000
root@worker-default-245e883c-npkl4:/workspace# cat /proc/36/oom_score_adj
-1000
root@worker-default-245e883c-npkl4:/workspace# cat /proc/1/oom_score_adj
-997
```

To reiterate what the man page says about this.

> The lowest possible value, -1000, is equivalent to disabling OOM-killing entirely for that task, since it will always report a badness score of 0.

Based on this, I think we do need to set the oom_score_adj and using foreground mode just gives us the benefit of not having a defunct juicefs process.

PR #948, opened 2025-02-07
Feat: Add cluster metrics for tasks

jsun-m:


pkg/api/v1/task.go
```diff
@@ -44,6 +44,7 @@ func NewTaskGroup(g *echo.Group, redisClient *common.RedisClient, taskRepo repos
 	g.GET("/:workspaceId/aggregate-by-time-window", auth.WithWorkspaceAuth(group.AggregateTasksByTimeWindow))
 	g.DELETE("/:workspaceId", auth.WithWorkspaceAuth(group.StopTasks))
 	g.GET("/:workspaceId/:taskId", auth.WithWorkspaceAuth(group.RetrieveTask))
+	g.GET("/cluster-metrics", auth.WithClusterAdminAuth(group.GetClusterTaskMetrics))
```
luke-lombardi, line 47: should this be called task metrics?

pkg/repository/base.go
```diff
@@ -144,6 +145,7 @@ type BackendRepository interface {
 	GetScheduledJob(ctx context.Context, deploymentId uint) (*types.ScheduledJob, error)
 	ListenToChannel(ctx context.Context, channel string) (<-chan string, error)
 	Ping() error
+	GetTaskClusterMetrics(ctx context.Context, periodStart time.Time, periodEnd time.Time) (types.TaskClusterMetrics, error)
```
luke-lombardi, line 148: GetTaskMetrics?

luke-lombardi (changes requested):

pkg/api/v1/task.go
jsun-m: the base is already `task`. Ideally we have one of these for each of the endpoints to retrieve high level data about the cluster. `/task/cluster-metrics`, `/deployment/cluster-metrics`, `/container/cluster-metrics`, etc

luke-lombardi: what about just /metrics then? to me that would be interesting cause we could add other stuff down the road

pkg/types/metric.go
```diff
@@ -8,3 +8,7 @@ var (
 	// Worker keys
 	MetricsWorkerContainerDuration = "container_duration_milliseconds"
 )
+
+type TaskClusterMetrics struct {
```
luke-lombardi, line 12: update type here

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #953, opened 2025-02-11
Feat: Transition to CDI for adding GPUs to containers

dleviminzi:
Resolve BE-2357

This PR makes large changes to how we add Nvidia devices to containers. It makes use of the nvidia container toolkit's ability to generate a [CDI spec](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html) for the devices available to the worker. Using this spec, we can leverage the package`tags.cncf.io/container-device-interface/pkg/cdi` to parse the spec and apply the required edits directly to the `spec.Spec` object. This greatly simplifies things on our end. 

Additionally, instead of using the original bundle mounted at (`/images/mnt`) this PR makes a change to use the merged layer in `tmp`. The benefit to this change is that we can write the spec to disk directly next to the bundle. This means we no longer need to use a custom runc build that supports a custom config path.

### DIFF
--- a/pkg/worker/lifecycle.go
+++ b/pkg/worker/lifecycle.go
@@ -16,6 +16,7 @@ import (
 	"github.com/beam-cloud/beta9/pkg/storage"
 	types "github.com/beam-cloud/beta9/pkg/types"
 	pb "github.com/beam-cloud/beta9/proto"
+	"tags.cncf.io/container-device-interface/pkg/cdi"
 
 	runc "github.com/beam-cloud/go-runc"
 	"github.com/opencontainers/runtime-spec/specs-go"
@@ -27,6 +28,7 @@ const (
 	defaultContainerDirectory string        = "/mnt/code"
 	specBaseName              string        = "config.json"
 	initialSpecBaseName       string        = "initial_config.json"
+	nvidiaVendorPrefix        string        = "nvidia.com/gpu"
 	runcEventsInterval        time.Duration = 5 * time.Second
 	containerInnerPort        int           = 8001 // Use a fixed port inside the container
 )
@@ -306,7 +308,6 @@ func (s *Worker) readBundleConfig(imageId string, isBuildRequest bool) (*specs.S
 // Generate a runc spec from a given request
 func (s *Worker) specFromRequest(request *types.ContainerRequest, options *ContainerOptions) (*specs.Spec, error) {
 	os.MkdirAll(filepath.Join(baseConfigPath, request.ContainerId), os.ModePerm)
-	configPath := filepath.Join(baseConfigPath, request.ContainerId, specBaseName)
 
 	spec, err := s.newSpecTemplate()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/worker/lifecycle.go
```diff
@@ -558,15 +539,43 @@ func (s *Worker) spawn(request *types.ContainerRequest, spec *specs.Spec, output
...
+		if err != nil {
+			log.Error().Str("container_id", request.ContainerId).Msgf("failed to inject devices: %v", err)
+			return
+		}
+		if len(unresolvable) > 0 {
+			log.Error().Str("container_id", request.ContainerId).Msgf("unresolvable devices: %v", unresolvable)
+			return
+		}
+	}
+
```
nickpetrovic, line 568: Curious why this [updated] logic was moved to the spawn function.

pkg/worker/lifecycle.go
```diff
@@ -27,6 +28,7 @@ const (
 	defaultContainerDirectory string        = "/mnt/code"
 	specBaseName              string        = "config.json"
 	initialSpecBaseName       string        = "initial_config.json"
+	nvidiaVendorPrefix        string        = "nvidia.com/gpu"
```
nickpetrovic, line 31: Nitpick, but the specification calls this the device kind.

> `vendor.com/class=unique_name`
> The combination of vendor ID and device class (vendor.com/class in the above example) is referred to as the device kind.

pkg/worker/lifecycle.go
```diff
@@ -558,15 +539,43 @@ func (s *Worker) spawn(request *types.ContainerRequest, spec *specs.Spec, output
...
+			log.Error().Str("container_id", request.ContainerId).Msgf("failed to assign GPUs: %v", err)
+			return
+		}
+
+		cdiCache := cdi.GetDefaultCache()
+
+		var devicesToInject []string
+		for _, device := range assignedDevices {
+			devicePath := fmt.Sprintf("%s=%d", nvidiaVendorPrefix, device)
+			devicesToInject = append(devicesToInject, devicePath)
```
nickpetrovic, line 555: Can you remind me what `device` becomes?

nickpetrovic:
Thinking about this more broadly and before these changes were introduced even, could we run into a race condition when running multiple workers on a multi-gpu node? Specifically with assigning GPUs to a container?

I don't think there's a lock on the GPU assignment logic and I don't think we keep track of any state per node for the GPUs assigned.

luke-lombardi:
> Thinking about this more broadly and before these changes were introduced even, could we run into a race condition when running multiple workers on a multi-gpu node? Specifically with assigning GPUs to a container?
> 
> I don't think there's a lock on the GPU assignment logic and I don't think we keep track of any state per node for the GPUs assigned.

We actually do have a lock in memory for gpu assignment: https://github.com/beam-cloud/beta9/blob/main/pkg/worker/nvidia.go#L135

In terms of state / multi-gpu workers, the nvidia-device-plugin sets NVIDIA_VISIBLE_DEVICES, which we use to filter gpus out of the assignment pool at the worker level. That + the in-memory lock provides safety from the race conditions. There's a test for this here btw: https://github.com/beam-cloud/beta9/blob/main/pkg/worker/nvidia_test.go#L200

dleviminzi:
> Thinking about this more broadly and before these changes were introduced even, could we run into a race condition when running multiple workers on a multi-gpu node? Specifically with assigning GPUs to a container?
> 
> I don't think there's a lock on the GPU assignment logic and I don't think we keep track of any state per node for the GPUs assigned.

There is a lock on the `chooseDevices` call. That call also filters out devices that are not specifically assigned to the worker making the call. So on that front we should be safe.

pkg/worker/lifecycle.go
dleviminzi: Oh I'll change this.

luke-lombardi:
@nickpetrovic for reference here is the thing that 'filters' out GPUs that a worker should not assign (on multi-gpu nodes): https://github.com/beam-cloud/beta9/blob/main/pkg/worker/gpu_info.go#L83

pkg/worker/lifecycle.go
dleviminzi: This was moved while I was working on another issue. It turns out to not be a necessary move, so it can be moved back. 

That aside, I think we need to clean up this area a bit. Despite having a dedicated spec creating call, there are lots of parts of the spec that get set later in the spawn call. For instance the networking and root. Might be worth looping back and refactoring a bit at some point.

nickpetrovic:
> @nickpetrovic for reference here is the thing that 'filters' out GPUs that a worker should not assign (on multi-gpu nodes): https://github.com/beam-cloud/beta9/blob/main/pkg/worker/gpu_info.go#L83

Got it. This makes sense. Thanks for pointing all of this out. Also just a random realization, I didn't realize until now that the format of devices in `NVIDIA_VISIBLE_DEVICES` was `GPU-{uuid}` - would have thought this was an integer.

```
echo $NVIDIA_VISIBLE_DEVICES
GPU-de09ea7d-ae7f-793d-6fca-642ffae25025

nvidia-smi --query-gpu=pci.domain,pci.bus_id,index,uuid --format=csv,noheader,nounits
0x0000, 00000000:00:1E.0, 0, GPU-de09ea7d-ae7f-793d-6fca-642ffae25025
```

pkg/worker/lifecycle.go
nickpetrovic: Yeah lets move it back. And yeah I agree on the clean up. It's getting more chaotic as we add more functionality.

nickpetrovic (approved):
Going to approve this for now, but did add a few comments that should be addressed.

PR #960, opened 2025-02-14
feat: cleanup workers in external pools

nickpetrovic:
- Cleanup workers on external worker pools
- Delete k8s worker jobs when worker state no longer exists
- Rename worker cleanup config keys
- Move cleanup interval to config

Resolve BE-2343

### DIFF
--- a/pkg/scheduler/pool.go
+++ b/pkg/scheduler/pool.go
@@ -191,6 +191,7 @@ func cleanupWorkers(ctx context.Context, poolName string, workerConfig types.Wor
 		for _, pod := range pods.Items {
 			workerId, ok := pod.Labels[Beta9WorkerLabelIDKey]
 			if !ok {
+				log.Warn().Str("job_name", job.Name).Str("pod_name", pod.Name).Msg("worker id not found in pod labels")
 				continue
 			}
 
--- a/pkg/scheduler/pool.go
+++ b/pkg/scheduler/pool.go
@@ -13,6 +13,7 @@ import (
 	"github.com/beam-cloud/beta9/pkg/types"
 	"github.com/google/uuid"
 	"golang.org/x/sync/errgroup"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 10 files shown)


luke-lombardi (approved):
This looks good - I'm wondering if we should be tracking this somewhere (via alert, or event or something) so that we can alert when this is happening consistently.

pkg/scheduler/pool.go
```diff
@@ -159,3 +166,82 @@ func parseMemoryPercentage(percentStr string) (float32, error) {
...
+		fmt.Sprintf("%s=%s", Beta9WorkerLabelPoolNameKey, poolName),
+	}, ",")
+
+	jobs, err := kubeClient.BatchV1().Jobs(workerConfig.Namespace).List(ctx, metav1.ListOptions{LabelSelector: jobSelector})
+	if err != nil {
+		log.Error().Str("pool_name", poolName).Err(err).Msg("failed to list jobs for controller")
+		return
+	}
+
+	for _, job := range jobs.Items {
```
dleviminzi, line 182: Could we delete the job here instead with the delete propagation set to foreground? 

https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#DeleteOptions

nickpetrovic: Can you elaborate a little more? There's more logic after this that checks the pod's status and creation time which is important in knowing if we should delete the job.

nickpetrovic:
> This looks good - I'm wondering if we should be tracking this somewhere (via alert, or event or something) so that we can alert when this is happening consistently.

Good idea. I've been looking into this. I think I might have something where we can dynamically add metrics to victoria-metrics for any event. The value would always be 1, which _I think_ is the only value that makes sense for a time series database.

luke-lombardi:
> > This looks good - I'm wondering if we should be tracking this somewhere (via alert, or event or something) so that we can alert when this is happening consistently.
> 
> Good idea. I've been looking into this. I think I might have something where we can dynamically add metrics to victoria-metrics for any event. The value would always be 1, which _I think_ is the only value that makes sense for a time series database _and_ these event types.

So would that mean for each pending worker we would add a tick of 1? and then sum them on the grafana side?

pkg/scheduler/pool.go
dleviminzi: I think I misunderstood the loop here

nickpetrovic:
> > > This looks good - I'm wondering if we should be tracking this somewhere (via alert, or event or something) so that we can alert when this is happening consistently.
> > 
> > 
> > Good idea. I've been looking into this. I think I might have something where we can dynamically add metrics to victoria-metrics for any event. The value would always be 1, which _I think_ is the only value that makes sense for a time series database _and_ these event types.
> 
> So would that mean for each pending worker we would add a tick of 1? and then sum them on the grafana side? Also would want to tack on info like pool name, etc ofc

Yep exactly. I may have found a way to do dynamically add labels via a lua script. These are the fields I was thinking of adding for the worker lifecycle

```go
type EventWorkerLifecycleSchema struct {
	WorkerID  string              `json:"worker_id"`
	MachineID string              `json:"machine_id"`
	Status    string              `json:"status"`
	PoolName  string              `json:"pool_name"`
	NodeName  string              `json:"node_name"`
	Reason    DeletedWorkerReason `json:"reason"`
}

type DeletedWorkerReason string


const (
	DeletedWorkerReasonNonExistantState        DeletedWorkerReason = "non_existant_state"
	DeletedWorkerReasonExceededPendingAgeLimit DeletedWorkerReason = "exceeded_pending_age_limit"
)
```

pkg/scheduler/pool.go
```diff
@@ -159,3 +165,117 @@ func parseMemoryPercentage(percentStr string) (float32, error) {
...
+		})
+		if err != nil {
+			continue
+		}
+
+		if len(podList.Items) > 0 {
+			continue
+		}
+
+		if worker.Status == types.WorkerStatusAvailable {
```
luke-lombardi, line 203: With this check, we're basically saying -- is there state that doesn't show a corresponding pod right? I was sort of imagining we'd look at the _pods_ themselves, and then check if there's corresponding worker state.

nickpetrovic: That's being done in the Pods().List() call above. We're using a label selector that looks for exactly this worker id. If we find a pod, then we're good - don't need to delete state. If we don't find a pod, then we should delete the state. 

I added the worker.Status check because I found that state could be deleted right after they were created by the scheduler. I was hoping to use a timestamp for this check instead e.g. older than 5 mins, but I didn't see one.

nickpetrovic: Pushed up new changes. I still have to test this, but take a look when you can.

luke-lombardi: Looks good to me, did you have a chance to test this out yet?

nickpetrovic: Just did - we'll have to manually delete any current jobs that are stuck since this adds an additional annotation to filter the job/pod results, but other than that, it seems good to me.

PR #962, opened 2025-02-19
feat: Use gRPC Health Check server

nickpetrovic:
**Changes**

- Replaces HTTP health check with gRPC
- Enables gRPC reflection when in debug mode
- Configures aggressive readiness and liveness probes

**Summary**

This adds a new [gRPC health](https://grpc.io/docs/guides/health-checking/) endpoint. 

The new readiness probe values basically say the pod will _stop receiving traffic_ if the gRPC health endpoint fails for 6s (2s * 3s). It also runs every 3 seconds. Alternatively, this starts sending the pod traffic after 6s of successful checks.

The new liveness probe values basically say the pod will _restart_ if the gRPC health endpoint fails for 30s (10s * 3s). It also runs every 3 seconds. Alternatively, this will not restart the pod after 6s of successful checks.

Because the liveness checks take no more than 30s, the shutdown timeout will no longer be relevant when it is larger than the liveness check's failure threshold.

Resolve BE-2386

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -16,7 +16,6 @@ import (
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
 	"golang.org/x/sync/errgroup"
-	"google.golang.org/grpc"
 
 	"github.com/beam-cloud/beta9/pkg/abstractions/container"
 	"github.com/beam-cloud/beta9/pkg/abstractions/endpoint"
@@ -46,6 +45,11 @@ import (
 	"github.com/beam-cloud/beta9/pkg/task"
 	"github.com/beam-cloud/beta9/pkg/types"
 	pb "github.com/beam-cloud/beta9/proto"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/reflection"
 )
 
 type Gateway struct {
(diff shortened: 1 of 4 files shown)


pkg/gateway/gateway.go
```diff
@@ -506,6 +520,8 @@ func (g *Gateway) Start() error {
 // Shutdown gracefully shuts down the gateway.
 // This function is blocking and will only return when the gateway has been shut down.
 func (g *Gateway) shutdown() {
+	g.healthServer.Shutdown()
```
luke-lombardi, line 523: curious why we need to explicitly shut this down? we can't pass a context into it somehow?

luke-lombardi (approved):

pkg/gateway/gateway.go
nickpetrovic: Yeah possibly in a goroutine that waits on the global context. Is the issue having this on the Gateway struct?

dleviminzi (approved):

pkg/gateway/gateway.go
luke-lombardi: nah I don't know just feels like it should be handled using the global context like the rest of the services. also it's weird that its called "Server" but not sure what we can do about that since it's a built-in

nickpetrovic: Just updated this to call .Shutdown() when the global context is cancelled. This is probably a little cleaner.

luke-lombardi (approved):

pkg/gateway/gateway.go
luke-lombardi: Yeah lgtm

PR #970, opened 2025-02-20
Feat: beam pod

luke-lombardi:
This adds the ability to run generic (not just python based) services in beta9. For example, you can run a golang webserver. For now, it's restricted to HTTP based applications (or anything that runs over HTTP, like websockets). 

The model for deployment is mostly the same as the rest of the platform, however deployments are by default unable to scale to zero - for now. There are two modes of operation:

1. Create a `Pod` programmatically:

app.py
```
from beta9 import Pod

pod = Pod(
    ports=[8080],
    entrypoint=["sh", "-c", "python -m http.server 8080"],
    secrets=["MY_SECRET"],
)
pod.create()
```

2. Via CLI:
app.py
```
from beta9 import Pod

pod = Pod(
    ports=[8080],
    entrypoint=["sh", "-c", "python -m http.server 8080"],
    secrets=["MY_SECRET"],
)
```

- To run a one-off container: `beta9 run app.py:pod`
- To deploy a stateful, scalable set of containers (load balanced behind a common name): `beta9 deploy app.py:pod --name some-pod`


pkg/abstractions/image/build.go
```diff
@@ -385,10 +388,12 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
 			return err
 		}
 		setupCommands = append(setupCommands, installCmd)
+	} else if opts.IgnorePython {
+		opts.PythonVersion = ""
 	}
 
 	// Generate the pip install command and prepend it to the commands list
-	if len(opts.PythonPackages) > 0 {
+	if len(opts.PythonPackages) > 0 && opts.PythonVersion != "" {
```
dleviminzi, line 396: Should we throw some kind of error if there are python packages but no python version?

pkg/abstractions/image/build.go
```diff
@@ -364,9 +366,10 @@ func (b *Builder) Build(ctx context.Context, opts *BuildOpts, outputChan chan co
 	}
 
 	var setupCommands []string
+
 	// Detect if python3.x is installed in the container, if not install it
 	checkPythonVersionCmd := fmt.Sprintf("%s --version", opts.PythonVersion)
-	if resp, err := client.Exec(containerId, checkPythonVersionCmd); (err != nil || !resp.Ok) && !micromambaEnv {
+	if resp, err := client.Exec(containerId, checkPythonVersionCmd); (err != nil || !resp.Ok) && !micromambaEnv && !opts.IgnorePython {
```
dleviminzi, line 372: I hate the nesting here as it is, but I think it might make more sense to check `opts.IgnorePython` at an out level. Otherwise we still execute the command and the user will see the result and potentially wonder why/what is happening.

PR #975, opened 2025-02-24
Feat: attach to running container

luke-lombardi:
- Adds a command to connect to a running container and see the logs (up to that point)

Screenshot 2025-02-24 at 1 47 12 PM

Only issue with this pattern is that we flush the logs in the runc servers buffer, which only captures whatever logs are there -- not historical logs.


jsun-m (approved):
lgtm

dleviminzi (approved):

PR #977, opened 2025-02-24
Feat: add deployment id to container list if available

luke-lombardi:
Requires an extra query for each container, but for CLI usage it should be fine. Better would be to cache this ID in state somewhere to avoid the extra queries. Querying by stub ID should be safe to look up the deployment since we `force_create_stub` when preparing the runtime for any new deployment -- so it should be one-to-one


sdk/src/beta9/cli/container.py
```diff
@@ -31,10 +32,12 @@ def management():
 
 
 AVAILABLE_LIST_COLUMNS = {
-    "container_id": "ID",
+    "container_id": "Id",
```
jsun-m, line 35: I think this should be consistent with the other id's names

sdk/src/beta9/cli/container.py
```diff
@@ -56,64 +59,66 @@ def management():
...
     """,
 )
 @extraclick.pass_service_client
 @click.pass_context
 def list_containers(ctx: click.Context, service: ServiceClient, format: str, columns: str):
     res = service.gateway.list_containers(ListContainersRequest())
-
     if not res.ok:
         terminal.error(res.error_msg)
+        return
```
jsun-m, line 74: this should already exit from the terminal error

pkg/gateway/services/container.go
```diff
@@ -88,29 +95,28 @@ func (gws GatewayService) getContainersAsAdmin() ([]types.ContainerState, map[st
...
 		}, nil
 	}
 
 	err = gws.scheduler.Stop(&types.StopContainerArgs{ContainerId: in.ContainerId})
 	if err != nil {
 		return &pb.StopContainerResponse{
 			Ok:       false,
-			ErrorMsg: "Unable to stop container",
+			ErrorMsg: fmt.Sprintf("Unable to stop container: %s", in.ContainerId),
 		}, nil
```
dleviminzi, line 120: Should we log the error here since we're not returning it? It looks like the actual `Stop` method itself only logs the error in 1/3 cases.

sdk/src/beta9/cli/container.py
```diff
@@ -31,10 +32,12 @@ def management():
 
 
 AVAILABLE_LIST_COLUMNS = {
-    "container_id": "ID",
+    "container_id": "Id",
     "status": "Status",
-    "stub_id": "Stub Id",
+    "stub_id": "Stub ID",
```
dleviminzi, line 37: changed to `ID` in one place and `Id` in the other.

dleviminzi (approved):

jsun-m (changes requested):

pkg/gateway/services/container.go
luke-lombardi: Updated

sdk/src/beta9/cli/container.py
luke-lombardi: Updated

pkg/gateway/services/container.go
```diff
@@ -36,18 +36,25 @@ func (gws GatewayService) ListContainers(ctx context.Context, in *pb.ListContain
...
-			WorkspaceId: state.WorkspaceId,
-			Status:      string(state.Status),
-			ScheduledAt: timestamppb.New(time.Unix(state.ScheduledAt, 0)),
-			StartedAt:   timestamppb.New(time.Unix(state.StartedAt, 0)),
-			WorkerId:    containerWorkerMap[state.ContainerId].WorkerId,
-			MachineId:   containerWorkerMap[state.ContainerId].MachineId,
+	containers := []*pb.Container{}
+	for _, state := range containerStates {
+		deploymentId := ""
+		deployment, err := gws.backendRepo.GetDeploymentByStubExternalId(ctx, authInfo.Workspace.Id, state.StubId)
```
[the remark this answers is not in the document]
luke-lombardi: I think I'll handle this in a follow up - I didn't want to have to map them to the container states after doing the query

sdk/src/beta9/cli/container.py
luke-lombardi: Done

jsun-m (approved):

PR #978, opened 2025-02-24
feat: lock during worker pool cleanup

nickpetrovic:
- Only run one cleanup per pool at a time
- Move pool cleaner logic to its own file

Resolve BE-2343


pkg/scheduler/pool_cleaner.go
nickpetrovic, line 1: No logic changed here. Just moved to a new file.

dleviminzi (approved):

PR #990, opened 2025-02-26
fix: reduce timeouts

nickpetrovic:
- Lower connect timeout for checkAddressIsReady in buffer.go from 175s to 2s
- Implement Tailscale.DialTimeout and modify Tailscale.Dial so it does not have a timeout, then adjust code to use these two functions

### DIFF
--- a/pkg/abstractions/endpoint/buffer.go
+++ b/pkg/abstractions/endpoint/buffer.go
@@ -350,7 +350,8 @@ func (rb *RequestBuffer) getHttpClient(address string) (*http.Client, error) {
 		return rb.httpClient, nil
 	}
 
-	conn, err := network.ConnectToHost(rb.ctx, address, types.RequestTimeoutDurationS, rb.tailscale, rb.tsConfig)
+	ctx := context.WithValue(rb.ctx, "caller", "RequestBuffer.getHttpClient")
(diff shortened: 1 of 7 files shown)


pkg/abstractions/endpoint/buffer.go
```diff
@@ -444,7 +446,7 @@ func (rb *RequestBuffer) handleHttpRequest(req *request, c container) {
 		requestBody = io.NopCloser(bytes.NewReader(payloadBytes))
 	}
 
-	httpClient, err := rb.getHttpClient(c.address)
+	httpClient, err := rb.getHttpClient(c.address, handleHttpRequestClientTimeout)
```
luke-lombardi, line 449: Will this constrain the request duration to handleHttpRequestClientTimeout? Worried about long running requests

luke-lombardi (changes requested):

pkg/abstractions/endpoint/buffer.go
nickpetrovic: There's a lot going on within how tailscale dial works, so I've restored this timeout to its original value of 175s.

luke-lombardi (approved):

pkg/abstractions/endpoint/buffer.go
luke-lombardi: Ok - approved, will need to resolve conflicts as I made another change to this file (unrelated to timeouts).

PR #1001, opened 2025-02-28
Add instrumentation

dleviminzi:
Resolve BE-2377


pkg/abstractions/endpoint/buffer.go
```diff
@@ -353,10 +354,13 @@ func (rb *RequestBuffer) getHttpClient(address string, timeout time.Duration) (*
 		return rb.httpClient, nil
 	}
 
+	start := time.Now()
 	conn, err := network.ConnectToHost(rb.ctx, address, timeout, rb.tailscale, rb.tsConfig)
 	if err != nil {
 		return nil, err
 	}
+	// Use goroutine to avoid potential blocking from rate limiter's internal mutex
+	go metrics.RecordDialTime(time.Since(start))
```
luke-lombardi, line 363: a bit confused on this, why would it block? the rate limiter doesn't just return if its not "allowed". I could see that other weird issues

pkg/gateway/gateway.go
```diff
@@ -68,7 +68,7 @@ type Gateway struct {
 	WorkerPoolRepo repository.WorkerPoolRepository
 	EventRepo      repository.EventRepository
 	Tailscale      *network.Tailscale
-	metricsRepo    repository.MetricsRepository
+	metricsRepo    repository.UsageRepository
```
luke-lombardi, line 71: should this be called usageRepo here?

pkg/scheduler/scheduler_test.go
```diff
@@ -66,7 +66,7 @@ func NewSchedulerForTest() (*Scheduler, error) {
 		workerPoolManager: workerPoolManager,
 		requestBacklog:    requestBacklog,
 		containerRepo:     containerRepo,
-		schedulerMetrics:  schedulerMetrics,
+		schedulerUsage:    schedulerMetrics,
```
luke-lombardi, line 69: names should match

pkg/worker/image.go
```diff
@@ -138,8 +142,10 @@ func (c *ImageClient) PullLazy(ctx context.Context, request *types.ContainerRequ
 			} else {
 				c.logger.Log(request.ContainerId, request.StubId, "unable to cache image nearby <%s>: %v\n", imageId, err)
 			}
+			metrics.RecordImagePullTime(time.Since(pullStartTime))
 		}
 	}
+	log.Info().Str("local_cache_path", localCachePath).Msg("local cache path")
```
luke-lombardi, line 148: do we need this log anymore? this was for debugging right

pkg/worker/image.go
```diff
@@ -293,7 +300,14 @@ func (c *ImageClient) BuildAndArchiveImage(ctx context.Context, outputLogger *sl
 	if err != nil {
 		return err
 	}
+	ociImageInfo, err := os.Stat(ociPath)
+	if err != nil {
+		log.Warn().Err(err).Msg("unable to inspect image size")
```
luke-lombardi, line 305: if this fails, won't the following code not work? ```float64(ociImageInfo.Size()) / 1024 / 1024``` or will it just report 0MB?

pkg/worker/image.go
```diff
@@ -318,7 +332,14 @@ func (c *ImageClient) BuildAndArchiveImage(ctx context.Context, outputLogger *sl
 		}
 	}
 
-	err = c.Archive(ctx, tmpBundlePath, request.ImageId, nil)
+	bundleInfo, err := os.Stat(tmpBundlePath)
+	if err != nil {
+		log.Warn().Err(err).Msg("unable to inspect image size")
+	}
+	bundleMB := float64(bundleInfo.Size()) / 1024 / 1024
+	metrics.RecordImageUnpackSpeed(bundleMB, time.Since(startTime))
```
luke-lombardi, line 340: same comment as above

pkg/worker/image.go
```diff
@@ -345,30 +366,43 @@ func (c *ImageClient) PullAndArchiveImage(ctx context.Context, outputLogger *slo
...
 	err = c.skopeoClient.Copy(ctx, *request.BuildOptions.SourceImage, dest, request.BuildOptions.SourceImageCreds)
 	if err != nil {
 		return err
 	}
+	metrics.RecordImageCopySpeed(imageMB, time.Since(startTime))
 
 	outputLogger.Info("Unpacking image...\n")
 	tmpBundlePath := filepath.Join(baseTmpBundlePath, request.ImageId)
-	err = c.unpack(ctx, baseImage.Repo, baseImage.Tag, tmpBundlePath)
+	err = c.unpack(ctx, baseImage.Repo, baseImage.Tag, tmpBundlePath, imageMB)
```
luke-lombardi, line 385: maybe call this imageSizeMB

pkg/worker/image.go
```diff
@@ -345,30 +366,43 @@ func (c *ImageClient) PullAndArchiveImage(ctx context.Context, outputLogger *slo
...
+	err = c.Archive(ctx, tmpBundlePath, request.ImageId, nil, imageMB)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
 
-func (c *ImageClient) unpack(ctx context.Context, baseImageName string, baseImageTag string, bundlePath string) error {
+func (c *ImageClient) unpack(ctx context.Context, baseImageName string, baseImageTag string, bundlePath string, imageMB float64) error {
```
luke-lombardi, line 401: same comment as above. although its kinda weird that we should even have to pass this in here, feels wrong.

luke-lombardi: maybe would push this out

pkg/worker/worker.go
```diff
@@ -56,7 +56,7 @@ type Worker struct {
 	containerLock           sync.Mutex
 	containerWg             sync.WaitGroup
 	containerLogger         *ContainerLogger
-	workerMetrics           *WorkerMetrics
+	workerMetrics           *WorkerUsage
```
luke-lombardi, line 59: workerUsageMetrics? or workerUsage? I think in general all of these places where we have the repo called usage etc, it might still make sense to call it UsageMetrics. Usage alone isn't really clear imo

luke-lombardi (changes requested):

pkg/abstractions/endpoint/buffer.go
dleviminzi: There is a mutex inside the rate limiter to acquire the count. Its likely a small and quick acquisition, but it could still block

luke-lombardi, line 363: a little concerned about spinning up a goroutine for every single request just for this data point.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	metricImageArchiveSpeed         = "worker_image_archive_speed_mbps"
+	metricImagePushSpeed            = "worker_image_push_speed_mbps"
+	metricS3PutSpeed                = "s3_put_speed_mbps"
+	metricS3GetSpeed                = "s3_get_speed_mbps"
+	metricDialTime                  = "dial_time_ms"
+	metricContainerStartLatency     = "container_start_latency_ms"
+)
+
+func InitializeMetricsRepository(config types.VictoriaMetricsConfig) {
+	// ENV to collect for default labels
```
luke-lombardi, line 35: this comment isn't that clear. I would probably just remove it

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	vmetrics.GetDefaultSet().InitPushWithOptions(
+		context.Background(),
+		config.PushURL,
+		time.Duration(config.PushSecs)*time.Second,
+		opts,
+	)
+
+	dialMetricLimiter = rate.NewLimiter(2, 1)
```
luke-lombardi, line 61: what are these args? maybe use consts here so its obvious what they do?

pkg/registry/registry.go
```diff
@@ -151,6 +164,14 @@ func (s *S3Store) Get(ctx context.Context, key string, localPath string) error {
 		return err
 	}
 
+	info, err := os.Stat(localPath)
+	if err != nil {
```
luke-lombardi, line 168: were we not handling this error before? or was it handled a level up?

pkg/scheduler/scheduler.go
```diff
@@ -120,7 +121,7 @@ func (s *Scheduler) Run(request *types.ContainerRequest) error {
 		}
 	}
 
-	go s.schedulerMetrics.CounterIncContainerRequested(request)
+	go s.schedulerUsage.CounterIncContainerRequested(request)
```
luke-lombardi, line 124: should this also be schedulerUsageMetrics to align with workerUsageMetrics?

pkg/scheduler/scheduler.go
```diff
@@ -298,7 +299,12 @@ func (s *Scheduler) StartProcessingRequests() {
 		err = s.scheduleRequest(worker, request)
 		if err != nil {
 			s.addRequestToBacklog(request)
+			continue
```
luke-lombardi, line 302: is this continue safe here?

pkg/scheduler/scheduler.go
```diff
@@ -465,6 +471,8 @@ func (s *Scheduler) addRequestToBacklog(request *types.ContainerRequest) error {
 		return s.requestBacklog.Push(request)
 	}
 
+	go metrics.RecordRequestRetry(request)
```
luke-lombardi, line 474: in general i feel like the goroutines, if we're using them, should be behind the interface in the implementation.

luke-lombardi (changes requested):

pkg/scheduler/scheduler.go
dleviminzi: I think so. Previously it would just continue in the outer loop. However, with the added metrics below, I tweaked it to `continue` if `scheduleRequest fails. The only reason I did this was to avoid emitting the metrics when scheduling the request fails.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,148 @@
...
+
+	vmetrics "github.com/VictoriaMetrics/metrics"
+	"github.com/beam-cloud/beta9/pkg/types"
+	"github.com/rs/zerolog/log"
+	"tailscale.com/tstime/rate"
+)
+
+const (
+	// Rate limiter that allows 2 records per second
+	dialMetricLimiterRate  = 2
```
luke-lombardi, line 17: is this across all invocations?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func RecordS3GetSpeed(sizeInMB float64, duration time.Duration) {
+	if sizeInMB > 0 {
+		vmetrics.GetDefaultSet().GetOrCreateHistogram(metricS3GetSpeed).Update(sizeInMB / duration.Seconds())
+	}
+}
+
+func RecordDialTime(duration time.Duration) {
+	if dialMetricLimiter.Allow() {
```
luke-lombardi, line 141: if we're using the goroutines, should push in here i.e. :

```
func RecordDialTime(duration time.Duration) {
	if dialMetricLimiter.Allow() {
	go func() {
vmetrics.GetDefaultSet().GetOrCreateHistogram(metricDialTime).Update(float64(duration.Milliseconds())) } ()
	}
}

File: pkg/scheduler/scheduler_test.go [hunk 18]
```diff
@@ -45,8 +45,8 @@ func NewSchedulerForTest() (*Scheduler, error) {
 	config := configManager.GetConfig()
 	eventRepo := repo.NewTCPEventClientRepo(config.Monitoring.FluentBit.Events)
 
-	schedulerMetrics := SchedulerMetrics{
-		metricsRepo: nil,
+	schedulerUsage := SchedulerUsageMetrics{
```
luke-lombardi, line 48: name should probably match here `schedulerUsageMetrics`

pkg/worker/image.go
```diff
@@ -334,30 +368,43 @@ func (c *ImageClient) PullAndArchiveImage(ctx context.Context, outputLogger *slo
 
 	dest := fmt.Sprintf("oci:%s:%s", baseImage.Repo, baseImage.Tag)
 
-	outputLogger.Info("Pulling image...\n")
+	imageBytes, err := c.skopeoClient.InspectSizeInBytes(ctx, *request.BuildOptions.SourceImage, request.BuildOptions.SourceImageCreds)
```
luke-lombardi, line 371: how does this handle 0 values?

luke-lombardi (changes requested):

pkg/worker/image.go
dleviminzi: If the size is 0 then the calculated image size will end up being zero and it will be skipped inside the metrics repo call.

pkg/abstractions/endpoint/buffer.go
```diff
@@ -353,10 +354,13 @@ func (rb *RequestBuffer) getHttpClient(address string, timeout time.Duration) (*
 		return rb.httpClient, nil
 	}
 
+	start := time.Now()
 	conn, err := network.ConnectToHost(rb.ctx, address, timeout, rb.tailscale, rb.tsConfig)
 	if err != nil {
 		return nil, err
 	}
+	// Use goroutine to avoid potential blocking from rate limiter's internal mutex
```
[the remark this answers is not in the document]
dleviminzi: removed

pkg/abstractions/pod/proxy.go
```diff
@@ -178,11 +179,13 @@ func (pb *PodProxyBuffer) handleConnection(conn *connection) {
 		return
 	}
 
+	start := time.Now()
 	containerConn, err := network.ConnectToHost(request.Context(), container.addressMap[int32(port)], containerDialTimeoutDurationS, pb.tailscale, pb.tsConfig)
 	if err != nil {
 		conn.ctx.String(http.StatusServiceUnavailable, "Failed to connect to service")
 		return
 	}
+	go metrics.RecordDialTime(time.Since(start))
```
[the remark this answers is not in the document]
dleviminzi: removed

pkg/metrics/metrics.go
dleviminzi: only the gateway dials things, so this basically just means that the Gateway will record metrics for at most two dials each second

luke-lombardi (approved):

PR #1004, opened 2025-03-03
Dont limit build logs

dleviminzi:
Resolve BE-2427

The current log limits are imposed to build containers, which might prevent users from seeing critical build logs.

### DIFF
--- a/pkg/worker/logger.go
+++ b/pkg/worker/logger.go
@@ -14,6 +14,7 @@ import (
 	"golang.org/x/time/rate"
 
 	"github.com/beam-cloud/beta9/pkg/common"
+	types "github.com/beam-cloud/beta9/pkg/types"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/worker/logger.go
```diff
@@ -59,7 +59,7 @@ func (r *ContainerLogger) Log(containerId, stubId string, format string, args ..
 	return nil
 }
 
-func (r *ContainerLogger) CaptureLogs(containerId string, logChan chan common.LogRecord) error {
+func (r *ContainerLogger) CaptureLogs(containerId string, logChan chan common.LogRecord, isBuildRequest bool) error {
```
luke-lombardi, line 62: should we just pass the entire container request in here?

luke-lombardi (changes requested):

pkg/worker/logger.go
dleviminzi: done

luke-lombardi (approved):

PR #1006, opened 2025-03-03
feat: use default docker creds if they are configured

nickpetrovic:
Resolve BE-2430

### DIFF
--- a/pkg/abstractions/image/build.go
+++ b/pkg/abstractions/image/build.go
@@ -37,6 +37,8 @@ const (
 	pipCommandType        string = "pip"
 	shellCommandType      string = "shell"
 	micromambaCommandType string = "micromamba"
+
+	dockerHubRegistry string = "docker.io"
 )
 
 type Builder struct {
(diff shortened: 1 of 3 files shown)


pkg/abstractions/image/build.go
```diff
@@ -222,6 +223,10 @@ func (i *BaseImage) String() string {
 	return fmt.Sprintf("%s/%s:%s", i.Registry, i.Repo, i.Tag)
 }
 
+func (b *Builder) shouldUseDefaultImageCreds(opts *BuildOpts) bool {
+	return opts.BaseImageCreds == "" && opts.DefaultImageCreds.Username != "" && opts.DefaultImageCreds.Password != ""
+}
```
nickpetrovic, line 228: TODO: fix so its for docker hub only

pkg/abstractions/image/build.go
```diff
@@ -222,6 +225,14 @@ func (i *BaseImage) String() string {
 	return fmt.Sprintf("%s/%s:%s", i.Registry, i.Repo, i.Tag)
 }
 
+func (b *Builder) shouldUseDefaultImageCreds(opts *BuildOpts) bool {
+	// TODO: make dockerHubRegistry configurable
```
luke-lombardi, line 229: what might this look like in the config?

nickpetrovic: I was thinking we might need to allow overriding of the dockerHubRegistry constant, but I don't think so anymore. Just removed the TODO.

luke-lombardi: kk sounds good

luke-lombardi (approved):

PR #1013, opened 2025-03-06
Feat: Optionally emit container cost from worker

dleviminzi:


pkg/clients/container_cost.go
```diff
@@ -0,0 +1,62 @@
...
+
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return 0, err
+	}
+	defer resp.Body.Close()
+
+	var response ContainerCostResponse
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return 0, fmt.Errorf("failed to decode response: %v", err)
```
luke-lombardi, line 53: do we really need a custom error here? i would just return the original error

pkg/worker/lifecycle.go
```diff
@@ -649,6 +649,8 @@ func (s *Worker) spawn(request *types.ContainerRequest, spec *specs.Spec, output
 	}
 
 	// Log metrics
+
+	log.Info().Str("container_id", request.ContainerId).Str("cost_per_ms", fmt.Sprintf("%f", request.CostPerMs)).Msg("container cost per ms")
```
luke-lombardi, line 653: should this log be in the client or something? what happens if you don't configure the hook? is it zero here?

pkg/worker/metrics.go
```diff
@@ -47,21 +53,55 @@ func (wm *WorkerMetrics) metricsContainerDuration(request *types.ContainerReques
...
 	for {
 		select {
 		case <-ticker.C:
 			go wm.metricsContainerDuration(request, time.Since(cursorTime))
+			go wm.metricsContainerCost(request, time.Since(cursorTime))
 			cursorTime = time.Now()
 		case <-ctx.Done():
 			// Consolidate any remaining time
 			go wm.metricsContainerDuration(request, time.Since(cursorTime))
+			go wm.metricsContainerCost(request, time.Since(cursorTime))
```
luke-lombardi, line 88: does this mean we emit from two separate goroutines? can't we consolidate it into one if so?

luke-lombardi (changes requested):

pkg/worker/metrics.go
dleviminzi: This is actually already inside a go routine so I removed the goroutine creation for both, since I don't see why we need them.

luke-lombardi: Well, I guess one argument could so that you don't miss data points right? if one of these takes a long time, it could slowly lead to drift / missed data no?

luke-lombardi (approved):

PR #1019, opened 2025-03-10
Write restore logs to tmp

dleviminzi:
* Changes the base working directory for restores to `/tmp` to avoid write issues with the cache/mountpoint-s3 filesystems.


pkg/worker/criu_nvidia.go
```diff
@@ -180,6 +186,7 @@ func (c *NvidiaCRIUManager) cacheDir(containerId, checkpointPath string) error {
 	}))
 
 	wg.Wait() // Wait for all tasks to finish
+	log.Info().Str("container_id", containerId).Msgf("cached checkpoint: %s", checkpointPath)
```
luke-lombardi, line 189: won't this log success even if it errors out?

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1020, opened 2025-03-10
Feat: Storage refactor

luke-lombardi:
Modifies storage architecture to move away from a single JFS mount and towards a WorkspaceStorage model. There are a few pieces still missing from this, but I've opened it up for review. 

**TODO**
- [x] Migrate output stat / creation to work in both the workspace storage model and legacy storage model
- [x] Modify volume services to allow listing via StorageClient instead of direct mounts (conditionally, if workspace storage is enabled)
- [ ] Fix symlinks in geesefs
- [x] Clean up / ensure stub object upload works for legacy workspaces as well as new workspaces. This may require a couple of small changes
- [x] Implement geeseFS unmount logic
- [x] Implement fuse mount cleanup by workspace (opportunistically, after a grace period) in worker. It is currently done naively on worker shutdown
- [x] Review config changes and flatten anything that can be flattened

---

These need to be done in a follow up:
- [ ] Automatically create buckets for users on workspace creation
- [ ] Prepare migration script to create workspace storage objects and migrate data


pkg/storage/geese.go
```diff
@@ -0,0 +1,140 @@
...
+
+	log.Info().Str("local_path", localPath).Msg("geesefs filesystem mounted")
+	return nil
+}
+
+func (s *GeeseStorage) Format(fsName string) error {
+	return nil
+}
+
+func (s *GeeseStorage) Unmount(localPath string) error {
```
jsun-m, line 114: Is this part incomplete or does it umount automatically without issues?

pkg/storage/manager.go
```diff
@@ -0,0 +1,121 @@
...
+
+func NewStorageManager(config types.StorageConfig) (*StorageManager, error) {
+	return &StorageManager{
+		mounts: common.NewSafeMap[Storage](),
+		config: config,
+		mu:     sync.Mutex{},
+	}, nil
+}
+
+func (s *StorageManager) Create(workspaceName string, storage Storage) {
```
jsun-m, line 26: Is this used anywhere?

pkg/types/config.go
```diff
@@ -182,13 +183,20 @@ type PythonStandaloneConfig struct {
...
-	ObjectPath     string           `key:"objectPath" json:"object_path"`
-	JuiceFS        JuiceFSConfig    `key:"juicefs" json:"juicefs"`
-	CunoFS         CunoFSConfig     `key:"cunofs" json:"cunofs"`
-	MountPoint     MountPointConfig `key:"mountpoint" json:"mountpoint"`
+	Mode             string                 `key:"mode" json:"mode"`
+	FilesystemName   string                 `key:"fsName" json:"filesystem_name"`
+	FilesystemPath   string                 `key:"fsPath" json:"filesystem_path"`
+	ObjectPath       string                 `key:"objectPath" json:"object_path"`
+	JuiceFS          JuiceFSConfig          `key:"juicefs" json:"juicefs"`
+	Geese            GeeseConfig            `key:"geese" json:"geese"`
```
jsun-m, line 191: Should this be updated to GeeseFS for clarity and consistency? As well as the other places

pkg/worker/mount.go
```diff
@@ -14,36 +14,39 @@ import (
...
+
+			objectPath := request.Mounts[i].LocalPath
 			localUserSource := tempUserCodeDir(request.ContainerId)
-			err := copyDirectory(source, localUserSource)
-			if err != nil {
-				log.Error().Str("container_id", request.ContainerId).Err(err).Msg("failed to eagerly copy remote user code to local /mnt/code")
-			} else {
-				request.Mounts[i].LocalPath = localUserSource
+
+			if _, err := os.Stat(localUserSource); os.IsNotExist(err) {
```
jsun-m, line 35: We should handle the other errors aside from os.IsNotExist

pkg/worker/util.go
```diff
@@ -92,3 +95,49 @@ func (fl *FileLock) Release() error {
 	fl.file = nil
 	return nil
 }
+
+// extractZip extracts a zip file from src to dest
+func extractZip(src, dest string) error {
```
jsun-m, line 100: Can this be done with an extract library? https://github.com/saracen/fastzip

jsun-m (dismissed):

pkg/common/config.default.yaml
```diff
@@ -9,6 +9,7 @@ database:
     password: password
     name: main
     timezone: UTC
+    encryptionKey: sk_pKz38fK8v7lz01AneJI8MJnR70akmP2CtDNf1IufKcY=
```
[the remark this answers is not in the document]
luke-lombardi: yeah this is fine. This is just a placeholder (like `password`)

pkg/worker/util.go
luke-lombardi: yeah i don't think we need this. lots of shitty test code in this branch still

pkg/types/config.go
luke-lombardi: yeah probably, wll make that change

pkg/worker/mount.go
luke-lombardi: this will be removed once I merge the other branch with serve fixes

pkg/repository/backend_postgres.go
```diff
@@ -163,6 +164,29 @@ func (r *PostgresBackendRepository) GetWorkspaceByExternalId(ctx context.Context
 	return workspace, nil
 }
 
+func (r *PostgresBackendRepository) GetWorkspaceWithRelated(ctx context.Context, workspaceId uint) (types.WorkspaceWithRelated, error) {
+	var workspace types.WorkspaceWithRelated
+
+	query := `
+	SELECT w.id, w.name, w.created_at, w.concurrency_limit_id, w.volume_cache_enabled, w.multi_gpu_enabled,
```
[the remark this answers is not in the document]
luke-lombardi: Yeah agreed that this is getting really verbose

pkg/gateway/services/object.go
```diff
@@ -35,42 +34,47 @@ func (gws *GatewayService) HeadObject(ctx context.Context, in *pb.HeadObjectRequ
...
-		return &pb.PutObjectResponse{
+		return &pb.CreateObjectResponse{
 			Ok:       false,
 			ErrorMsg: "Unable to create object",
 		}, nil
 	}
 
-	filePath := path.Join(objectPath, newObject.ExternalId)
-	err = os.WriteFile(filePath, in.ObjectContent, 0644)
+	presignedURL, err := storageClient.GeneratePresignedURL(ctx, "objects/"+newObject.ExternalId, 60*60*24)
```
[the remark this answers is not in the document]
luke-lombardi: we use it on the client side, create object RPC returns the presigned url, then client uploads it

pkg/storage/geese.go
luke-lombardi: yeah incomplete

pkg/storage/manager.go
luke-lombardi: not yet, will create the bucket for the user

pkg/gateway/services/object.go
```diff
@@ -2,75 +2,103 @@ package gatewayservices
...
+				ObjectMetadata: &pb.ObjectMetadata{
+					Name: existingObject.Hash,
+					Size: existingObject.Size,
+				},
+				ObjectId: existingObject.ExternalId,
+			}, nil
+		} else {
+			err = gws.backendRepo.DeleteObjectByExternalId(ctx, existingObject.ExternalId)
+			if err != nil {
+				log.Error().Err(err).Msgf("Unable to delete object %s", existingObject.ExternalId)
```
luke-lombardi, line 50: This could instead mark the stub or object as invalid somehow.

pkg/clients/storage.go
```diff
@@ -0,0 +1,129 @@
...
+		Bucket: aws.String(*c.WorkspaceStorage.BucketName),
+		Prefix: aws.String(prefix),
+	})
+	if err != nil {
+		return nil, err
+	}
+	return resp.Contents, nil
+}
+
+func (c *StorageClient) GeneratePresignedURL(ctx context.Context, key string, expiresInSeconds int64) (string, error) {
```
dleviminzi, line 106: nitpick: maybe change name to presigned put url?

pkg/common/object.go
```diff
@@ -3,13 +3,11 @@ package common
...
 	"context"
 	"os"
-	"path"
 
-	"github.com/beam-cloud/beta9/pkg/types"
 	"github.com/mholt/archiver/v3"
 )
 
-func ExtractObjectFile(ctx context.Context, objectId, workspaceName, destPath string) error {
+func ExtractObjectFile(ctx context.Context, objectPath, destPath string) error {
```
dleviminzi, line 10: Are we still using this or was this just updated for backward compatibility for now?

pkg/gateway/services/object.go
dleviminzi, line 50: nitpick: maybe check if the err is not nil and do the clean up and return. That way the happy path can just continue after

pkg/gateway/services/object.go
```diff
@@ -2,75 +2,103 @@ package gatewayservices
...
 	objectPath := path.Join(types.DefaultObjectPath, authInfo.Workspace.Name)
 	os.MkdirAll(objectPath, 0644)
 
+	storageClient, err := clients.NewStorageClient(ctx, authInfo.Workspace.Name, authInfo.Workspace.Storage)
+	if err != nil {
+		return &pb.CreateObjectResponse{
+			Ok:       false,
+			ErrorMsg: "Unable to create storage client",
+		}, nil
+	}
```
dleviminzi, line 72: It would probably be good to some kind of client caching here. If a user is creating a bunch of objects at once we might end up with a bunch of unnecessary client creations/deletions. Could save the client with some TTL.

sdk/src/beta9/sync.py
```diff
@@ -181,33 +183,45 @@ def _sync(self) -> FileSyncResult:
...
         if not head_response.exists:
             metadata = ObjectMetadata(name=hash, size=size)
 
-            def stream_requests():
+            def _upload_object() -> requests.Response:
                 with terminal.progress_open(temp_zip_name, "rb", description=None) as file:
-                    while chunk := file.read(CHUNK_SIZE):
-                        yield PutObjectRequest(chunk, metadata, hash, False)
+                    response = requests.put(presigned_url, data=file)
+                return response
```
dleviminzi, line 193: Have you tested this with very large files?

pkg/types/config.go
```diff
@@ -207,12 +215,22 @@ type JuiceFSConfig struct {
 	BufferSize   int64  `key:"bufferSize" json:"buffer_size"`
 }
 
-type CunoFSConfig struct {
-	LicenseKey    string `key:"licenseKey" json:"license_key"`
-	S3AccessKey   string `key:"s3AccessKey" json:"s3_access_key"`
-	S3SecretKey   string `key:"s3SecretKey" json:"s3_secret_key"`
-	S3EndpointUrl string `key:"s3EndpointURL" json:"s3_endpoint_url"`
-	S3BucketName  string `key:"s3BucketName" json:"s3_bucket_name"`
```
dleviminzi, line 215: Since we're removing this, we should probably take it out of the dockerfile for the worker, gateway, etc.

pkg/common/object.go
luke-lombardi: It's being used, but mostly for backwards compatibility. We can probably simplify this once we have all users migrated off the old storage

pkg/clients/storage.go
luke-lombardi: done

pkg/types/config.go
luke-lombardi: Yeah I removed it already

pkg/abstractions/volume/multipart.go
```diff
@@ -224,10 +265,30 @@ func (s *GlobalVolumeService) CompleteMultipartUpload(ctx context.Context, in *p
 		}, nil
 	}
 
-	s3Client := s.getS3Client()
+	var s3Client *s3.Client
+
+	bucket := s.config.BucketName
```
luke-lombardi, line 270: This could be consolidated since we do the same thing a few places -- but not sure the func would actually be that much better

sdk/src/beta9/sync.py
luke-lombardi: not really, we can probably switch it to use multi-part in a follow up though

dleviminzi (approved):

PR #1022, opened 2025-03-11
Fix: Improve exit code logging

dleviminzi:
Container stopped by scheduler because keep warm expired.
Screenshot 2025-03-11 at 13 49 51

Stopping container with `beta9 container stop`
Screenshot 2025-03-11 at 13 50 14

In the future it would be nice to send even more information along with scheduler stops.

### DIFF
--- a/pkg/types/worker.go
+++ b/pkg/types/worker.go
@@ -88,7 +88,27 @@ const (
 	WorkerContainerExitCodeSuccess            = 0
 	WorkerContainerExitCodeOomKill            = 137 // 128 + 9 (base value + SIGKILL), used to indicate OOM kill
 	WorkerContainerExitCodeSigterm            = 143 // 128 + 15 (base value + SIGTERM), used to indicate a graceful termination
+	WorkerContainerExitCodeScheduler          = 558
+	WorkerContainerExitCodeTtl                = 559
+	WorkerContainerExitCodeUser               = 560
+	WorkerContainerExitCodeAdmin              = 561
 )
+
+const (
+	WorkerContainerExitCodeOomKillMessage   = "Container was killed due to an out-of-memory error"
+	WorkerContainerExitCodeSchedulerMessage = "Container was stopped by the scheduler"
+	WorkerContainerExitCodeTtlMessage       = "Container was stopped due to TTL expiration"
+	WorkerContainerExitCodeUserMessage      = "Container was stopped by the user"
+	WorkerContainerExitCodeAdminMessage     = "Container was stopped by the admin"
+)
+
+var ExitCodeMessages = map[int32]string{
+	WorkerContainerExitCodeOomKill:   WorkerContainerExitCodeOomKillMessage,
+	WorkerContainerExitCodeScheduler: WorkerContainerExitCodeSchedulerMessage,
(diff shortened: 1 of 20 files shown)


pkg/types/worker.go
```diff
@@ -88,8 +88,28 @@ const (
 	WorkerContainerExitCodeSuccess            = 0
 	WorkerContainerExitCodeOomKill            = 137 // 128 + 9 (base value + SIGKILL), used to indicate OOM kill
 	WorkerContainerExitCodeSigterm            = 143 // 128 + 15 (base value + SIGTERM), used to indicate a graceful termination
+	WorkerContainerExitCodeScheduler          = 558
```
luke-lombardi, line 91: Currently if we store a non-zero exit code, we consider that a "failed container". We may need to adjust [this](https://github.com/beam-cloud/beta9/blob/main/pkg/repository/container_redis.go#L408) logic to make sure that containers stopped through "normal" / non-failure means don't get marked by the rest of the system as failed containers.

pkg/worker/lifecycle.go
```diff
@@ -720,7 +735,12 @@ func (s *Worker) runContainer(ctx context.Context, request *types.ContainerReque
 		if err == nil {
 			return exitCode, containerId, err
 		}
-		log.Error().Str("container_id", request.ContainerId).Msgf("failed to run container with checkpoint/restore got exit code %d: %v, attempting to run container from bundle", exitCode, err)
+
+		if exitCode == types.WorkerContainerExitCodeOomKill {
```
luke-lombardi, line 739: not a huge fan of this nesting since it appears both here and below

pkg/worker/lifecycle.go
```diff
@@ -729,7 +749,11 @@ func (s *Worker) runContainer(ctx context.Context, request *types.ContainerReque
 		Started:      startedChan,
 	})
 	if err != nil {
-		log.Error().Str("container_id", request.ContainerId).Msgf("failed to run container: %v", err)
+		if exitCode == types.WorkerContainerExitCodeOomKill {
+			log.Warn().Str("container_id", request.ContainerId).Msg("container was killed")
```
luke-lombardi, line 753: referring to this above ^

pkg/types/worker.go
```diff
@@ -88,8 +88,28 @@ const (
...
+	WorkerContainerExitCodeTtl                = 559
+	WorkerContainerExitCodeUser               = 560
+	WorkerContainerExitCodeAdmin              = 561
 )
 
+const (
+	WorkerContainerExitCodeOomKillMessage   = "Container killed due to an out-of-memory error"
+	WorkerContainerExitCodeSchedulerMessage = "Container stopped by scheduler"
```
luke-lombardi, line 99: I don't think the user would necessarily understand this message "Container stopped by scheduler". We might need a different message - even though this is technically correct.

luke-lombardi (changes requested):

pkg/types/worker.go
dleviminzi: Ahh this makes sense. I will adjust the condition

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1024, opened 2025-03-11
Fix: Set container request status to failed if scheduler can't find a worker

dleviminzi:
Currently, users see something like this
```
error timeout reached while trying to get worker addr
```
This is confusing.

After this change they will see a message like the one below: 
Screenshot 2025-03-11 at 16 45 02

If the issue is unrelated to scheduling the default message is now 
```
failed to schedule container, container id:
```

### DIFF
--- a/pkg/repository/container_redis.go
+++ b/pkg/repository/container_redis.go
@@ -3,7 +3,6 @@ package repository
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


pkg/repository/container_redis.go
```diff
@@ -538,3 +544,16 @@ func (cr *ContainerRedisRepository) DeleteStubState(stubId string) error {
 	stateKey := common.RedisKeys.SchedulerStubState(stubId)
 	return cr.rdb.Del(context.TODO(), stateKey).Err()
 }
+
+func (cr *ContainerRedisRepository) SetContainerRequestStatus(containerId string, status types.ContainerRequestStatus) error {
```
luke-lombardi, line 548: these should probably have a TTL/expiry

luke-lombardi (changes requested):

pkg/repository/container_redis.go
```diff
@@ -294,14 +295,21 @@ func (cr *ContainerRedisRepository) GetWorkerAddress(ctx context.Context, contai
...
 			}
-			return "", errors.New("context cancelled while trying to get worker addr")
+			return "", fmt.Errorf("context cancelled while trying to get worker addr, container id: %s", containerId)
 		case <-ticker.C:
 			hostname, err = cr.rdb.Get(ctx, common.RedisKeys.SchedulerWorkerAddress(containerId)).Result()
 			if err == nil {
 				return hostname, nil
 			}
+
+			requestStatus, err := cr.GetContainerRequestStatus(containerId)
```
luke-lombardi, line 307: Instead of the err == nil check, and the subsequent if inside that block -- what if we make the signature of this similar to checking existence of a key in a map, i.e.:

```if requestStatus, exists := cr.GetContainerRequestStatus(containerId); exists {
           // do something with request status
 ```
 
 Can then handle the errors / check for unexpected errors internally.

File: pkg/repository/container_redis.go [hunk 3]
```diff
@@ -14,6 +13,8 @@ import (
 	redis "github.com/redis/go-redis/v9"
 )
 
+const containerRequestStatusTTL = 10 * time.Minute
```
luke-lombardi, line 16: this should be probably be in types next to the status type.

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1033, opened 2025-03-13
add field for requested gpus to prometheus metric

dleviminzi:


pkg/metrics/metrics.go
```diff
@@ -167,8 +175,12 @@ func RecordS3GetSpeed(sizeInMB float64, duration time.Duration) {
 
 func RecordDialTime(duration time.Duration, host string) {
 	if dialMetricLimiter.Allow() {
-		metricName := fmt.Sprintf("%s{host=\"%s\"}", metricDialTime, host)
-		vmetrics.GetDefaultSet().GetOrCreateHistogram(metricName).Update(float64(duration.Milliseconds()))
+		// Only want the machine-xxxxxxxx part of host (machine-xxxxxxxx.ttttttttt.ts.net:00000)
```
luke-lombardi, line 178: is this always gonna work? thinking of non-remote nodes etc. they don't always have tsnet in the hostname

luke-lombardi (changes requested):

pkg/metrics/metrics.go
dleviminzi: From what I'm seeing about whats being reported, this is always the format of the host name. In the case of the non-remote nodes I think its ok too because presumably we're not as interested in monitoring those.

luke-lombardi (approved):

PR #1038, opened 2025-03-14
Fix: any should not fail for serve

dleviminzi:
Below T4 is local spot and A10g is remote lambda node. 
Screenshot 2025-03-14 at 16 14 24

### DIFF
--- a/pkg/gateway/services/stub.go
+++ b/pkg/gateway/services/stub.go
@@ -126,49 +126,33 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
 			}, nil
 		}
 
-		gpuCounts, err := gws.workerRepo.GetFreeGpuCounts()
-		if err != nil {
-			return &pb.GetOrCreateStubResponse{
-				Ok:     false,
-				ErrMsg: "Failed to get GPU counts.",
-			}, nil
-		}
-
 		if types.StubType(in.GetStubType()).IsServe() {
-			gpuNames := []string{}
-			hasCapacity := false
-
-			if slices.Contains(gpus, types.GPU_ANY) {
-				hasCapacity = true
-			} else {
-				for _, gpu := range gpus {
-					if gpuCounts[gpu.String()] > 0 {
-						hasCapacity = true
-						break
-					}
-					gpuNames = append(gpuNames, gpu.String())
-				}
+			hasCapacity, err := gws.anyGpuAvailable(gpus)
+			if err != nil {
+				return &pb.GetOrCreateStubResponse{
+					Ok:     false,
+					ErrMsg: "Failed to check GPU availability.",
+				}, nil
 			}
 
 			if !hasCapacity {
 				return &pb.GetOrCreateStubResponse{
 					Ok:     false,
-					ErrMsg: fmt.Sprintf("There is currently no GPU capacity for %s.", strings.Join(gpuNames, ", ")),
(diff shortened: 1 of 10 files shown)


pkg/gateway/services/stub.go
```diff
@@ -137,12 +138,16 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
 			gpuNames := []string{}
 			hasCapacity := false
 
-			for _, gpu := range gpus {
-				if gpuCounts[gpu.String()] > 0 {
-					hasCapacity = true
-					break
+			if slices.Contains(gpus, types.GpuType("any")) {
```
luke-lombardi, line 141: isn't there a specific type for this? 


const (
	GPU_A10G    GPUType = "A10G"
	GPU_A100_40 GPUType = "A100-40"
	GPU_A100_80 GPUType = "A100-80"
	GPU_L4      GPUType = "L4"
	GPU_T4      GPUType = "T4"
	GPU_H100    GPUType = "H100"
	GPU_A6000   GPUType = "A6000"
	GPU_RTX4090 GPUType = "RTX4090"

	NO_GPU  GPUType = "NO_GPU"
	GPU_ANY GPUType = "any"
)

luke-lombardi (changes requested):

pkg/repository/worker_redis.go
```diff
@@ -259,6 +259,97 @@ func (r *WorkerRedisRepository) getWorkerFromKey(key string) (*types.Worker, err
...
+	for _, w := range workers {
+		if w.Gpu != "" {
+			gpuCounts[w.Gpu] += int(w.FreeGpuCount)
+		}
+	}
+
+	return gpuCounts, nil
+}
+
+func (r *WorkerRedisRepository) GetSpotGpus() []string {
```
luke-lombardi, line 304: I don't think we should expose this from the perspective of beta9. I mean I don't think beta9 should know of the concept of "spot'. We do have pre-emptable which maps to spot under the hood in most cases

pkg/gateway/services/stub.go
```diff
@@ -125,45 +126,33 @@ func (gws *GatewayService) GetOrCreateStub(ctx context.Context, in *pb.GetOrCrea
...
-			gpuNames := []string{}
-			hasCapacity := false
-
-			for _, gpu := range gpus {
-				if gpuCounts[gpu.String()] > 0 {
-					hasCapacity = true
-					break
-				}
-				gpuNames = append(gpuNames, gpu.String())
+			hasCapacity, err := gws.anyGpuAvailable(gpus)
```
luke-lombardi, line 130: Feels like we might add a bit of overhead with this, how does it feel?

luke-lombardi (changes requested):

pkg/gateway/services/stub.go
dleviminzi: In stage I couldn't feel any difference

pkg/repository/worker_redis.go
dleviminzi: I can switch to use the pre-emptable flag. That said, should beta9 not know when a pool is spot? I think it could be potentially useful from the user's perspective if we are able to report something like "count was zero, spot instance will be spun up so latency might be longer than usual"

luke-lombardi: yeah i mean maybe in the future but right now there's no concept of spot since node selector could be anyything. id use pre-emptible for now

luke-lombardi (approved):

PR #1046, opened 2025-03-19
Feat: Add ability to clone stubs

jsun-m:
**Summary**

This PR adds the ability to clone the stub only if it is public. The public field is necessary else user's can just clone any stub that exists.

I make sure to only clone information that can be shared. Secrets are repopulated from the user that is cloning the stub and will fail if the secrets do not exist. I also make sure to clone the object from parent.

**Running the stub**

Since each abstraction has its unique way of running their stub (functions vs pods), I do not try to condense their functionality into a umbrella `Run` method. That means you must call them from their individual invocation endpoints

function: /function/:id/:stubId
pod: /pod/run/:stubId

Notes:
- I assume that the object is free to clone if attached to a public stub
- I skipped cloning onDeploy stubs to simplify the PR

### DIFF
--- a/pkg/api/v1/stub.go
+++ b/pkg/api/v1/stub.go
@@ -2,15 +2,12 @@ package apiv1
 
 import (
 	"context"
-	"crypto/sha256"
 	"database/sql"
-	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"net/http"
 	"os"
 	"path"
-	"strings"
 
 	"github.com/beam-cloud/beta9/pkg/auth"
 	"github.com/beam-cloud/beta9/pkg/common"
@@ -208,14 +205,11 @@ func (g *StubGroup) copyObjectContents(ctx context.Context, workspace *types.Wor
 	parentObjectPath := path.Join(types.DefaultObjectPath, stub.Workspace.Name)
 	parentObjectFilePath := path.Join(parentObjectPath, parentObject.ExternalId)
 
-	hash := sha256.Sum256([]byte(parentObject.Hash + workspace.ExternalId))
-	hashStr := hex.EncodeToString(hash[:])
-
-	if existingObject, err := g.backendRepo.GetObjectByHash(ctx, hashStr, workspace.Id); err == nil {
+	if existingObject, err := g.backendRepo.GetObjectByHash(ctx, parentObject.Hash, workspace.Id); err == nil {
 		return existingObject.Id, nil
 	}
 
-	newObject, err := g.backendRepo.CreateObject(ctx, hashStr, parentObject.Size, workspace.Id)
+	newObject, err := g.backendRepo.CreateObject(ctx, parentObject.Hash, parentObject.Size, workspace.Id)
 	if err != nil {
 		return 0, err
 	}
@@ -259,13 +253,10 @@ func (g *StubGroup) cloneStub(ctx context.Context, workspace *types.Workspace, s
 		}
 	}
 
-	// Get secrets
-	missingSecrets := []string{}
 	for _, secret := range parentSecrets {
 		secret, err := g.backendRepo.GetSecretByName(ctx, workspace, secret.Name)
 		if err != nil {
 			if err == sql.ErrNoRows {
-				missingSecrets = append(missingSecrets, secret.Name)
 				continue
 			}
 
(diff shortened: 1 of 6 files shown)


pkg/api/v1/stub.go
```diff
@@ -146,3 +156,143 @@ func (g *StubGroup) GetURL(ctx echo.Context) error {
...
+func (g *StubGroup) copyObjectContents(ctx context.Context, workspace *types.Workspace, stub *types.StubWithRelated) (uint, error) {
+	parentObject, err := g.backendRepo.GetObjectByExternalStubId(ctx, stub.ExternalId, stub.WorkspaceId)
+	if err != nil {
+		return 0, err
+	}
+
+	parentObjectPath := path.Join(types.DefaultObjectPath, stub.Workspace.Name)
+	parentObjectFilePath := path.Join(parentObjectPath, parentObject.ExternalId)
+
+	hash := sha256.Sum256([]byte(parentObject.Hash + workspace.ExternalId))
```
jsun-m, line 211: There is a unique_constraint on hash so I either do this to redo the hash for this specific workspace or update the unqiue_constraint to be unique together by the hash and workspace

luke-lombardi: hm, I could see this being an issue because the hash is the hash _of_ the contents, meaning you can use the hash as a lookup key if necessary. I would argue we should modify the unique constraint to be on workspace + hash instead of creating this new hash that doesn't represent the real content of the object.

pkg/api/v1/stub.go
```diff
@@ -146,3 +156,143 @@ func (g *StubGroup) GetURL(ctx echo.Context) error {
...
+	if existingObject, err := g.backendRepo.GetObjectByHash(ctx, hashStr, workspace.Id); err == nil {
+		return existingObject.Id, nil
+	}
+
+	newObject, err := g.backendRepo.CreateObject(ctx, hashStr, parentObject.Size, workspace.Id)
+	if err != nil {
+		return 0, err
+	}
+
+	newObjectPath := path.Join(types.DefaultObjectPath, workspace.Name)
```
luke-lombardi, line 223: Hm, I wonder about copying the contents. I think it probably does make sense because the "parent" object could be potentially deleted. I wonder if there's a cleaner way to do this. Also keep in mind I changed this behavior in the storage refactor so that it goes through an external S3 client, so we'll need to change that again.

pkg/api/v1/stub.go
```diff
@@ -146,3 +156,143 @@ func (g *StubGroup) GetURL(ctx echo.Context) error {
...
+		concurrencyLimit, err := g.backendRepo.GetConcurrencyLimitByWorkspaceId(ctx, workspace.ExternalId)
+		if err != nil && concurrencyLimit != nil && concurrencyLimit.GPULimit <= 0 {
+			return nil, HTTPBadRequest("GPU concurrency limit is 0.")
+		}
+	}
+
+	// Get secrets
+	missingSecrets := []string{}
+	for _, secret := range parentSecrets {
+		secret, err := g.backendRepo.GetSecretByName(ctx, workspace, secret.Name)
```
luke-lombardi, line 265: What does this do exactly? It's not cloning the secrets "values" right? It just complains if there are missing secrets? That feels a little off

luke-lombardi (changes requested):

pkg/api/v1/stub.go
jsun-m: I checked how the hash was created and it didn't seem to have been done by content. It is based off the name of the tmp file that is generated locally.
```
temp_zip_name = tempfile.NamedTemporaryFile(delete=False).name
...
hash = self._calculate_sha256(temp_zip_name)
```

pkg/api/v1/stub.go
jsun-m: It would have to be copied because the parent contents are under /data/objects/[:workspace.name] which would be inaccessible by the child stub.

pkg/api/v1/stub.go
jsun-m: Yea we remove the parent's secrets and add the secrets that are required from the child. I can just not error

luke-lombardi: What do you mean the secrets that are required from the child? I don't follow

pkg/api/v1/stub.go
jsun-m: oh that makes more sense

pkg/api/v1/stub.go
luke-lombardi: yeah I think failing during runtime makes more sense

luke-lombardi (approved):

PR #1047, opened 2025-03-19
temp fix for skopeo build

dleviminzi:


docker/Dockerfile.worker
```diff
@@ -62,6 +42,20 @@ make BUILDTAGS="cni"
...
 EOT
 
+ENV CGO_ENABLED="0"  \
+    DISABLE_DOCS="1" \
+    GO_DYN_FLAGS=""  \
+    BUILDTAGS="containers_image_openpgp"
+
+RUN <<EOT
+set -eux
+git clone https://github.com/containers/skopeo.git /workspace/skopeo
```
luke-lombardi, line 52: can we pin this to a specific commit? also what exactly happened to the build -- need more context

luke-lombardi (changes requested):

docker/Dockerfile.worker
dleviminzi: Its the same thing that happened to the other builds. They require go 1.23 and our base is using 1.22. We need to get around to figuring out that panic that comes with 1.23 so we can remove all this hacky stuff. 

We can pin to a commit, but we could also just download binaries instead of building all of these.

luke-lombardi (approved):

PR #1049, opened 2025-03-20
Feat: Update protobuf for Stubs

jsun-m:


pkg/types/types.proto
```diff
@@ -65,6 +65,7 @@ message ContainerRequest {
   bool checkpoint_enabled = 19;
   BuildOptions build_options = 20;
   repeated uint32 ports = 21;
+  double cost_per_ms = 22;
```
luke-lombardi, line 68: @dleviminzi can we remove this? Wasn't this from before we decided to do the cost calculation post-scheduling / in the worker

dleviminzi: Yeah that's right. I think this slipped through

luke-lombardi (approved):

PR #1059, opened 2025-03-21
local blobcache metadata

dleviminzi:


pkg/types/config.go
```diff
@@ -61,6 +62,31 @@ type RedisConfig struct {
 	RouteByLatency     bool          `key:"routeByLatency" json:"route_by_latency"`
 }
 
+type BlobCacheMetadataConfig struct {
```
luke-lombardi, line 65: I don't think this config should live here. If we're referencing external config structs from blobcache they should live there

luke-lombardi (changes requested):

pkg/worker/worker.go
```diff
@@ -157,6 +157,14 @@ func NewWorker() (*Worker, error) {
 
 	var cacheClient *blobcache.BlobCacheClient = nil
 	if config.Worker.BlobCacheEnabled {
+		if config.BlobCacheMetadata.ValkeyConfig.Enabled {
+			config.BlobCache.Metadata.RedisAddr = "127.0.0.1:26379"
+			config.BlobCache.Metadata.RedisPasswd = config.BlobCacheMetadata.ValkeyConfig.Password
+			config.BlobCache.Metadata.RedisTLSEnabled = false
+			config.BlobCache.Metadata.RedisMode = blobcache.RedisModeSentinel
+			config.BlobCache.Metadata.RedisMasterName = "bc-metadata-master"
```
dleviminzi, line 165: Going to bump these out into constants

go.sum
```diff
@@ -87,8 +87,8 @@ github.com/aws/aws-sdk-go-v2/service/sts v1.30.1/go.mod h1:jiNR3JqT15Dm+QWq2SRgh
 github.com/aws/smithy-go v1.20.3 h1:ryHwveWzPV5BIof6fyDvor6V3iUL7nTfiTKXHiW05nE=
 github.com/aws/smithy-go v1.20.3/go.mod h1:krry+ya/rV9RDcV/Q16kpu6ypI4K2czasz0NC3qS14E=
 github.com/aybabtme/rgbterm v0.0.0-20170906152045-cc83f3b3ce59/go.mod h1:q/89r3U2H7sSsE2t6Kca0lfwTK8JdoNGS/yzM/4iH5I=
-github.com/beam-cloud/blobcache-v2 v0.0.0-20250228212315-d1ce0d5702fc h1:jxwjRjAN7z9UpP5JVlb5ePcBdtxoaNyD2iufIakOwYk=
-github.com/beam-cloud/blobcache-v2 v0.0.0-20250228212315-d1ce0d5702fc/go.mod h1:fZM8IC6mVPRlVV1Rvaj5HaOd/vf9DrKKdhtOS3+t8S4=
+github.com/beam-cloud/blobcache-v2 v0.0.0-20250327165758-9061aa33f327 h1:tuUSTjTqiILvHbrvic1QQHGuKxzj5JZa2hFjNOjNSMI=
```
luke-lombardi, line 90: reminder to bump this once we merge the other stuff

pkg/api/v1/machine.go
```diff
@@ -33,18 +45,34 @@ func NewMachineGroup(g *echo.Group, providerRepo repository.ProviderRepository,
...
+func (g *MachineGroup) ListPoolMachines(ctx echo.Context) error {
+	cc, _ := ctx.(*auth.HttpAuthContext)
+	if (cc.AuthInfo.Token.TokenType != types.TokenTypeMachine) && (cc.AuthInfo.Token.TokenType != types.TokenTypeWorker) {
+		return HTTPForbidden("Invalid token")
+	}
+
+	poolName := ctx.QueryParam("pool_name")
+	providerName := ctx.QueryParam("provider_name")
+
+	neighbors, err := g.providerRepo.ListAllMachines(providerName, poolName, false)
```
luke-lombardi, line 61: prolly would just call this `machines` instead of neighbors

pkg/types/config.go
```diff
@@ -11,21 +11,22 @@ import (
...
+	GatewayService    GatewayServiceConfig              `key:"gateway" json:"gateway_service"`
+	FileService       FileServiceConfig                 `key:"fileService" json:"file_service"`
+	ImageService      ImageServiceConfig                `key:"imageService" json:"image_service"`
+	Storage           StorageConfig                     `key:"storage" json:"storage"`
+	Worker            WorkerConfig                      `key:"worker" json:"worker"`
+	Providers         ProviderConfig                    `key:"providers" json:"providers"`
+	Tailscale         TailscaleConfig                   `key:"tailscale" json:"tailscale"`
+	Proxy             ProxyConfig                       `key:"proxy" json:"proxy"`
+	Monitoring        MonitoringConfig                  `key:"monitoring" json:"monitoring"`
+	BlobCacheMetadata blobcache.BlobCacheMetadataConfig `key:"blobcacheMetadata" json:"blobcache_metadata"`
```
luke-lombardi, line 27: I see why you're doing this, but it feels weird to have two top level structs like this. I would argue for either having a single Blobcache one, or have a top level Caching struct or something that contains both. I think the argument for the top level struct might be stronger since we've wanted to separate client configuration for a while.

pkg/worker/worker.go
luke-lombardi: bump

luke-lombardi (approved):

PR #1068, opened 2025-03-31
Feat: Add app abstraction

jsun-m:
This PR is part of a [stack](https://docs.mrge.io/overview), managed by [mrge](https://mrge.io):

* `main` (default branch)
* **[#1068: Feat: Add app abstraction](https://github.com/beam-cloud/beta9/pull/1068)** ⬅️ Current PR ([View on mrge](https://mrge.io/pr/beam-cloud/beta9/pull/1068))
* [#1082: Feat: Add additional app endpoints and new serialization methods](https://github.com/beam-cloud/beta9/pull/1082)
* [#1086: Feat: Add serializer library to better parse, sanitize, and organize data models for public APIs](https://github.com/beam-cloud/beta9/pull/1086)

---

This PR will add an additional object when creating stubs and deployments that will group all stubs (deployments, stubs, serves, function runs) under a namespace within a workspace.

These are the following things that are required when releasing
- [x] Goose migration will automatically create NULLable app_ids (we do not want them to be nullable eventually. all stubs and deployments will have an app associated with it)
- [x] Run migration script to migrate app stubs and deployments under an app
- [x] Add another migration in another PR to make app_id field not nullable
- ~~Update all events to include app_id~~ (ticket created)

Update: All above is down in single goose migration under a transaction

Format of migration:
1. Stubs associated with deployments will use the deployment name as the app name
2. Stubs not associated with deployments will use the stub name as the app name


pkg/repository/backend_postgres.go
```diff
@@ -792,10 +792,11 @@ func (r *PostgresBackendRepository) GetOrCreateStub(ctx context.Context, name, s
 		return types.Stub{}, err
 	}
 
+	// @Luke slightly concerned that this is can get the wrong stub. If name + stubtype + objectId + config matches (Likely can happen from clone but haven't seen it yet)
```
jsun-m, line 795: @jsun-m check this again

pkg/api/v1/app.go
```diff
@@ -0,0 +1,96 @@
...
+			return HTTPBadRequest("Failed to get apps")
+		}
+
+		if deployments != nil && len(deployments) > 0 {
+
+			appsWithLatest.Data[i].Deployment = &deployments[0]
+			continue
+		}
+
+		// App has no deployments we get a stub
```
luke-lombardi, line 78: Can you elaborate on this comment / what this is doing?

pkg/repository/backend_postgres.go
```diff
@@ -839,14 +848,18 @@ func (r *PostgresBackendRepository) GetOrCreateStub(ctx context.Context, name, s
...
-	if err := r.client.GetContext(ctx, &stub, queryCreate, name, stubType, string(configJSON), objectId, workspaceId); err != nil {
+    INSERT INTO stub (name, type, config, object_id, workspace_id, app_id)
+    VALUES ($1, $2, $3, $4, $5, $6)
+    RETURNING id, external_id, name, type, config, config_version, object_id, workspace_id, created_at, updated_at, app_id;
+  `
+	if err := r.client.GetContext(ctx, &stub, queryCreate, name, stubType, string(configJSON), objectId, workspaceId, appId); err != nil {
 		return types.Stub{}, err
 	}
 
+	if err := r.updateAppActivity(ctx, appId); err != nil {
```
luke-lombardi, line 859: what is the activity thing? is this just for the dashboard?

sdk/src/beta9/abstractions/endpoint.py
```diff
@@ -178,6 +182,8 @@ class ASGI(Endpoint):
     Decorator which allows you to create an ASGI application.
 
     Parameters:
+        app (str):
+            The name of the app to associated with
```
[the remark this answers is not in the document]
luke-lombardi: cc @jsun-m , I think we should improve these docstrings

luke-lombardi (changes requested):

pkg/api/v1/app.go
jsun-m: If the app doesn't have a deployment, we get the latest stub.
This goes back to how this method is suppose to get the latest deployment or stub

pkg/repository/backend_postgres.go
jsun-m: this updates the `updated_at` field of an app whenever a deployment or stub is created

sdk/src/beta9/clients/gateway/__init__.py
```diff
@@ -332,6 +333,7 @@ class Deployment(betterproto.Message):
     workspace_name: str = betterproto.string_field(9)
     created_at: datetime = betterproto.message_field(10)
     updated_at: datetime = betterproto.message_field(11)
+    app_id: str = betterproto.string_field(12)
```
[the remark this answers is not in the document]
jsun-m: this is correct. it is a str

sdk/src/beta9/abstractions/taskqueue.py
```diff
@@ -33,6 +33,8 @@ class TaskQueue(RunnerAbstraction):
     in python through the .put() method.
 
     Parameters:
+        app (str):
+            The name of the app to associated the task queue with
```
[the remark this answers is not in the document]
jsun-m: updated copy

pkg/api/v1/stub.go
```diff
@@ -348,7 +348,12 @@ func (g *StubGroup) cloneStub(ctx context.Context, workspace *types.Workspace, s
 		return nil, HTTPInternalServerError("Failed to configure volumes")
 	}
 
-	newStub, err := g.backendRepo.GetOrCreateStub(ctx, stub.Name, string(stub.Type), *stubConfig, objectId, workspace.Id, true)
+	app, err := g.backendRepo.GetOrCreateApp(ctx, workspace.Id, stub.Name)
+	if err != nil {
+		return nil, HTTPInternalServerError("Failed to create app")
+	}
+
+	newStub, err := g.backendRepo.GetOrCreateStub(ctx, stub.Name, string(stub.Type), *stubConfig, objectId, workspace.Id, true, app.Id)
```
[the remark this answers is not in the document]
jsun-m: This is the cloneStub method. the stub refers to the stub we are cloning. newStub is the new stub that we generate

pkg/api/v1/app.go
jsun-m: bad comment. updated

sdk/src/beta9/abstractions/function.py
```diff
@@ -122,6 +126,7 @@ def __init__(
             name=name,
             task_policy=task_policy,
             on_deploy=on_deploy,
+            app=app,
```
dleviminzi, line 129: For at least functions, I think it makes sense to set the app name to the file name or directory name by default if no name is specified.

dleviminzi (changes requested):

sdk/src/beta9/abstractions/function.py
jsun-m: it's set on the gateway side if there no name

pkg/api/v1/app.go
```diff
@@ -0,0 +1,96 @@
...
+
+	for i := range apps.Data {
+		appsWithLatest.Data[i].App = apps.Data[i]
+
+		deployments, err := a.backendRepo.ListDeploymentsWithRelated(ctx.Request().Context(), types.DeploymentFilter{AppId: apps.Data[i].ExternalId})
+		if err != nil {
+			return HTTPBadRequest("Failed to get apps")
+		}
+
+		if deployments != nil && len(deployments) > 0 {
```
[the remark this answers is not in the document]
jsun-m: It's a thing with sql in golang. If the list returns is empty, the value of the array would be nil. Not sure why we didnt handle this in the backend_postgres functions.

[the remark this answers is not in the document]
dleviminzi: The return type is a list and nil can be returned as a list. The `len` function handles this.
```
ListDeploymentsWithRelated(ctx context.Context, filters types.DeploymentFilter) ([]types.DeploymentWithRelated, error)
```
https://goplay.tools/snippet/DLjBisT3mjC

sdk/src/beta9/abstractions/function.py
dleviminzi: Right, but if a user launches multiple functions from a single file will they automatically be grouped into the same App? I think it would make sense if they were and I think the App should probably be inferred from their project directory.

pkg/api/v1/app.go
```diff
@@ -0,0 +1,96 @@
...
+	g.GET("/:workspaceId/latest", auth.WithWorkspaceAuth(group.ListAppWithLatestActivity))
+
+	return group
+}
+
+type AppWithLatestStubOrDeployment struct {
+	types.App
+	Stub       *types.StubWithRelated       `json:"stub"`
+	Deployment *types.DeploymentWithRelated `json:"deployment"`
+}
```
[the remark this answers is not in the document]
dleviminzi: I think I agree with this. If we're going to do it this way from now on, we should make a ticket to clean up places where we do it the old way.

pkg/repository/backend_postgres.go
dleviminzi: Did this end up being a problem?

pkg/api/v1/app.go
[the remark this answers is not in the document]
jsun-m: got it. updated it

sdk/src/beta9/abstractions/function.py
jsun-m: I didn't think about that but that is indeed a really good idea.

dleviminzi (approved):

PR #1071, opened 2025-04-02
feat: add workspace storage creation endpoint

luke-lombardi:


pkg/gateway/services/object.go
```diff
@@ -87,6 +87,20 @@ func (gws *GatewayService) CreateObject(ctx context.Context, in *pb.CreateObject
...
+				Ok:       false,
+				ErrorMsg: "Unable to generate presigned URL",
+			}, nil
+		}
+
+		return &pb.CreateObjectResponse{
+			Ok:           true,
+			ObjectId:     existingObject.ExternalId,
+			PresignedUrl: presignedURL,
+		}, nil
```
dleviminzi, line 103: This feels a bit weird. 

Maybe just check `!in.Overwrite` inside the `err==nil` block and bail early instead of the if-else double check

dleviminzi (approved):

pkg/gateway/services/object.go
luke-lombardi: updated. for context this is to handle an edge case where the object was deleted manually from the remote bucket.

PR #1076, opened 2025-04-02
Update: Add presigned url upload logic for workspace storage

jsun-m:


pkg/abstractions/volume/http.go
```diff
@@ -327,9 +360,19 @@ func (g *volumeGroup) generatePresignGetURL(ctx context.Context, workspace *type
 	}
 
 	key := path.Join(types.DefaultVolumesPrefix, volume.ExternalId, volumePath)
-	url, err := storageClient.GeneratePresignedGetURL(ctx, key, PresignedGetURLExpiration)
-	if err != nil {
-		return "", echo.NewHTTPError(http.StatusInternalServerError, "Failed to generate download token")
+	var url string
+	if urlType == "PUT" {
```
luke-lombardi, line 364: can we use consts from HTTP package for these?

pkg/abstractions/volume/http.go
```diff
@@ -184,10 +185,11 @@ func (g *volumeGroup) DownloadFile(ctx echo.Context) error {
 	}
 
 	if cc.AuthInfo.Workspace.StorageAvailable() {
-		presignedUrl, err := g.generatePresignGetURL(
+		presignedUrl, err := g.generatePresignedURL(
 			ctx.Request().Context(),
 			cc.AuthInfo.Workspace,
 			decodedVolumePath,
+			"GET",
```
luke-lombardi, line 192: use http const for this

pkg/abstractions/volume/http.go
luke-lombardi: can we use also use a switch here

luke-lombardi (approved):

PR #1113, opened 2025-04-16
Fetch and unpack stub code from s3

dleviminzi:
## Summary by mrge
Improved container code fetching by directly downloading stub code from S3 instead of relying on GeeseFS. This change avoids eventual consistency issues and provides a more reliable way to access container code.

**Refactors**
- Added direct S3 download functionality to fetch stub code objects instead of using the filesystem.
- Implemented zip extraction logic to unpack code directly from memory to the container workspace.
- Updated the container mount setup function to accept a context parameter for proper cancellation support.

**Bug Fixes**
- Fixed potential eventual consistency issues when accessing code through GeeseFS.
- Added path validation to prevent zip slip vulnerabilities during extraction.

To avoid eventual consistency problems with GeeseFS, we go direct to object storage for the stub object.

### DIFF
--- a/pkg/worker/mount.go
+++ b/pkg/worker/mount.go
@@ -40,7 +40,7 @@ func (c *ContainerMountManager) SetupContainerMounts(ctx context.Context, reques
 					return err
 				}
 			} else {
-				if err := getMntCodeAndExtract(ctx, request); err != nil {
+				if err := getAndExtractStubCode(ctx, request); err != nil {
 					return err
 				}
 			}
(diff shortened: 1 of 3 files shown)


pkg/worker/mount.go
```diff
@@ -123,3 +126,22 @@ const (
 func checkpointSignalDir(containerId string) string {
 	return fmt.Sprintf("/tmp/%s/criu", containerId)
 }
+
+// getMntCodeAndExtract downloads the object from storage and extracts it to the temp location that will be mounted
+func getMntCodeAndExtract(ctx context.Context, request *types.ContainerRequest) error {
```
luke-lombardi, line 131: let's call this getAndExtractStubCode?

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1114, opened 2025-04-16
Redirect to presigned deployment package

dleviminzi:
## Summary by mrge
Added redirect to presigned URLs for deployment package downloads. This improves performance and security by using temporary authorized links instead of direct file serving.

**New Features**
- Added support for generating presigned URLs with a 10-minute expiration for deployment package downloads.
- Implemented HTTP redirect (307) to the presigned URL when a storage client is available.

**Refactors**
- Modified the DownloadDeploymentPackage handler to first attempt using presigned URLs before falling back to direct file serving.

### DIFF
--- a/pkg/api/v1/deployment.go
+++ b/pkg/api/v1/deployment.go
@@ -17,7 +17,7 @@ import (
 	"github.com/beam-cloud/beta9/pkg/types"
 )
 
-const presignedURLExpiration = 10 * 60
+const presignedURLExpirationSec = 10 * 60
 
 type DeploymentGroup struct {
 	routerGroup   *echo.Group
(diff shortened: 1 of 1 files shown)


pkg/api/v1/deployment.go
```diff
@@ -231,8 +234,21 @@ func (g *DeploymentGroup) DownloadDeploymentPackage(ctx echo.Context) error {
...
 		return HTTPInternalServerError("Failed to get object")
 	}
-	path := getPackagePath(workspace.Name, object.ExternalId)
 
+	if workspaceStorage, err := g.backendRepo.GetWorkspaceStorage(ctx.Request().Context(), workspace.Id); err == nil {
+		storageClient, err := clients.NewStorageClient(ctx.Request().Context(), workspace.Name, workspaceStorage)
+		if err != nil {
+			return HTTPInternalServerError("Failed to get object")
+		}
+		presignedURL, err := storageClient.GeneratePresignedGetURL(ctx.Request().Context(), "objects/"+object.ExternalId, presignedURLExpiration)
```
luke-lombardi, line 243: I believe there is a const for the object prefix in types

luke-lombardi (changes requested):

pkg/api/v1/deployment.go
```diff
@@ -9,12 +9,15 @@ import (
...
 
 	"github.com/beam-cloud/beta9/pkg/auth"
+	"github.com/beam-cloud/beta9/pkg/clients"
 	"github.com/beam-cloud/beta9/pkg/common"
 	"github.com/beam-cloud/beta9/pkg/repository"
 	"github.com/beam-cloud/beta9/pkg/scheduler"
 	"github.com/beam-cloud/beta9/pkg/types"
 )
 
+const presignedURLExpiration = 10 * 60
```
luke-lombardi, line 19: is this seconds? I would clarify in the var name

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1119, opened 2025-04-18
Use geese fork with fixed symlinks

dleviminzi:
## Summary by mrge
Switched to a custom geesefs fork with fixed symlink handling and enabled the --symlink-zeroed option for mounts.

- **Dependencies**
  - Build geesefs from the beam-cloud fork instead of downloading the upstream release.
  - Add --symlink-zeroed flag to geesefs mount command.

### DIFF
--- a/docker/Dockerfile.worker
+++ b/docker/Dockerfile.worker
@@ -56,6 +56,18 @@ make install
 /usr/local/bin/skopeo --version
 EOT
 
+# build geesefs fork
+# ========================
+FROM golang AS geesefs
+
+WORKDIR /workspace
+
+RUN <<EOT
+git clone https://github.com/beam-cloud/geesefs.git
+cd geesefs
+make build
+EOT
+
 # nvidia-container-toolkit
 # ========================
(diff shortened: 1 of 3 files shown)


dleviminzi:
For reference 
https://github.com/yandex-cloud/geesefs/compare/master...beam-cloud:geesefs:master

docker/Dockerfile.worker
```diff
@@ -56,6 +56,18 @@ make install
...
 EOT
 
+# build geesefs fork
+# ========================
+FROM golang AS geesefs
+
+WORKDIR /workspace
+
+RUN <<EOT
+git clone https://github.com/beam-cloud/geesefs.git
```
luke-lombardi, line 66: Let's pin the commit here just for clarity

luke-lombardi (changes requested):

docker/Dockerfile.worker
dleviminzi: Yeah it does. Pinned commit

luke-lombardi (approved):

PR #1121, opened 2025-04-18
Add actual cost for elapsed duration

dleviminzi:
// If there's a non-AI description already, add a separator
## Summary by mrge
Added calculation and reporting of the actual container cost for the elapsed duration.

- **New Features**
  - Now tracks and emits the total cost for each container based on its runtime.


pkg/worker/usage.go
```diff
@@ -97,17 +98,15 @@ func (wm *WorkerUsageMetrics) EmitContainerUsage(ctx context.Context, request *t
...
+		return 0
 	}
 
 	costPerMs, err := wm.containerCostClient.GetContainerCostPerMs(request)
 	if err != nil {
 		log.Error().Str("container_id", request.ContainerId).Err(err).Msg("unable to get container cost per ms")
 	}
 
-	request.CostPerMs = costPerMs
+	return costPerMs
```
jsun-m, line 111: ah this makes more sense. Previously was just sending 1 tick of it?

dleviminzi: yeah exactly

jsun-m (approved):

PR #1131, opened 2025-04-24
shadow copy objects for image migration

luke-lombardi:
For what it's worth, I don't love this -- but it allows us to migrate any new image builds to new storage immediately. Old images will be asynchronously copied to the new bucket, so they will only cost us one more pull from the "secondary" registry. If anyone has a better idea let me know. 
    

## Summary by mrge
Added support for shadow copying image objects from a secondary S3 registry to a primary S3 registry during image pulls, enabling seamless image migration.

- **New Features**
  - Added config for primary and secondary S3 registries.
  - On image pull miss in the primary, fetches from secondary and copies objects back to primary in the background.
  - Extended object store interfaces to support streaming reads and writes.


mernit (approved):

pkg/registry/registry.go
```diff
@@ -31,20 +31,20 @@ type ImageRegistry struct {
 	ImageFileExtension string
 }
 
-func NewImageRegistry(config types.AppConfig) (*ImageRegistry, error) {
+func NewImageRegistry(config types.AppConfig, registry types.S3ImageRegistry) (*ImageRegistry, error) {
 	var err error
 	var store ObjectStore
 
-	var imageFileExtension string = localImageFileExtension
+	var imageFileExtension string = LocalImageFileExtension
```
dleviminzi, line 38: nitpick but prefer var block

pkg/registry/registry.go
```diff
@@ -74,31 +74,37 @@ func (r *ImageRegistry) Size(ctx context.Context, imageId string) (int64, error)
 	return r.store.Size(ctx, fmt.Sprintf("%s.%s", imageId, r.ImageFileExtension))
 }
 
+func (r *ImageRegistry) GetStore() ObjectStore {
+	return r.store
+}
```
dleviminzi, line 79: In general I don't like this pattern of adding objects to return other unexported structs. It feels hacky and gets confusing fast. 

For instance `CopyObjects` has the following signature 
```go
CopyObjects(ctx context.Context, keys []string, sourceRegistry, dstRegistry ObjectStore)
```
Yet, `sourceRegistry` and `dstRegistry` are not actually registry objects at all. They are object store objects.

To me this feels like there should be a `CopyFromRegistry` method on the registry type itself that takes a `sourceRegistry`. Then there can be methods like `GetObjectReader` and `PutObjectReader` on the registry type that internally access the `ObjectStore` methods needed.

pkg/worker/image.go
```diff
@@ -237,11 +244,23 @@ func (c *ImageClient) PullLazy(ctx context.Context, request *types.ContainerRequ
...
-			return elapsed, err
+
+			sourceRegistry = c.config.ImageService.Registries.S3.Secondary
+			err = c.secondaryRegistry.Pull(ctx, remoteArchivePath, imageId)
+			if err != nil {
+				return elapsed, err
+			}
+
+			// HACK: Async copy the archives to the primary registry if it exists in the secondary registry
+			fullArchivePath := fmt.Sprintf("%s/%s.%s", c.imageCachePath, imageId, registry.LocalImageFileExtension)
```
dleviminzi, line 262: Would be nice to bump this into its own method just to minimize adding to the sprawl

pkg/registry/registry.go
luke-lombardi: Updated

PR #1134, opened 2025-04-24
fix: send periodic keepalives to keep stream alive

luke-lombardi:
## Summary by mrge
Added periodic keepalive messages to container streams to prevent disconnections and improved object existence checks for storage operations.

- **Bug Fixes**
  - Send keepalive messages every 10 seconds during container streaming to keep the connection alive.
  - Use a faster and more reliable method to check if objects exist in storage.

### DIFF
--- a/pkg/gateway/services/container.go
+++ b/pkg/gateway/services/container.go
@@ -235,12 +235,9 @@ func (gws *GatewayService) AttachToContainer(stream pb.GatewayService_AttachToCo
 			case <-ctx.Done():
 				return
 			case <-keepaliveTicker.C:
-				err := stream.Send(&pb.AttachToContainerResponse{
+				stream.Send(&pb.AttachToContainerResponse{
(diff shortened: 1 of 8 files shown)


pkg/gateway/services/container.go
```diff
@@ -221,6 +225,28 @@ func (gws *GatewayService) AttachToContainer(stream pb.GatewayService_AttachToCo
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-keepaliveTicker.C:
+				err := stream.Send(&pb.AttachToContainerResponse{
+					Output: "",
+				})
+				if err != nil {
+					keepaliveErrCh <- err
```
jsun-m, line 243: Is this being used?

luke-lombardi: removed

jsun-m (approved):

PR #1137, opened 2025-04-25
Feat: Updating stub cloning with new storage class

jsun-m:
## Summary by mrge
Updated stub cloning to support workspaces with custom storage, allowing objects to be copied between different storage backends.

- **New Features**
  - Stub cloning now detects and uses workspace storage clients when available.
  - Added support for copying objects directly between storage backends or falling back to volume mounts.


pkg/types/backend.go
```diff
@@ -78,6 +78,10 @@ type WorkspaceWithRelated struct {
 	Storage          *WorkspaceStorage `db:"storage" json:"storage" serializer:"storage"`
 }
 
+func (w *WorkspaceWithRelated) StorageAvailable() bool {
```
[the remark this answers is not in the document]
jsun-m: this is currently necessary

luke-lombardi (approved):

luke-lombardi (changes requested):
@jsun-m I think we need to update this now that we changed the signatures in the storage client (most methods take bucket name now)

pkg/api/v1/stub.go
```diff
@@ -280,46 +286,114 @@ func (g StubGroup) configureVolumes(ctx context.Context, volumes []*pb.Volume, w
 	return nil
 }
 
-func (g *StubGroup) copyObjectContents(ctx context.Context, workspace *types.Workspace, stub *types.StubWithRelated) (uint, error) {
+func (g *StubGroup) copyObjectContents(ctx context.Context, childWorkspace *types.Workspace, stub *types.StubWithRelated) (uint, error) {
+	parentWorkspace, err := g.backendRepo.GetWorkspaceWithRelated(ctx, stub.Workspace.Id)
+	if err != nil {
+		return 0, err
+	}
+
```
dleviminzi, line 294: To me it feels like all of the logic in this method should be encapsulated in a workspace method. Also, I'm not sure that the terms of "parent" and "child" really make sense here. "source" and "destination" make more sense imo. 

In my opinion, in `cloneStub`, instead of this: 
```go
objectId, err := g.copyObjectContents(ctx, workspace, stub)
```
It should be 
```go
objectId, err := workspace.CopyStubObject(sourceStub)
```
Maybe even `CopyGenericStubObject` to make it clear that it serves a specific purpose.

pkg/api/v1/stub.go
```diff
@@ -280,46 +286,114 @@ func (g StubGroup) configureVolumes(ctx context.Context, volumes []*pb.Volume, w
...
+	newObjectVolumePath := path.Join(types.DefaultObjectPath, childWorkspace.Name)
+	newObjectVolumeFilePath := path.Join(newObjectVolumePath, newObject.ExternalId)
+	newObjectStorageFilePath := path.Join(types.DefaultObjectPrefix, newObject.ExternalId)
+
+	// If both workspaces have the storage client available, copy the object with the storage client
+	if parentWorkspace.StorageAvailable() && childWorkspace.StorageAvailable() {
+		storageClient, err := clients.NewDefaultStorageClient(ctx, g.config)
+		if err != nil {
+			return 0, err
+		}
```
dleviminzi, line 330: I think this is potentially problematic, right @luke-lombardi?

Will this always only need the default storage client or could a bucket the user registered need to be used here?

pkg/api/v1/stub.go
```diff
@@ -280,46 +286,114 @@ func (g StubGroup) configureVolumes(ctx context.Context, volumes []*pb.Volume, w
...
+		parentStorageClient, err := clients.NewWorkspaceStorageClient(ctx, parentWorkspace.Name, parentWorkspace.Storage)
+		if err != nil {
+			return 0, err
+		}
 
-	input, err := os.ReadFile(parentObjectFilePath)
-	if err != nil {
-		g.backendRepo.DeleteObjectByExternalId(ctx, newObject.ExternalId)
-		return 0, err
+		input, err = parentStorageClient.Download(ctx, parentObjectStorageFilePath)
```
dleviminzi, line 354: Should probably use GetObject and PutObject instead Download and Upload.

dleviminzi (changes requested):

pkg/api/v1/stub.go
jsun-m: Download and Upload are wrapper around GetObject, PutObject

pkg/api/v1/stub.go
jsun-m: @luke-lombardi is there anything wrong with using the DefaultStorage client in the situation involving cross-bucket interactions?

pkg/api/v1/stub.go
jsun-m: You mean workspace as in the type? I don't think we should be adding complex functionality into the types folder

pkg/api/v1/stub.go
jsun-m: done

dleviminzi (approved):

luke-lombardi (approved):

PR #1138, opened 2025-04-25
Cache from s3 instead of fuse

dleviminzi:
Resolve BE-2569

## Summary by mrge
Switched file caching to use S3 as the source instead of FUSE, improving reliability and simplifying the caching process.

- **Refactors**
  - Updated cache logic to pull files and images directly from S3.
  - Simplified related code and updated dependencies.

**Requires** https://github.com/beam-cloud/blobcache-v2/pull/49
https://github.com/beam-cloud/geesefs/pull/5


pkg/worker/image.go
```diff
@@ -181,15 +181,16 @@ func (c *ImageClient) PullLazy(ctx context.Context, request *types.ContainerRequ
...
+		sourcePath := filepath.Join(c.config.ImageService.Registries.S3.Primary.Region, c.config.ImageService.Registries.S3.Primary.BucketName, imageKey)
 
 		// Create constant backoff
 		b := backoff.NewConstantBackOff(300 * time.Millisecond)
 		imageLocked := false
 
 		operation := func() error {
 			baseBlobFsContentPath := fmt.Sprintf("%s/%s", baseFileCachePath, sourcePath)
-			if _, err := os.Stat(baseBlobFsContentPath); err == nil && c.cacheClient.IsPathCachedNearby(ctx, "/"+sourcePath) {
+			if _, err := os.Stat(baseBlobFsContentPath); err == nil && c.cacheClient.IsPathCachedNearby(ctx, "/"+imageKey) {
```
luke-lombardi, line 193: I would triple check that the routing key and the image key are exactly the same. This caused an issue in prod last night because the routing key and path passed here were different. I think below you're using the sourcePath (the full s3 qualifier), and up here in the filesystem you're using the image key.

luke-lombardi (changes requested):

PR #1191, opened 2025-05-12
fix: ensure workers can exit after flushing files (geesefs bug)

luke-lombardi:


pkg/worker/storage_manager.go
```diff
@@ -134,11 +134,31 @@ func (s *WorkspaceStorageManager) Unmount(workspaceName string) error {
...
 		return true
 	})
 
+	for _, workspaceName := range mountsToDelete {
+		log.Info().Str("workspace_name", workspaceName).Msg("unmounting storage")
+		err := s.Unmount(workspaceName)
+		if err != nil {
+			log.Error().Str("workspace_name", workspaceName).Err(err).Msg("failed to unmount storage")
+		}
+		log.Info().Str("workspace_name", workspaceName).Msg("unmounted storage")
```
dleviminzi, line 159: Probably don't need both outer logs here. Also the middle log will just fall through and log `unmounted storage`

dleviminzi (approved):
small thing with logs

PR #1198, opened 2025-05-13
fix context leak

dleviminzi:


pkg/worker/lifecycle.go
```diff
@@ -509,7 +509,8 @@ func (s *Worker) getContainerEnvironment(request *types.ContainerRequest, option
 }
 
 // spawn a container using runc binary
-func (s *Worker) spawn(ctx context.Context, request *types.ContainerRequest, spec *specs.Spec, outputLogger *slog.Logger, opts *ContainerOptions) {
+func (s *Worker) spawn(ctx context.Context, cancel context.CancelFunc, request *types.ContainerRequest, spec *specs.Spec, outputLogger *slog.Logger, opts *ContainerOptions) {
```
luke-lombardi, line 512: passing this in feels sloppy to me

luke-lombardi (changes requested):

pkg/worker/lifecycle.go
luke-lombardi: same with RunContainer

dleviminzi: I guess we could wrap it again.

luke-lombardi: you can just inherit from the context can't you? ```// spawn a container using runc binary
func (s *Worker) spawn(ctx context.Context, request *types.ContainerRequest, spec *specs.Spec, outputLogger *slog.Logger, opts *ContainerOptions) {
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()```

dleviminzi: yeah updated it

playground.go
```diff
@@ -0,0 +1,88 @@
+package main
```
luke-lombardi, line 1: remove

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1202, opened 2025-05-13
Feat: Add ability to delete apps and stop its containers

jsun-m:
Add filters for the deleted_at field in each list query.
Add endpoint to delete app: this will also delete all related deployments and stop all containers of the app's stubs.


pkg/api/v1/common.go
```diff
@@ -0,0 +1,55 @@
...
+			if scheduledJob, err := clients.backendRepo.GetScheduledJob(ctx, deployment.Id); err == nil {
+				clients.backendRepo.DeleteScheduledJob(ctx, scheduledJob)
+			}
+		}
+
+		// Stop active containers
+		containers, err := clients.containerRepo.GetActiveContainersByStubId(deployment.Stub.ExternalId)
+		if err == nil {
+			for _, container := range containers {
+				clients.scheduler.Stop(&types.StopContainerArgs{ContainerId: container.ContainerId})
```
luke-lombardi, line 34: should we pass force: true here?

luke-lombardi (changes requested):

pkg/api/v1/common.go
jsun-m: updated

luke-lombardi (approved):

PR #1207, opened 2025-05-15
Fix: Remove unavailable GPU type

NilayYadav:
Remove unavailable GPU types  Remove unavailable GPU types A100-80 and A6000


dleviminzi (dismissed):

dleviminzi (changes requested):
@luke-lombardi and I discussed this a bit last week. Instead of removing things from the SDK like this, we should add new config that allows us to enable/disable hardware types. We can then use it as a filter in the scheduler and SDK.

This is a bit more complicated than your current implementation so feel free to reach out to me and we can pair on it.

PR #1220, opened 2025-05-21
pod proxy reliability improvements

luke-lombardi:


mernit (approved):

pkg/abstractions/image/build.go
```diff
@@ -111,7 +111,7 @@ func (b *Build) resolvePythonVersionRequirement() error {
 		return nil
 	}
 
-	if b.opts.IgnorePython {
+	if b.opts.IgnorePython && len(b.opts.PythonPackages) == 0 {
```
jsun-m, line 114: Whats the use case for this?

pkg/gateway/gateway.go
```diff
@@ -195,7 +195,16 @@ func (g *Gateway) initHttp() error {
...
 
-	e.Pre(middleware.RemoveTrailingSlash())
+	skipSubdomainRoutes := func(c echo.Context) bool {
+		baseDomain := gatewaymiddleware.ParseHostFromURL(g.Config.GatewayService.HTTP.GetExternalURL())
+		subdomain := gatewaymiddleware.ParseSubdomain(c.Request().Host, baseDomain)
+		return subdomain != ""
+	}
+
+	e.Pre(middleware.RemoveTrailingSlashWithConfig(middleware.TrailingSlashConfig{
+		Skipper: skipSubdomainRoutes,
```
jsun-m, line 205: Whats this for?

jsun-m (commented):
Whats the context behind these?

pkg/abstractions/image/build.go
dleviminzi: There is a bug where pods (which ignore python installation) won't install requested packages because the python version gets dropped.

dleviminzi (approved):

pkg/gateway/gateway.go
luke-lombardi: we've been stripping of the trailing slash for all routes, which in some cases is not desired.. for example, the pod I was using had an endpoint like ` /events/?queryparm`, which didn't work because we were removing the trailing slash

PR #1223, opened 2025-05-22
Feat: connect shell to any existing container

jsun-m:
This PR adds the ability to connect `beta9 shell` to any existing container using
`beta9 shell --container-id [container-id]`

I replaced openssh-server with dropbear which is a lightweight SSH server that is compiled into a single binary that listens on port 2222.
On a regular container, this server is not spun up by default and requires `beta9 shell` to initialize it. 

Once the dropbear server is spun up, it persists until the container spins down.


pkg/worker/lifecycle.go
```diff
@@ -224,6 +224,10 @@ func (s *Worker) RunContainer(ctx context.Context, request *types.ContainerReque
 		request.Ports = []uint32{uint32(containerInnerPort)}
 	}
 
+	// Expose SSH port
+	request.Ports = append(request.Ports, uint32(2222))
```
luke-lombardi, line 228: Won't this always expose the SSH port? I don't think we always want to expose it right?

luke-lombardi (changes requested):

pkg/abstractions/shell/shell.go
```diff
@@ -35,11 +36,14 @@ const (
 	containerWaitTimeoutDurationS time.Duration = 5 * time.Minute
 	containerWaitPollIntervalS    time.Duration = 1 * time.Second
 	containerKeepAliveIntervalS   time.Duration = 5 * time.Second
+	startupScript                 string        = `exec /usr/local/bin/dropbear -p 2222 -R -E -F 2> /etc/dropbear/logs.txt`
```
luke-lombardi, line 39: if the prot is a const, can we use it here? in a format string

pkg/abstractions/shell/shell.go
```diff
@@ -141,12 +145,132 @@ func (ss *SSHShellService) handleTTLEvents() {
...
+		return false
+	}
+
+	conn, err := network.ConnectToHost(ctx, addr, time.Second*30, ss.tailscale, ss.config.Tailscale)
+	if err != nil {
+		return false
+	}
+
+	// Set read timeout so it doesn't hang forever
+	deadline := time.Now().Add(2 * time.Second)
```
luke-lombardi, line 165: can you make this time a const as well

pkg/abstractions/shell/shell.go
```diff
@@ -141,12 +145,132 @@ func (ss *SSHShellService) handleTTLEvents() {
...
+		}, nil
+	}
+
+	ok := ss.checkForExistingSSHServer(ctx, containerAddr)
+	if !ok {
+		go func() {
+			// This only dies if the container is stopped
+			_, err = runcClient.Exec(containerId, startupScript, []string{})
+			if err != nil {
+				log.Println("Failed to execute startup script:", err)
```
luke-lombardi, line 248: replace with zerologger instead of built in logger. also this goroutine logic is a little weird

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1225, opened 2025-05-25
Change mount path to avoid conflicts with packages installing uv

dleviminzi:


pkg/abstractions/image/build.go
```diff
@@ -307,7 +307,7 @@ func genContainerId() string {
 func generatePipInstallCommand(pythonPackages []string, pythonVersion string) string {
 	flagLines, packages := parseFlagLinesAndPackages(pythonPackages)
 
-	command := "uv pip install"
+	command := "uv-b9 pip install"
```
luke-lombardi, line 310: This is kind of a weird solution - shouldn't we just put in a specific path that we know user containers won't have instead of changing the executable name?

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1234, opened 2025-05-27
Expose VLLM_CACHE_ROOT in vLLM wrapper 

NilayYadav:
Expose VLLM_CACHE_ROOT as Configurable Environment Variable in the vLLM Class

### DIFF
--- a/sdk/src/beta9/abstractions/integrations/vllm.py
+++ b/sdk/src/beta9/abstractions/integrations/vllm.py
@@ -16,7 +16,7 @@ from ...config import ConfigContext
 from ...type import Autoscaler, GpuType, GpuTypeAlias, QueueDepthAutoscaler
 
 DEFAULT_VLLM_CACHE_DIR = "./vllm_cache"
-
+DEFAULT_VLLM_CACHE_ROOT = "./vllm_cache_root"
 
 # vllm/engine/arg_utils.py:EngineArgs
 @dataclass
@@ -149,6 +149,7 @@ class VLLMArgs:
     disable_chunked_mm_input: bool = False
     data_parallel_size: int = 1
     compilation_config: Optional[Any] = None
+    vllm_cache_root: Optional[str] = DEFAULT_VLLM_CACHE_ROOT
 
 
 class VLLM(ASGI):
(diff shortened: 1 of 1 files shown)


sdk/src/beta9/abstractions/integrations/vllm.py
```diff
@@ -221,10 +224,14 @@ def __init__(
...
         autoscaler: Autoscaler = QueueDepthAutoscaler(),
         vllm_args: VLLMArgs = VLLMArgs(),
+        vllm_cache_root: Optional[str] = None,
     ):
         if vllm_args.download_dir == DEFAULT_VLLM_CACHE_DIR:
             # Add default vllm cache volume to preserve it if custom volumes are specified for chat templates
             volumes.append(Volume(name="vllm_cache", mount_path=DEFAULT_VLLM_CACHE_DIR))
+        
+        if vllm_cache_root and vllm_cache_root not in [v.mount_path for v in volumes]:
+            volumes.append(Volume(name="vllm_cache_root", mount_path=vllm_cache_root))
```
dleviminzi, line 234: This is not really how we should be doing this. If you look at `VLLMArgs` above, you will see the `download_dir` option. We should be overriding that field if the user passes in a cache dir. No need to set environment variables anywhere.

dleviminzi (changes requested):

sdk/src/beta9/abstractions/integrations/vllm.py
NilayYadav: but `download_dir` specifies where model files should be downloaded and stored. On the other hand, `VLLM_CACHE_ROOT` sets the root directory for vLLM's internal cache files—it does not control where models are downloaded or stored for inference.

dleviminzi: Ok. I think we should do the following then: 

1. Add vllm_cache_root to the VLLMArgs class. 
2. Set a default value and always mount the volume.

luke-lombardi: @dleviminzi can you update this based on your comment so we can merge it as well (if not already updated)

PR #1244, opened 2025-05-30
Feat: Configurable vllm version

SorenDreano:
Using a configurable vllm version (along with huggingface_hub) enables users that chose to, to be as close to upstream as possible.

It probably also reduces the burden on the beta9 maintainers.

### DIFF
--- a/sdk/src/beta9/abstractions/integrations/vllm.py
+++ b/sdk/src/beta9/abstractions/integrations/vllm.py
@@ -210,6 +210,8 @@ class VLLM(ASGI):
         gpu: Union[GpuTypeAlias, List[GpuTypeAlias]] = GpuType.NoGPU,
         gpu_count: int = 0,
         image: Image = Image(python_version="python3.11"),
+        vllm_version: str = "0.8.4",
+        huggingface_hub_version: str = "0.30.2",
         workers: int = 1,
         concurrent_requests: int = 1,
         keep_warm_seconds: int = 60,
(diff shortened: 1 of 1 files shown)


dleviminzi:
I'm not opposed to this, but can you add a comment explaining that users will likely have to subclass VLLMArgs to get it working.

SorenDreano:
Indeed I forgot to include the newly added arguments in the docstring. Hopefully I made it clear enough that users might have to subclass

sdk/src/beta9/abstractions/integrations/vllm.py
```diff
@@ -167,6 +167,10 @@ class VLLM(ASGI):
         image (Union[Image, dict]):
             The container image used for the task execution. Whatever you pass here will have an additional `add_python_packages` call
             with `["fastapi", "vllm", "huggingface_hub"]` added to it to ensure that we can run vLLM in the container.
+        vllm_version (str):
+            The version of vLLM that will be installed from PyPI. As the configuration of the vLLM engine depend on the version of vLLM, using a non-default vllm_version might require subclassing VLLMArgs in order to add the missing configuration options. Default is version 0.8.4.
```
dleviminzi, line 171: nitpick: configuration of the vLLM engine depend_s_ on the version of vLLM.

dleviminzi:
```suggestion
            The version of vLLM that will be installed from PyPI. As the configuration of the vLLM engine depends on the version of vLLM, using a non-default vllm_version might require subclassing VLLMArgs in order to add the missing configuration options. Default is version 0.8.4.
```

PR #1286, opened 2025-06-20
Use multi-arch postgres image

### LINKED ISSUE #1285: Postgres image doesn't support arm64
I'm trying to set up the Beam development environment on my Mac (arm64/M1), but I ran into an issue with the Postgres container.

The [current image](https://github.com/beam-cloud/beta9/blob/main/manifests/kustomize/components/postgresql/helm-chart-generator.yaml#L26) used for Postgres (`public.ecr.aws/n4e0e1y0/postgres:16.1`) appears to be built only for linux/amd64. Attempting to pull it for arm64 results in the following error:

```
$ docker pull --platform arm64 public.ecr.aws/n4e0e1y0/postgres:16.1
image with reference public.ecr.aws/n4e0e1y0/postgres:16.1 was found but its platform (linux/amd64) does not match the specified platform (linux/arm64)
```

I would like to propose using a multi-arch image instead. For example  [`/docker/library/postgres`](https://gallery.ecr.aws/docker/library/postgres).

dils2k:
Fixes #1285.
    

---

## Summary by cubic
Switched the Postgres image to a multi-architecture version to support more deployment environments.

### DIFF
--- a/manifests/kustomize/components/postgresql/helm-chart-generator.yaml
+++ b/manifests/kustomize/components/postgresql/helm-chart-generator.yaml
@@ -23,7 +23,7 @@ valuesInline:
         database: main
   image:
     registry: public.ecr.aws
-    repository: n4e0e1y0/postgres
+    repository: docker/library/postgres
     tag: 16.1
     pullPolicy: IfNotPresent
   primary:


dils2k:
It appears that it does not create users and the database now

PR #1332, opened 2025-07-16
change return type of task.put

kyung-min-sun:
## Summary by cubic
Changed the return type of task.put to return a Task object instead of a boolean, allowing callers to access the enqueued task directly.

- **Refactors**
 - Instantiated the API client in the runner abstraction to support fetching the Task by ID.
 - Updated error handling for missing API client.

### DIFF
--- a/sdk/src/beta9/abstractions/base/runner.py
+++ b/sdk/src/beta9/abstractions/base/runner.py
@@ -10,6 +10,7 @@ from ... import terminal
 from ...abstractions.base import BaseAbstraction
 from ...abstractions.image import Image, ImageBuildResult
 from ...abstractions.volume import Volume
+from ...client.client import Client
 from ...clients.gateway import Autoscaler as AutoscalerProto
 from ...clients.gateway import (
     GatewayServiceStub,
@@ -178,6 +179,7 @@ class RunnerAbstraction(BaseAbstraction):
         self.pricing: Optional[PricingPolicy] = pricing
         self.inputs: Optional[Schema] = inputs
         self.outputs: Optional[Schema] = outputs
+        self.client = Client(token=self.config_context.token) if self.config_context.token else None
 
     def print_invocation_snippet(self, url_type: str = "") -> GetUrlResponse:
         """Print curl request to call deployed container URL"""
--- a/sdk/src/beta9/abstractions/base/runner.py
+++ b/sdk/src/beta9/abstractions/base/runner.py
@@ -179,7 +179,13 @@ class RunnerAbstraction(BaseAbstraction):
         self.pricing: Optional[PricingPolicy] = pricing
         self.inputs: Optional[Schema] = inputs
         self.outputs: Optional[Schema] = outputs
+        self.client: Optional[Client] = None
+
+    def get_client(self):
+        if self.client:
+            return self.client
         self.client = Client(token=self.config_context.token) if self.config_context.token else None
+        return self.client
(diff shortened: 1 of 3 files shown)


sdk/src/beta9/abstractions/taskqueue.py
```diff
@@ -250,13 +251,17 @@ def _serve(self, *, dir: str, timeout: int = 0):
...
 
-    def put(self, *args, **kwargs) -> bool:
+    def put(self, *args, **kwargs) -> Union[bool, Task]:
         if not self.parent.prepare_runtime(
             func=self.func,
             stub_type=TASKQUEUE_STUB_TYPE,
         ):
             return False
 
+        if not self.parent.client:
```
luke-lombardi: I don't think we need this check. If you don't have a token set, that means prepare_runtime would fail as well. So we'd never fail to instantiate the Client

sdk/src/beta9/abstractions/taskqueue.py
```diff
@@ -269,4 +274,6 @@ def put(self, *args, **kwargs) -> bool:
             return False
 
         terminal.detail(f"Enqueued task: {r.task_id}")
-        return True
+        task = self.parent.client.get_task_by_id(r.task_id)
```
luke-lombardi: Just return this in-line, i.e.

`return self.parent.client.get_task_by_id(r.task_id)`

sdk/src/beta9/abstractions/taskqueue.py
kyung-min-sun: make sense. guessing we should just ignore the type error then for `self.parent.client.get_task_by_id`?

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -178,6 +179,7 @@ def __init__(
         self.pricing: Optional[PricingPolicy] = pricing
         self.inputs: Optional[Schema] = inputs
         self.outputs: Optional[Schema] = outputs
+        self.client = Client(token=self.config_context.token) if self.config_context.token else None
```
luke-lombardi, line 187: should still include the types here

sdk/src/beta9/abstractions/base/runner.py
```diff
@@ -178,6 +179,13 @@ def __init__(
         self.pricing: Optional[PricingPolicy] = pricing
         self.inputs: Optional[Schema] = inputs
         self.outputs: Optional[Schema] = outputs
+        self.client: Optional[Client] = None
+
+    def get_client(self):
```
luke-lombardi: should add return types to this `-> Client | None`

sdk/src/beta9/abstractions/taskqueue.py
```diff
@@ -269,4 +270,5 @@ def put(self, *args, **kwargs) -> bool:
             return False
 
         terminal.detail(f"Enqueued task: {r.task_id}")
```
luke-lombardi: Last comment. I think now that we return the task object we can remove this "Enqueued task" log.

luke-lombardi (changes requested):

sdk/src/beta9/abstractions/taskqueue.py
kyung-min-sun: done!

luke-lombardi (approved):

PR #1369, opened 2025-08-05
criu -> fixed/stored ip addresses

luke-lombardi:
## Summary by cubic
Added support for assigning and storing fixed IP addresses for containers, including persisting the IP in checkpoint state and passing it through container lifecycle and API layers.

- **New Features**
  - Containers now receive a fixed IP address that is stored and restored with checkpoint state.
  - Updated network setup, scheduler, and API types to handle container IPs.
  - Added tests for IP assignment logic.


pkg/worker/network.go
```diff
@@ -870,3 +911,48 @@ func generateUniqueMAC() net.HardwareAddr {
...
+	}
+
+	if len(candidates) == 0 {
+		return nil, errors.New("unable to assign IP address in range")
+	}
+
+	mathrand.Seed(time.Now().UnixNano())
+	mathrand.Shuffle(len(candidates), func(i, j int) { candidates[i], candidates[j] = candidates[j], candidates[i] })
+
+	ip := candidates[0]
```
jsun-m, line 957: this is such a small range, i think we might still need a lock.

luke-lombardi: lock it still here, it wraps the whole function: 

```

func (m *ContainerNetworkManager) configureContainerNetwork(containerId string, containerVeth netlink.Link, request *types.ContainerRequest) error {
	lockResponse, err := handleGRPCResponse(m.workerRepoClient.SetNetworkLock(m.ctx, &pb.SetNetworkLockRequest{
		NetworkPrefix: m.networkPrefix,
		Ttl:           10,
		Retries:       10,
	}))
	if err != nil {
		return err
	}
	defer m.workerRepoClient.RemoveNetworkLock(m.ctx, &pb.RemoveNetworkLockRequest{
		NetworkPrefix: m.networkPrefix,
		Token:         lockResponse.Token,
	})
	````

PR #1372, opened 2025-08-07
Implement Stub Config Modification Endpoint

cooper-grc:
## Summary by cubic
Added a PATCH endpoint to allow workspace admins to update specific fields in a stub's config using a JSON field path and value.

- **New Features**
  - Supports updating nested config fields by path (e.g., "runtime.cpu").
  - Handles type conversion for common field types.

## Example Test
```
curl -X PATCH http://localhost:1994/api/v1/stub/{workspaceId}/{stubId}/config \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer {token}" \
  -d '{
    "fields": {
      "checkpoint_enabled": true,
      "runtime.gpu": "T4",
      "keep_warm_seconds": 200
    }
  }'
{"message":"Stub config updated successfully. Updated fields: [checkpoint_enabled runtime.gpu keep_warm_seconds]","updated_fields":["checkpoint_enabled","runtime.gpu","keep_warm_seconds"]}
```


pkg/api/v1/stub.go
```diff
@@ -561,3 +564,305 @@ func (g *StubGroup) GetConfig(ctx echo.Context) error {
...
+		return HTTPNotFound()
+	}
+
+	cc, _ := ctx.(*auth.HttpAuthContext)
+	if cc.AuthInfo.Workspace.Id != stub.WorkspaceId {
+		return HTTPNotFound()
+	}
+
+	// Parse the request body
+	var updateReq UpdateConfigRequest
```
luke-lombardi, line 587: In general, we only leave comments when something about the logic is unclear. There are a bunch of comments here where the code speaks for itself.

pkg/api/v1/stub.go
```diff
@@ -561,3 +564,305 @@ func (g *StubGroup) GetConfig(ctx echo.Context) error {
...
+	if err := ctx.Bind(&updateReq); err != nil {
+		return HTTPBadRequest("Failed to decode request body")
+	}
+
+	if len(updateReq.Fields) == 0 {
+		return HTTPBadRequest("At least one field must be provided")
+	}
+
+	// Parse the current config
```
luke-lombardi: same comment as above

pkg/api/v1/stub.go
```diff
@@ -561,3 +564,305 @@ func (g *StubGroup) GetConfig(ctx echo.Context) error {
...
+		return HTTPInternalServerError("Failed to update stub config")
+	}
+
+	return ctx.JSON(http.StatusOK, map[string]interface{}{
+		"message":        fmt.Sprintf("Stub config updated successfully. Updated fields: %v", updatedFields),
+		"updated_fields": updatedFields,
+	})
+}
+
+func (g *StubGroup) updateConfigField(config *types.StubConfigV1, fieldPath string, value interface{}) error {
```
luke-lombardi, line 628: we should add a set of unit tests for this. Doesn't have to mock the database interaction, just the update process for a variety of fields -- particularly important imo is the nested case

pkg/api/v1/stub.go
```diff
@@ -561,3 +564,305 @@ func (g *StubGroup) GetConfig(ctx echo.Context) error {
...
+
+		if jsonTag == fieldName {
+			return v.Field(i), true
+		}
+	}
+
+	return reflect.Value{}, false
+}
+
+func (g *StubGroup) convertValue(targetType reflect.Type, value interface{}) (reflect.Value, error) {
```
luke-lombardi: might be worth moving all of these reflection methods out into a helper/outside of the stub group.

luke-lombardi (changes requested):

PR #1373, opened 2025-08-08
Fix: Make sure to verify image before build server-side

jsun-m:


pkg/abstractions/image/image.go
```diff
@@ -236,6 +185,74 @@ func (is *RuncImageService) BuildImage(in *pb.BuildImageRequest, stream pb.Image
...
+	if in.Dockerfile != "" {
+		opts.addPythonRequirements()
+	}
+
+	imageId, err := getImageID(opts)
+	if err != nil {
+		valid = false
+	}
+
+	exists, err := is.builder.Exists(ctx, imageId)
```
luke-lombardi, line 247: This works, but I think we should add an additional database check into the verifyImage function

luke-lombardi (changes requested):

pkg/abstractions/image/image.go
```diff
@@ -236,6 +189,74 @@ func (is *RuncImageService) BuildImage(in *pb.BuildImageRequest, stream pb.Image
 	return nil
 }
 
+func (is *RuncImageService) verifyImage(ctx context.Context, in *pb.VerifyImageBuildRequest) (string, bool, bool, *BuildOpts, error) {
```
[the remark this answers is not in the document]
jsun-m: this code is supposed to dedupe it

pkg/abstractions/image/image.go
```diff
@@ -236,6 +184,83 @@ func (is *RuncImageService) BuildImage(in *pb.BuildImageRequest, stream pb.Image
...
+		valid = false
+	}
+
+	exists, err := is.builder.Exists(ctx, imageId)
+	if err != nil {
+		return "", false, false, nil, err
+	}
+
+	_, err = is.backendRepo.GetImageClipVersion(ctx, imageId)
+	if err != nil {
```
luke-lombardi, line 253: Can you clean up this logic? the `exists = exists | false check is super messy`

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1391, opened 2025-08-20
Expose Pod, Image, and Gateway over HTTP using GRPC-Gateway

cooper-grc:
## Summary by cubic
Expose Pod, Image, and Gateway gRPC services over HTTP using grpc-gateway, mounted at /api/v1/grpc-gateway. This lets clients call JSON endpoints without a gRPC client.

- **New Features**
  - Added grpc-gateway mux and server wiring (initGrpcGateway) and mounted at /api/v1/grpc-gateway/*.
  - Registered PodService, ImageService, and GatewayService handlers against the gRPC port.
  - Authorization header is forwarded to gRPC metadata; gRPC interceptors handle auth.
  - Updated gen_proto.sh to generate grpc-gateway code for pod/image/gateway with generate_unbound_methods=true.
  - Added generated reverse-proxy files: proto/pod.pb.gw.go, proto/image.pb.gw.go, proto/gateway.pb.gw.go.

- **Dependencies**
  - Added protoc-gen-grpc-gateway@v2.27.1 to bin/setup.sh.
  - Promoted github.com/grpc-ecosystem/grpc-gateway/v2 v2.25.0 to a direct dependency.


bin/gen_proto.sh
```diff
@@ -21,17 +21,17 @@ protoc -I ./pkg/types/ --go_out=./proto --go_opt=paths=source_relative --go-grpc
...
+protoc -I ./googleapis -I ./pkg/types -I ./pkg/gateway/services/repository/ --go_out=./proto --go_opt=paths=source_relative --go-grpc_out=./proto --go-grpc_opt=paths=source_relative ./pkg/gateway/services/repository/container_repo.proto
+protoc -I ./googleapis -I ./pkg/types -I ./pkg/gateway/services/repository/ --go_out=./proto --go_opt=paths=source_relative --go-grpc_out=./proto --go-grpc_opt=paths=source_relative ./pkg/gateway/services/repository/worker_repo.proto
 
 # Generate code for abstractions
-protoc -I $PROTOC_INCLUDE_PATH -I ./pkg/abstractions/image/ --go_out=./proto --go_opt=paths=source_relative --go-grpc_out=./proto --go-grpc_opt=paths=source_relative ./pkg/abstractions/image/image.proto
-protoc -I $PROTOC_INCLUDE_PATH -I ./pkg/abstractions/image/ --python_betterproto_beta9_out=./sdk/src/beta9/clients/ ./pkg/abstractions/image/image.proto
+protoc -I ./googleapis -I ./pkg/abstractions/image/ --go_out=./proto --go_opt=paths=source_relative --go-grpc_out=./proto --go-grpc_opt=paths=source_relative ./pkg/abstractions/image/image.proto
```
jsun-m: What is change for the include path to googleapis? It doesn't exist in the repo

pkg/gateway/gateway.go
```diff
@@ -255,6 +257,30 @@ func (g *Gateway) initGrpc() error {
 	return nil
 }
 
+// initGrpcGateway exposes gRPC services as HTTP endpoints.
+func (g *Gateway) initGrpcGateway(grpcAddr string) error {
+	ctx, cancel := context.WithCancel(context.Background())
```
jsun-m: you should use the context of the gateway as the parent context
ctx, cancel := context.WithCancel(g.ctx)

pkg/gateway/gateway.go
```diff
@@ -255,6 +257,30 @@ func (g *Gateway) initGrpc() error {
...
+	if err := pb.RegisterImageServiceHandlerFromEndpoint(ctx, mux, grpcAddr, opts); err != nil {
+		return err
+	}
+	if err := pb.RegisterGatewayServiceHandlerFromEndpoint(ctx, mux, grpcAddr, opts); err != nil {
+		return err
+	}
+	// No need to add auth middleware: grpc-gateway maps the 'Authorization' header
+	// to gRPC metadata, and the destination gRPC server's interceptor will handle
+	// authorization for every request.
+	g.baseRouteGroup.Any("/grpc-gateway/*", echo.WrapHandler(http.StripPrefix(apiv1.HttpServerBaseRoute+"/grpc-gateway", mux)))
```
jsun-m: not sure if I like the naming. This is more like an /api/v2 path instead of calling it /api/v1/grpc-gateway. People using this endpoint shouldn't need to associated it with an http gateway in front of GRPC. @luke-beamcloud 

Also is this some form of behavior that will allow us to add additional endpoints to the grpc->http generated route groups?
If so, then we can actually call this /api/v2 and either build out all functionality to grpc

pkg/gateway/gateway.go
```diff
@@ -255,6 +257,30 @@ func (g *Gateway) initGrpc() error {
 	return nil
 }
 
+// initGrpcGateway exposes gRPC services as HTTP endpoints.
+func (g *Gateway) initGrpcGateway(grpcAddr string) error {
```
luke-lombardi: this could use a better name - maybe initGrpcProxy

luke-lombardi (changes requested):
one small comment -- also need to fix merge conflict w/ pod proto. Can probably just regenerate things after merging main into your branch

luke-lombardi (approved):

PR #1399, opened 2025-08-27
Add: Checkpoint/restore for pods

jsun-m:
## Summary by cubic
Adds CRIU-based checkpoint/restore for pods with an optional user-defined checkpoint condition. This lets pods start from a pre-warmed state to reduce cold starts.

- New Features
  - SDK: Pod supports checkpoint_enabled and checkpoint_condition; a lightweight runner (beta9.runner.checkpoint) waits for the condition and coordinates checkpoint across workers.
  - Gateway/Proto: GetOrCreateStubRequest adds checkpoint_condition; server passes CHECKPOINT_ENABLED and CHECKPOINT_CONDITION env vars to containers.
  - Worker: Integrates checkpoint/restore with improved error handling and only exposes ports after checkpoint or restore; falls back to a normal run if no checkpoint is found; uses a stronger stop signal for reliability.
  - CLI: Deployment now cleans up artifacts even on failures.

- Migration
  - To enable: Pod(checkpoint_enabled=True, checkpoint_condition=my_func). The function should return True when the pod is ready to be checkpointed.
  - No changes needed for existing deployments (feature is off by default).


pkg/worker/criu.go
```diff
@@ -123,31 +129,30 @@ func (s *Worker) attemptCheckpointOrRestore(ctx context.Context, request *types.
...
-	// If a checkpoint exists but is not available (previously failed), run the container normally
-	bundlePath := filepath.Dir(configPath)
-
-	exitCode, err := s.runcHandle.Run(s.ctx, request.ContainerId, bundlePath, &runc.CreateOpts{
-		OutputWriter: outputWriter,
-		Started:      startedChan,
-	})
-
-	return exitCode, request.ContainerId, err
+	return -1, "", fmt.Errorf("checkpoint not found")
```
jsun-m, line 150: The parent function already handles this exact intent of running the code without checkpoints as a fallback. This is just duplicated.

pkg/worker/criu.go
```diff
@@ -214,6 +219,11 @@ func (s *Worker) createCheckpoint(ctx context.Context, request *types.ContainerR
 		if updateStateErr != nil {
 			log.Error().Str("container_id", request.ContainerId).Msgf("failed to update checkpoint state: %v", updateStateErr)
 		}
+
+		err = exposeNetwork()
```
jsun-m, line 223: When we are checkpointing, we do not want requests to taint the checkpoint. It should only checkpoint the ready state of the pod and not any running requests. In the case of an API server, the running request would be a connection.

pkg/worker/lifecycle.go
```diff
@@ -668,21 +680,27 @@ func (s *Worker) spawn(request *types.ContainerRequest, spec *specs.Spec, output
...
+
 	// Expose the bind ports
-	for idx, bindPort := range opts.BindPorts {
-		err = s.containerNetworkManager.ExposePort(containerId, bindPort, int(request.Ports[idx]))
-		if err != nil {
-			log.Error().Str("container_id", containerId).Msgf("failed to expose container bind port: %v", err)
-			return
-		}
-	}
+	// for idx, bindPort := range opts.BindPorts {
```
jsun-m: @luke-lombardi Would it be problem if we delay the exposing of ports right before the container starts instead?

sdk/src/beta9/abstractions/pod.py
```diff
@@ -281,6 +305,12 @@ def deploy(
         if is_custom_image:
             ignore_patterns = ["**"]
 
+        if self.checkpoint_enabled and self.checkpoint_condition is not None:
+            self.entrypoint = [
+                "(python -m beta9.runner.checkpoint &) &&",
```
jsun-m, line 310: With regards to the current snippet of entrypoint code we have, this is the best place to put the script. 

we can also move this to the backend and pass `USER_CODE_DIR`
```
["sh", "-c", f"cd {USER_CODE_DIR} && {' '.join(self.entrypoint)}"]
```

sdk/src/beta9/cli/deployment.py
```diff
@@ -180,17 +180,21 @@ def create_deployment(
...
+            if not ok:
+                terminal.error("Deployment failed ☠️")
+                return
+        except Exception as e:
+            terminal.error(f"Deployment failed ☠️: {e}")
             return
-
-        if not module and hasattr(user_obj, "cleanup_deployment_artifacts"):
-            user_obj.cleanup_deployment_artifacts()
+        finally:
```
jsun-m, line 195: GRPC errors left pod artifacts in the repo and they were not properly cleaned up.

pkg/worker/criu.go
```diff
@@ -123,31 +129,30 @@ func (s *Worker) attemptCheckpointOrRestore(ctx context.Context, request *types.
 			configPath: configPath,
 		})
 		if err != nil {
-			updateStateErr := s.updateCheckpointState(request, types.CheckpointStatusRestoreFailed)
-			if updateStateErr != nil {
-				log.Error().Str("container_id", request.ContainerId).Msgf("failed to update checkpoint state: %v", updateStateErr)
+			var e *runc.ExitError
```
jsun-m, line 132: [TODO]: This logic is not complete. @luke-lombardi I want to chat with you about stop and error conditions related to a running RestoreCheckpoint process. 

A SIGKILL from container stop event from user would put the checkpoint into a `restore_failed` state which is not what we want but this SIGKILL can also occur for other factors that are related to a failed restore. We need to hash about better indicators of failure in a restore checkpoint process.

PR #1400, opened 2025-08-28
Feat: Add restricted token type for users using api token in public clients such as frontend

jsun-m:


pkg/auth/http.go
```diff
@@ -114,6 +114,24 @@
 	}
 }
 
+// This prevents users with lower level tokens from accessing an endpoint even if they have access to the workspace.
+func WithRestrictedWorkspaceAuth(next func(ctx echo.Context) error) func(ctx echo.Context) error {
```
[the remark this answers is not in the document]
jsun-m: Duplicating the code here is fine. This was meant for a lightweight implementation of restricted auth. I want to discuss a better RBAC system for the team which will change the middleware and wrappers will work

pkg/api/v1/workspace.go
```diff
@@ -33,9 +33,9 @@ func NewWorkspaceGroup(g *echo.Group, backendRepo repository.BackendRepository,
 
 	g.POST("", group.CreateWorkspace)
 	g.GET("/current", auth.WithAuth(group.CurrentWorkspace))
-	g.GET("/:workspaceId/export", auth.WithWorkspaceAuth(group.ExportWorkspaceConfig))
-	g.POST("/:workspaceId/set-external-storage", auth.WithWorkspaceAuth(group.SetExternalWorkspaceStorage))
-	g.POST("/:workspaceId/create-storage", auth.WithWorkspaceAuth(group.CreateWorkspaceStorage))
+	g.GET("/:workspaceId/export", auth.WithRestrictedWorkspaceAuth(group.ExportWorkspaceConfig))
```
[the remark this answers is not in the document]
jsun-m: This is correct and also shows how this implementation is just a quick addition. We need to discuss custom RBAC in more depth.

pkg/auth/http.go
```diff
@@ -114,6 +118,28 @@ func WithWorkspaceAuth(next func(ctx echo.Context) error) func(ctx echo.Context)
...
+		cc, ok := ctx.(*HttpAuthContext)
+		if !ok {
+			return echo.NewHTTPError(http.StatusUnauthorized)
+		}
+
+		if cc.AuthInfo == nil {
+			return echo.NewHTTPError(http.StatusUnauthorized)
+		}
+
+		if (cc.AuthInfo.Workspace.ExternalId != workspaceId && cc.AuthInfo.Token.TokenType != types.TokenTypeClusterAdmin) || cc.AuthInfo.Token.TokenType == types.TokenTypeWorkspaceRestricted {
```
luke-lombardi: Can you clean this logic up a little

pkg/auth/grpc.go
```diff
@@ -71,7 +71,8 @@ func (ai *AuthInterceptor) validateToken(md metadata.MD) (*AuthInfo, bool) {
 		}
 	}
 
-	if !token.Active || token.DisabledByClusterAdmin {
+	// For now, restricted tokens should not be allowed to access grpc calls
+	if !token.Active || token.DisabledByClusterAdmin || token.TokenType == types.TokenTypeWorkspaceRestricted {
```
luke-lombardi: will this work with the grpc gateway proxy stuff?

pkg/auth/http.go
```diff
@@ -114,6 +118,28 @@ func WithWorkspaceAuth(next func(ctx echo.Context) error) func(ctx echo.Context)
 	}
 }
 
+// This prevents users with lower level tokens from accessing an endpoint even if they have access to the workspace.
```
luke-lombardi: Can you improve this comment

pkg/auth/grpc.go
```diff
@@ -143,3 +153,7 @@ func (ai *AuthInterceptor) Unary() grpc.UnaryServerInterceptor {
 func (ai *AuthInterceptor) newContextWithAuth(ctx context.Context, authInfo *AuthInfo) context.Context {
 	return context.WithValue(ctx, authContextKey, authInfo)
 }
+
+func CheckWorkspaceRestrictedToken(authInfo *AuthInfo) bool {
```
luke-lombardi: I would call this IsWorkspaceRestrictedToken

pkg/gateway/services/container.go
```diff
@@ -98,6 +98,12 @@ func (gws GatewayService) getContainersAsAdmin() ([]types.ContainerState, map[st
 func (gws GatewayService) CheckpointContainer(ctx context.Context, in *pb.CheckpointContainerRequest) (*pb.CheckpointContainerResponse, error) {
 	authInfo, _ := auth.AuthInfoFromContext(ctx)
 	workspaceId := authInfo.Workspace.ExternalId
+	if auth.CheckWorkspaceRestrictedToken(authInfo) {
```
luke-lombardi: The nomenclature is confusing here -- we're calling it `WithRestrictedWorkspaceAuth`, but it seems that those tokens have MORE permissions. And here, the check is saying access denied for those more permissive tokens? I feel like this is either backwards or the naming is bad

luke-lombardi (changes requested):

PR #1442, opened 2025-09-29
fix chart

luke-lombardi:
## Summary by cubic
Pin the Redis image in the Helm chart to docker.io/bitnamilegacy/redis:7.2.4-debian-11-r2 to ensure reliable deployments in the beta9 namespace and avoid upstream image changes.

### DIFF
--- a/manifests/kustomize/components/redis/helm-chart-generator.yaml
+++ b/manifests/kustomize/components/redis/helm-chart-generator.yaml
@@ -7,7 +7,6 @@ releaseName: *name
 name: redis
 repo: https://charts.bitnami.com/bitnami
 version: 18.9.1
-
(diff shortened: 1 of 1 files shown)


manifests/kustomize/components/redis/helm-chart-generator.yaml
```diff
@@ -7,6 +7,7 @@ releaseName: *name
 name: redis
 repo: https://charts.bitnami.com/bitnami
 version: 18.9.1
+
```
cooper-grc: Is there a reason for this newline?

luke-lombardi: no, removed

PR #1445, opened 2025-09-30
Improve SandboxProcessError

cooper-grc:
## Summary by cubic
Improves sandbox exec error propagation so clients get the exact error message, and simplifies SandboxProcessError text.

- **Bug Fixes**
  - Return ErrorMsg from RunCServer on exec failures and forward it through PodSandboxExec.
  - Stop wrapping SandboxProcessError with a redundant prefix; use the raw message.

### DIFF
--- a/pkg/worker/runc_server.go
+++ b/pkg/worker/runc_server.go
@@ -454,8 +454,9 @@ func (s *RunCServer) handleSandboxExec(ctx context.Context, in *pb.RunCSandboxEx
 		log.Error().Str("container_id", in.ContainerId).Msgf("failed to execute sandbox process: %v", err)
 
 		return &pb.RunCSandboxExecResponse{
-			Ok:  false,
-			Pid: -1,
+			Ok:       false,
+			Pid:      -1,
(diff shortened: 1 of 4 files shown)


pkg/worker/runc_server.go
```diff
@@ -454,8 +454,9 @@ func (s *RunCServer) handleSandboxExec(ctx context.Context, in *pb.RunCSandboxEx
 		log.Error().Str("container_id", in.ContainerId).Msgf("failed to execute sandbox process: %v", err)
 
 		return &pb.RunCSandboxExecResponse{
-			Ok:  false,
-			Pid: -1,
+			Ok:       false,
```
luke-lombardi, line 457: Curious if you confirmed what types of error messages this results in for invalid binary names / args?

cooper-grc: For invalid binary names I get: `exec: "invalid-command": executable file not found in $PATH` but invalid args don't follow the same error path. To see an error for invalid args (at least the scenarios I've tested) you have to get the logs from a process.

PR #1513, opened 2025-11-07
Add Allow List to Sandboxes

cooper-grc:
## Summary by cubic
Add outbound network controls to Sandboxes and Pods. You can block all outbound traffic or allow only specific CIDR ranges, while reply traffic to exposed ports stays allowed.

- **New Features**
  - Added block_network to stubs/sandboxes/pods (SDK, gateway/types proto, OpenAPI). Enforced via host-namespace iptables for IPv4/IPv6; reply traffic (ESTABLISHED, RELATED) is allowed.
  - Added allow_list to permit outbound only to specified CIDR ranges (validated/normalized in worker; supports IPv4 and IPv6). When set, all other outbound is blocked. Cannot be used together with block_network. Limit 10 CIDR entries; invalid CIDR values error immediately.

- **Migration**
  - No changes required. Use block_network=True to block all outbound, or allow_list=["<cidr>"] to allow specific ranges. Do not set both. Allow list must contain valid CIDRs (max 10).

<sup>Written for commit 9700ed80c0a9d3dcae6c5dbf5160328511b7988a. Summary will update automatically on new commits.</sup>


pkg/worker/network_util.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// validateCIDR validates CIDR notation and returns:
+// - normalized CIDR string
+// - boolean indicating if it's IPv6
+// - error if invalid
+func validateCIDR(entry string) (string, bool, error) {
+	_, ipNet, err := net.ParseCIDR(entry)
+	if err != nil {
+		return "", false, fmt.Errorf("not a valid CIDR notation (e.g., 8.8.8.8/32 or 10.0.0.0/8)")
```
luke-lombardi: What happens if it's not a valid CIDR? Does the container fail to spin up?

luke-lombardi (changes requested):

luke-lombardi (approved):

PR #1814, opened 2026-07-29
beam-thunder integration

rohanphadnis-thunder:
# Beam-Thunder Integration

This PR is for integrating beta9 with Thunder Compute. Key changes:

## Phase 1: `gpu_virtualized`

`gpu_virtualized`, a boolean value is propagated from the Python beam client SDK to the beam gateway to each individual worker. Note that this is still a temporary change, preserved for its usefulness in prototyping. Eventually, GPU virtualization is controlled at the level of the pool. If a pool is virtualized, all its containers would be virtualized as well.

## Phase 2: Beam Worker

The beam worker implements `beta9/pkg/worker/thunder.go`. This contains the `ContainerThunderManager`, a struct which implements the `GPUManager` interface. It mounts `nvidia-smi`, `libcuda.so`, and `libnvidia-ml.so` into the container from the host, but blocks the mounting of any physical GPU. For assigning and unassigning GPUs, it will use the gateway's thunder service and mint/revoke enrollment tokens as needed. Finally, when the container starts, the worker will run the curl installer in the container.

## Phase 3: Beam Agent

The beam agent change is very easy.
1. Tailscale is initialized on the host node. This was already the case. Thunder will simply need an IP address which is discoverable to all nodes and container clients in the pool. It will read the IP address of the `tailscale0` interface.
2. Next, the gateway will use newly added RPC call `CreateNodeEnrollment` to get a node enrollment token and run the node curl installer.
3. Once the installer is run, the agent can continue its setup.

## Phase 4: Beam Gateway

The beam gateway provides RPC calls to the beam agent and the beam worker in order to mint these enrollment tokens. It acts as a central client for the Thunder Compute API. This design ensures that the global API token isn't propagated to each beam agent/worker.

To make the Thunder information stateful, 3 new redis maps are added:
1. for mapping client id to enrollment token id. This is so that when the client is unenrolled (ie during sandbox teardown), the thunder client's enrollment can be revoked.
2. for mapping machine id to enrollment token id. This is so that if nodes ever need to be unregistered, the enrollment tokens for the nodes can be revoked.
3. for mapping beam pools to thunder zone IDs. Each pool gets its own Thunder Zone ID.

---
## Summary by cubic
Integrates Thunder virtual GPUs with pool-level control and startup safety. Previously only physical GPUs were used and containers published “started” immediately; now pools can enable virtualization, workers choose Thunder per request, and “started” waits until the Thunder installer completes.

- Pool control: adds `WorkerPoolConfig.gpuVirtualized` (surfaced in OpenAPI and Python SDK) and propagates to workers via `WORKER_GPU_VIRTUALIZED`.
- Agent flow: queries pool mode with `GetAgentPoolVirtualization`, discovers a cache-local private `thunderd` IP, installs via `CreateNodeEnrollment`, and uninstalls on shutdown (`ExecStopPost`); increases `TimeoutStopSec` to 60s and soft-fails with logs if unavailable.
- Worker behavior: selects Thunder vs physical GPUs per request, injects Thunder `LD_PRELOAD`, mounts `nvidia-smi`, `libcuda.so.1`, `libnvidia-ml.so.1`, and the Thunder library, and uses pre-init/startup hooks so “started” is withheld until install finishes; creates/revokes client enrollments on start/teardown.
- Gateway: registers a new `ThunderService` (Redis-backed) using `github.com/Thunder-Compute/thunder-sdk` for client/node enrollments and per-pool zones; exposes `CreateNodeEnrollment`/`DeleteNodeEnrollment` via agent-token auth and returns a clear error if Thunder is disabled; adds consistent Redis keys and a pool lock.
- Tooling: generates `pkg/gateway/services/thunder/thunder.proto`, adds `NewThunderServiceClient`, and updates `bin/gen_proto.sh`; updates GPU manager interfaces to accept context/request.

- Migration
  - Set `THUNDER_API_URL` and `THUNDER_API_TOKEN` in the gateway environment.
  - Ensure agents can resolve a private IP for `thunderd`; node enrollment runs automatically.
  - Enable virtualization per pool via `WorkerPoolConfig.gpuVirtualized`.
  - Expect container “started” events to delay until the Thunder install completes.

<sup>Written for commit 5ab900bda267f3ef7d77c3bf35d66d8023277069. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/beam-cloud/beta9/pull/1814?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pkg/abstractions/common/instance.go
```diff
@@ -337,6 +337,7 @@ func (i *AutoscaledInstance) CheckConcurrencyLimit() error {
 	request := &types.ContainerRequest{
 		Cpu:              i.StubConfig.Runtime.Cpu,
 		GpuCount:         uint32(gpuCount),
+		GpuVirtualized:   i.StubConfig.Runtime.GpuVirtualized,
```
luke-lombardi: Is this still required? I know we talked about removing from the client, but I feel like it should also be removed from the container request as well.

pkg/worker/lifecycle.go
```diff
@@ -269,6 +269,8 @@ func (s *Worker) markContainerStopping(containerId string) {
 }
 
 func (s *Worker) deleteContainer(containerId string) {
+	s.thunderSetupTracker.Delete(containerId)
```
luke-lombardi: feels like the thundersetuptracker calls could be pushed down into the runtime layer potentially? feels like it could get confusing having those calls up here next to the core container lifecycle pieces

pkg/worker/lifecycle.go
```diff
@@ -1452,7 +1482,7 @@ func (s *Worker) spawn(request *types.ContainerRequest, spec *specs.Spec, output
 		s.setupOOMWatcher(ctx, containerId, pid, spec, request, outputLogger, &isOOMKilled)
 	}()
 
-	exitCode, _ = s.runContainer(ctx, request, outputLogger, outputWriter, startedChan, checkpointPIDChan, opts.StartupStartedAt, opts.StartupPortBindings, opts.CheckpointFilesystemRestore)
+	exitCode, _ = s.runContainer(ctx, request, outputLogger, outputWriter, startedChan, checkpointPIDChan, thunderInstallResult, opts.StartupStartedAt, opts.StartupPortBindings, opts.CheckpointFilesystemRestore)
```
luke-lombardi: same comment as above, if theres a way to get this to be jammed into a runtime (under pkg/runtime) I think that would be cleaner

luke-lombardi (changes requested):
High level this looks good, mostly just some nit-picks around code organization

pkg/abstractions/pod/pod.go
```diff
@@ -471,6 +471,14 @@ func (s *GenericPodService) run(ctx context.Context, authInfo *auth.AuthInfo, st
 		gpuCount = 1
 	}
 
+	log.Info().
```
luke-lombardi: Is this log required? we log this in the scheduler too I think

pkg/agent/transport.go
```diff
@@ -80,8 +81,20 @@ func runTSNetRouteProxy(ctx context.Context, client pb.GatewayServiceClient, age
...
-			hostname = strings.TrimSuffix(status.Self.DNSName, ".")
+		if status, err := localClient.Status(ctx); err == nil {
+			if status.Self != nil && status.Self.DNSName != "" {
+				hostname = strings.TrimSuffix(status.Self.DNSName, ".")
+			}
+		}
+	}
+	poolVirtualized, err := requestAgentPoolGPUVirtualized(ctx, client, agentToken)
+	if err != nil {
+		fmt.Fprintf(stderr, "Thunder node enrollment skipped: %v\n", err)
```
luke-lombardi: should use the structured logger for this from zerolog I think

pkg/gateway/services/thunder/db.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	_, err = r.rdb.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
+		pipe.Set(ctx, common.RedisKeys.ThunderZone(state.WorkspaceID, state.PoolName), data, 0)
+		pipe.SAdd(ctx, common.RedisKeys.ThunderZoneIndex(state.WorkspaceID), state.PoolName)
+		return nil
+	})
+	return err
+}
+
+func (r *RedisRepository) DeleteZone(ctx context.Context, workspaceID, poolName string) error {
```
luke-lombardi: I think it may make sense to push this into the repository layer. repository/thunder_redis.go or something

pkg/runtime/startup_hook_wrapper.go
```diff
@@ -0,0 +1,171 @@
...
+import (
+	"context"
+	"fmt"
+	"syscall"
+	"time"
+
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+const startupHookShutdownTimeout = 5 * time.Second
```
luke-lombardi: This code can probably also live inside startup_hook.go maybe?

luke-lombardi (changes requested):

pkg/gateway/services/thunder/service.go
```diff
@@ -0,0 +1,446 @@
...
+			MachineID:         attrs.machineID,
+			PoolName:          attrs.poolName,
+			EnrollmentTokenID: enrollment.EnrollmentTokenID,
+		}, 0); err != nil {
+			if _, deleteErr := s.client.DeleteEnrollmentTokenNode(ctx, enrollment.EnrollmentTokenID); deleteErr != nil && !isThunderNotFound(deleteErr) {
+				return fmt.Errorf("failed to save Thunder client enrollment: %w; additionally failed to revoke Thunder enrollment token %q: %v", err, enrollment.EnrollmentTokenID, deleteErr)
+			}
+			return err
+		}
+		s.revokePreviousEnrollmentToken(ctx, previousEnrollmentTokenID, enrollment.EnrollmentTokenID, attrs.workspaceID, attrs.poolName, "container", containerID)
```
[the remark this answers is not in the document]
rohanphadnis-thunder: Enrollment tokens do not hold capacity - they are like a one-time password for acquiring a virtual GPU. Unclaimed enrollment tokens are of no consequence. An enrollment token garbage collector or reconciler is, therefore, not critical. Can be implemented in future PRs.

pkg/worker/repository.go
```diff
@@ -61,6 +61,17 @@ func NewBackendRepositoryClient(ctx context.Context, config types.AppConfig, tok
 	return pb.NewBackendRepositoryServiceClient(conn), nil
 }
 
+// NewThunderServiceClient creates a new Thunder service client.
+func NewThunderServiceClient(ctx context.Context, config types.AppConfig, token string) (pb.ThunderServiceClient, error) {
```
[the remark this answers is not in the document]
rohanphadnis-thunder: Ignoring this comment - there was already redundancy in the backend and worker setups. Minimizing the surface of thunder changes.

pkg/repository/thunder_redis.go
```diff
@@ -0,0 +1,262 @@
...
+	if state == nil || state.ContainerID == "" {
+		return errThunderStateRequired
+	}
+	data, err := json.Marshal(state)
+	if err != nil {
+		return err
+	}
+	_, err = r.rdb.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
+		pipe.Set(ctx, common.RedisKeys.ThunderClientEnrollment(state.ContainerID), data, ttl)
+		pipe.SAdd(ctx, common.RedisKeys.ThunderClientEnrollmentIndex(), state.ContainerID)
```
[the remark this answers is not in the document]
rohanphadnis-thunder: removed TTL param from methods

pkg/gateway/services/thunder/service.go
```diff
@@ -0,0 +1,446 @@
...
+		}
+		return s.repo.DeleteClientEnrollment(ctx, containerID)
+	})
+	if err != nil {
+		return &pb.DeleteClientEnrollmentResponse{ErrorMsg: err.Error()}, nil
+	}
+	return &pb.DeleteClientEnrollmentResponse{Ok: true}, nil
+}
+
+func (s *Service) CreateNodeEnrollment(ctx context.Context, req *pb.CreateNodeEnrollmentRequest) (*pb.CreateNodeEnrollmentResponse, error) {
```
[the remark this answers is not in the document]
rohanphadnis-thunder: This is a style suggestion. Ignoring this comment since the node and client handling may nuance going forward - separating them may be better. Code is modular and readable so the proposed cleanup is not extremely necessary.

pkg/gateway/services/thunder/service_test.go
```diff
@@ -0,0 +1,854 @@
...
+		GpuCount:    2,
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	var createZoneCalls int
+	var createTokenCalls int
+	var deleteTokenCalls int
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if got := r.Header.Get("Authorization"); got != "Bearer central-token" {
```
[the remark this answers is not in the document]
rohanphadnis-thunder: resolved

pkg/gateway/services/compute.go
```diff
@@ -130,6 +130,20 @@ func (gws *GatewayService) RequestAgentTransportCredential(ctx context.Context,
 	return gws.computeService.RequestAgentTransportCredential(ctx, in)
 }
 
+func (gws *GatewayService) CreateNodeEnrollment(ctx context.Context, in *pb.CreateNodeEnrollmentRequest) (*pb.CreateNodeEnrollmentResponse, error) {
```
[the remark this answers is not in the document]
rohanphadnis-thunder: resolved

pkg/gateway/gateway.proto
```diff
@@ -1884,4 +1904,5 @@ message AgentPoolRuntimeConfig {
   string durable_disks_path = 7;
   AgentPoolCacheConfig cache = 8;
   string config_group = 9;
+  bool gpu_virtualized = 10;
```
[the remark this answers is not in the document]
rohanphadnis-thunder: resolved on regeneration

pkg/agent/transport.go
rohanphadnis-thunder: done

pkg/agent/transport.go
```diff
@@ -106,6 +119,10 @@ func runTSNetRouteProxy(ctx context.Context, client pb.GatewayServiceClient, age
 	return newRouteProxy(client, agentToken, machineID, listener, proxyTarget, workers, stdout, stderr).run(ctx)
 }
 
+func shouldSetupThunderNode(gpuVirtualized bool) bool {
```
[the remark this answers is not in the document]
rohanphadnis-thunder: done

pkg/gateway/services/compute/agent.go
```diff
@@ -356,6 +364,19 @@ func (s *Service) RequestAgentTransportCredential(ctx context.Context, in *pb.Re
 	}, nil
 }
 
+func (s *Service) GetAgentPoolVirtualization(ctx context.Context, in *pb.GetAgentPoolVirtualizationRequest) (*pb.GetAgentPoolVirtualizationResponse, error) {
+	agentState, errMsg := s.requireAgentState(ctx, in.GetAgentToken())
+	if errMsg != "" {
+		return &pb.GetAgentPoolVirtualizationResponse{Ok: false, ErrMsg: errMsg}, nil
```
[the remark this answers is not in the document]
rohanphadnis-thunder: This pattern was intentional. The surface of thunder changes should be minimized.

pkg/worker/lifecycle.go
```diff
@@ -185,8 +185,8 @@ func (s *Worker) clearContainer(containerId string, request *types.ContainerRequ
 	}
 
 	// De-allocate GPU devices so they are available for new containers
-	if request.Gpu != "" {
-		s.containerGPUManager.UnassignGPUDevices(containerId)
+	if request != nil && request.RequiresGPU() {
```
luke-lombardi, line 406: Right now, is this only limited to sandboxes? 

It looks like we're adding the LD_PRELOAD everywhere, but for things besides sandboxes we're not actually installing/setting up the thunder client? Maybe we can centralize that in a clean way somewhere?

pkg/worker/thunder.go
```diff
@@ -0,0 +1,193 @@
+package worker
```
luke-lombardi, line 1: If curl fails, this might still return success /  need to make sure error propagates? Also I don't think we should require sudo here if possible cause we're running the entire installer / setup script as sudo?

luke-lombardi (changes requested):

pkg/worker/thunder.go
rohanphadnis-thunder: resolved!

pkg/worker/lifecycle.go
rohanphadnis-thunder: Fixed with a pre-init hook instead of a startup hook. The pre-init hook modifies the entry point of the container, allowing all containers to receive the installer before any process lands.

luke-lombardi (approved):

PR #1832, opened 2026-08-07
fix worker capacity and startup cancellation races

luke-lombardi:
## Summary by cubic
Fixes two races: controller worker updates no longer overwrite live scheduler capacity, and container startup/stop now honor cancellation and escalation reliably. Improves runtime stop handling with bounded timeouts, verification, and safe worker restart if the runtime is wedged.

- **Bug Fixes**
  - Preserve live capacity on existing workers using an atomic Redis script that updates controller-owned fields and recomputes free resources from “used” values.
  - Ignore STOPPING container states when checking for outstanding work during worker rollout; capacity is still accounted for by the runtime.
  - Bound runtime control calls with `stopContainerWithContext`; add retry + verify + force-delete path, and restart the worker if the runtime remains live (without releasing capacity).
  - Re-arm SIGTERM→SIGKILL escalation when a container starts after a pre-runtime stop request.
  - Propagate request context to spawn and checkpoint restore so startup cancels correctly and doesn’t proceed after cancellation.
  - Return context error when image load is canceled.
  - Correct memory free calculation on resize (test expectation updated).
  - Tests added for rollout behavior, stale-capacity protection, bounded stop, force-delete, stuck-runtime recovery, and checkpoint-restore cancellation.

<sup>Written for commit 9ea85440f7f89a17ebbcddfd4e9fc8ba24b62bd3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/beam-cloud/beta9/pull/1832?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/repository/worker_redis.go
+++ b/pkg/repository/worker_redis.go
@@ -327,6 +327,39 @@ local version = redis.call("HINCRBY", KEYS[1], "resource_version", ARGV[4])
 return {1, redis.call("HGET", KEYS[1], "machine_id") or "", free_cpu, free_memory, free_gpu, version}
 `)
 
+// updateExistingWorkerScript updates controller-owned worker fields without
+// replacing live scheduler capacity. It runs atomically with capacity updates,
+// preserving the resources already in use if the machine's totals changed.
+var updateExistingWorkerScript = redis.NewScript(`
+if redis.call("EXISTS", KEYS[1]) == 0 then
+    return 0
+end
+
+local old_total_cpu = tonumber(redis.call("HGET", KEYS[1], "total_cpu") or "0")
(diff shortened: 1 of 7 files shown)


luke-lombardi:
Closing this stale, conflicting multi-area race bundle. The production scheduler investigation is now isolated in #1901 as a current-main, single-commit change with focused provisioning/cleanup race tests. Any remaining worker lifecycle work from this PR should be reintroduced independently against current main.

PR #1868, opened 2026-09-04
more perf changes

luke-lombardi:
## Summary by cubic
Sandbox startup no longer waits on four serial worker–gateway RPCs or full image-layer materialization: `ClaimContainer` handles the pre-start exchange, and checkpoint-enabled containers read layers lazily. Builds and snapshots previously uploaded full filesystem content; supported paths now publish overlay changes, while stream and lifecycle failures are isolated instead of taking down whole batches, and workspace sync uploads only added or modified files as a delta archive the gateway merges into the previous object. A prepaid credit gate refuses new containers for workspaces out of credit and periodically stops their running ones (default `noop`), and usage events the metering endpoint refuses are queued and resent.

**Startup path**

- `ClaimContainer` collapses the pre-start exchange into one RPC; `UpdateContainerStatus` echoes the persisted status so workers can detect refused terminal transitions.
- Preformatted qcow spares are adopted by rename, remembered across worker restarts, and released under NBD pressure; `mkfs` skips the unnecessary discard.
- Complete local layers read as page-file descriptors, others copy after a 2-second grace, and ephemeral overlays mount `volatile` so `fsync` on `/` skips the EBS writeback.
- Network slots avoid netlink re-reads and prime 4 at a time at startup; cache discovery runs immediately, SDK mounts prune to existing `site-packages` paths, and teardown no longer holds the reservation lock during GPU or network cleanup.
- The keepalive reply carries `idle`; only idle workers evaluate headroom by a deterministic id-order rule, the spindown refresh is bounded and fails closed, and headroom-only workers exit after a jittered max age (default 6h) so rollouts and drains are not blocked.
- Addresses use IPv6-aware `joinHostPort`; goproc readiness follows the container context, CPU metrics read cgroup `cpu.stat` interval deltas, metric scans read newest-first and flag truncation, and cachefs reads clamp to the splice pipe size.

**Reliability and image delivery**

- Events, logs, and metrics share an SSE writer with 15-second `ping` events; lifecycle relays drop invalid events and stop retrying after a bounded budget.
- Supported layered builds run the rendered Dockerfile subset, install packages with uv (compiling bytecode) from a node-local wheel cache, pack overlay uppers into gzip OCI layers split over 512 MiB, and fall back to the existing path when needed.
- Built Dockerfile prefixes are reused as `FROM`; base layers restore from compressed blobs or decompressed cache entries, fetching missing layers concurrently.
- Build `RUN` steps see synced files at `/mnt/code`; apt timeouts, retries, and an optional mirror are bind-mounted into builds without touching the published image.
- Workspace sync uploads only added/modified files as a delta the gateway merges with server-side range copies; full uploads fall back when needed and now use concurrent multipart parts.
- JuiceFS gains cache and write-back settings; GeeseFS gains a multipart size; clip updates add aligned read-ahead, deeper prefetch, splice-safe reads, and demand-based restore pacing.
- Disk-cache writes evict before refusing, background layer warms pace to a per-mount budget, off-ring hosts are probed after a miss, raw-read admission queues briefly instead of answering busy, and a failed layer materialization is retried for waiting reads.
- Snapshot layers open upper-dir files with `O_NOFOLLOW` per component and verify dev/ino; files deleted mid-snapshot are skipped, xattrs are preserved, and hard links are counted once.
- Migration 048 adds workspace-prefixed stub and task list indexes.

<sup>Written for commit de2bea721ca52c72c9ccc58b8eed79b858f921a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/beam-cloud/beta9/pull/1868?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luke-lombardi:
## Review triage — all 105 open cubic threads checked, fixed in 43dbffcd

Every thread was read against the code; valid ones are fixed (with tests where behavioural), the rest are recorded below so nothing is silently dropped. `go test ./...`, SDK tests, `verify_proto` and CI are green. The commit was rolled to staging (`perf-startup-43dbffcd`: worker, gateway, cache-server) and the workload harness re-run: 10x/25x plain and torch sandboxes 100% ok, cold start p50 ~200 ms, ML image build 17.9 s, typical image build 10.3 s, snapshot restore 410 ms.

### Fixed (highlights, P1s first)
- **image_layers.go** — snapshot no longer follows a symlink swapped in after the walk (per-component `openat(O_NOFOLLOW)` + dev/ino check against the walk); deleted files are skipped instead of aborting the layer; xattrs kept as PAX records; 16-layer cap enforced.
- **network.go** — pooled slots re-validate the netns mount and host veth (sysfs, no rtnl) before assignment.
- **worker_repo.go ClaimContainer** — returns the persisted state after the lease renewal, so a STOPPING that races the claim is no longer reported as PENDING.
- **image_snapshot.go** — `name.Insecure` only for the build registry, never for external bases; stale `v2ImageRefs` entry removed when the push fails.
- **build.go** — prefix builds no longer carry the base image's credentials; uv targets the resolved interpreter (venv-aware) and falls back to pip for flags uv rejects.
- **disk/manager.go** — `.spare-` keys rejected from `Attach`; plus the 7 other disk findings (relative-root symlink, atomic spare-sizes, `Close` joins in-flight builds, leaked run dirs, truthful `reclaimSpare`, single NBD validation, leftover runtime dir).
- **Headroom** (pool_headroom.go / worker_repo.go / worker.go) — deterministic id-order rule so two idle workers can't both conclude the other covers the minimum; lookups fail closed; only *idle* keepalives scan the pool (new `idle` field on the request); the spindown refresh is bounded and refuses to exit on error.
- **Gateway/object** — part-count overflow guard, abort on failed completion, delta staging key per base (concurrent syncs to the same manifest hash no longer clobber each other), detached-ctx abort, removal-only deltas spliced.
- **SDK sync** — ctime in the manifest cache, timeout + retry on part PUTs, malformed cache entries are a miss, recorded size is the archive size.
- Everything else in the thread list (events relay budget, gpuSamples, migration 048 invalid index, zero expiry, nested-container credentials, CPU metrics via cgroup `cpu.stat`, SDK mount pruning, `mkdirBindSource`, address-map join, layered-build parser fixes, apt symlink refusal, base-cache non-sha256 digests, progress reporting, …).

### Checked and not changed (reason)
- `build.go:385` `--compile-bytecode` — it *is* the canonical long form (`--compile` is the alias) and every staging build today used it.
- `build_prefix.go:39` secrets in the prefix key — `getImageID` already ignores secret values, so the prefix cache is no looser than the image cache it feeds.
- `cachefs.go:129` `max_read` — go-fuse appends its own `max_read=<MaxWrite>` after user options and negotiates `max_pages` in INIT; reads are capped either way.
- `client.go:1797` off-ring probe — a miss is an index lookup with no payload, one RTT, same as `HasContent`; the alternative is a registry fetch.
- `client.go:3049` `WaitForHosts` polling — once per worker at startup, ≤30 s.
- `overlay.go:372` `volatile` flag cached process-wide — only set when the option-less remount immediately succeeds; layer-specific EINVALs fail both mounts and are never cached.
- `sandbox.go:483` unbounded goproc wait — runs under the spawn ctx (container lifetime), and only for sandbox stubs; builds never reach it.
- `disk/spare.go:92` `RemoveAll(keyDir)` — the directory only ever holds `state.json` + `layers/`, which the slow path already removed on this input.
- `dockerfile_plan.go:105` COPY paths with spaces — Docker's shell form is whitespace-delimited too; JSON form already falls back to bud.
- `image_base_cache.go:79` eager config fetch — the config is needed on every path (diffIDs for the tar cache, layout write).
- `object_delta_splice.go:129` sort — archives are written by the SDK in sorted order, so the sort is a no-op on real inputs.
- `object_delta_splice_test.go:205` — the matrix runs in 0.03 s.
- `image_build_apt.go:21` ports.ubuntu.com — left untouched on purpose: `ubuntu-ports` is a separate tree not carried by archive mirrors; rewriting would 404 arm64 builds. A `PortsMirror` option can follow if we want ports accelerated.
- `sync.py:340` abort multipart on part-PUT failure — needs an abort RPC; the completion-failure half is handled server-side, orphaned uploads are covered by the bucket lifecycle rule.
- Style-only threads (dedup of `sse.go`, `object_delta.go` validation, `image_build_apt.go`, `worker_test.go` fakes, `raw_transport_test.go` literals, `test_filesyncer.py` list, `ENV/LABEL` parsing) — applied where trivial, otherwise left.

mernit (approved):

luke-lombardi:
## Follow-ups since the triage comment

**Run 6a import failures (5/25 `import torch`, truncated stderr) — root cause and fix (`204b635f`)**

Traced to clip: `ensureLayerCached` runs one materialization per layer under a singleflight, and when that single stream failed (a reset mid-download of a 1.5 GiB layer), every process waiting on that layer got `EIO` from the FUSE read with no retry — Python then fails `_load_unlocked` on whatever module source/`.so` it was reading. That is exactly the fan-out shape of run 6a (5 sandboxes, one worker, same instant). The path pre-dates this branch; the unpaced restores in the earlier pacer made the failure more likely to hit. clip `5e80513` now retries a failed materialization with backoff (4 attempts) for the owner and every waiter, and logs the underlying error when a read has to return `EIO` instead of leaving only the process-side traceback.

Re-tested on staging with the new worker image: 5 rounds × 30 concurrent first-use `import torch`, 150/150 OK. One round was the run-6a shape exactly — a 6-minute-old worker that had never seen a torch layer, no warm-up, 30 cold readers on a fresh layer: 30/30 in 11 s wall, no retries fired, no EIO.

**Workers living forever (`de2bea72`)**

Two staging workers sat idle for 8 h. They were each their pool's headroom holder (lowest-id ready worker with `poolSizing.minFree*` set), which the `pool_headroom` keepalive flag from `3088f025` keeps alive by design, with no upper bound. Hand-off works (both idled out within 5 min of a lower-id replacement taking the headroom), but the holder itself never rotated. New `worker.headroomWorkerMaxAge` (default 6 h, id-jittered by up to 1/8, `0` disables): an idle worker whose only reason to stay is headroom exits past that age and the sizer replaces it; busy workers are never aged out. Covered by `TestShouldExitIdle` / `TestHeadroomWorkerMaxAge`.

PR #1892, opened 2026-09-11
worker: fail a lazy OCI mount whose registry refuses the image

luke-lombardi:
A lazy mount never talks to the registry until the runtime's first read
of a layer the node does not hold, so a dead credential surfaced as EIO
from FUSE, which nvidia-container-cli's ldconfig turned into SIGBUS, one
container after another. Now, when any layer must come from the registry,
the worker HEADs the manifest with the vended credentials before mounting
and fails the container start on 401/403/404 with the registry's answer,
also written to the container log. Layers already in the layer cache or a
local page store need no registry; other registry errors stay lazy.

Verified: `go test ./pkg/worker/` passes; `remote.Head` against `ghcr.io/beam-cloud/vllm-harness` with an expired token yields the same `403 DENIED` transport error this catches (anonymous: `401 UNAUTHORIZED`).

---
## Summary by cubic
Fails a lazy OCI mount when the registry refuses the image, so a dead credential now fails the container start with the registry's answer instead of surfacing as EIO from FUSE, which nvidia-container-cli's ldconfig turned into SIGBUS.

**Details**
- When any layer must come from the registry, the worker HEADs the manifest with vended credentials before mounting.
- 401/403/404 responses fail the container start; the registry's answer is also written to the container log.
- Layers already in the layer cache or a local page store skip the registry check; other registry errors stay lazy.

<sup>Written for commit dfd0ef58dd0f9e8d336f94dca67006980cd88f98. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/beam-cloud/beta9/pull/1892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/worker/image.go
+++ b/pkg/worker/image.go
@@ -33,6 +33,10 @@ import (
 	"github.com/beam-cloud/clip/pkg/clip"
 	clipCommon "github.com/beam-cloud/clip/pkg/common"
 	clipStorage "github.com/beam-cloud/clip/pkg/storage"
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/hanwen/go-fuse/v2/fuse"
 	"github.com/opencontainers/umoci"
 	"github.com/opencontainers/umoci/oci/cas/dir"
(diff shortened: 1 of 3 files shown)


mernit (approved):

luke-lombardi:
Folded into #1894 (commit fd0da558) so the credential model and its runtime failure mode ship together.

PR #1893, opened 2026-09-11
image: resolve registry credentials from workspace secrets

luke-lombardi:
Image.from_registry(credentials=[...]) names environment variables, so a
deploy has to carry the registry credential itself, and a CI deploy hands
over whatever it has: for hosted-endpoints that was the GitHub Actions job
token, stored as the image's runtime credential and dead an hour later
(every replica of an unmodified image pulls with it).

A name the SDK cannot find in the environment is now sent empty and the
gateway fills it from the workspace secret of that name, so a long-lived
credential lives once, as a workspace secret, next to the token that
deploys. The gateway also refreshes the stored registry secret when the
image already exists, so a rotated credential reaches running deploys
instead of the first deploy's value persisting for the image's lifetime.

Tests: `go test ./pkg/abstractions/image/`, `pytest sdk/tests/test_image.py sdk/tests/test_managed_endpoint.py`.

---
## Summary by cubic
Resolves registry credentials from workspace secrets so a long-lived credential can be stored once and rotated credentials reach running deploys.

**Changes**
- When a credential name is not set in the environment, the SDK sends an empty value and the gateway fills it from the workspace secret of the same name.
- The gateway refreshes the stored registry secret when the image already exists, so a rotated credential is used by running deploys.
- The SDK no longer raises `ImageCredentialValueNotFound` for missing environment variables; missing keys are left empty for the gateway to resolve.

<sup>Written for commit 360d089d4747b8678bff3130e946e17bc0366e4a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/beam-cloud/beta9/pull/1893?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/abstractions/image/image.go
+++ b/pkg/abstractions/image/image.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"maps"
 	"sort"
 
 	abstractions "github.com/beam-cloud/beta9/pkg/abstractions/common"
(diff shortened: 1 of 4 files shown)


luke-lombardi:
Superseded by a simpler model: images name the workspace secrets holding their registry credentials; no snapshot secret, no env/workspace resolution. See the follow-up PR.

PR #1895, opened 2026-09-12
volume: prevent S3 path traversal outside volume root

jhapriyavart10:
## Problem

`VolumePath` is client-controlled input used to construct S3 object keys for volume operations. The previous implementation used `filepath.Clean` / `filepath.Join` without enforcing that the resulting key remained inside the authenticated volume's root prefix.

A path containing parent-directory traversal could therefore resolve outside the intended volume boundary. Because the resulting key is passed to S3 presigning and multipart operations, this can enable access to other object keys when the service's configured S3 credentials permit that access.

## Root Cause

`joinCleanPath` used OS filesystem path semantics for S3 object keys and performed no volume-boundary validation after resolving `..` components.

For example, with:

`volumes/ws-A/vol-A`

and:

`../../ws-B/vol-B/compromise.txt`

the existing implementation could construct:

`volumes/ws-B/vol-B/compromise.txt`

## Fix

- Replace filesystem-specific path handling with POSIX `path` semantics for S3 object keys.
- Centralize key construction and enforce strict containment:
  `key == rootPrefix || strings.HasPrefix(key, rootPrefix + "/")`
- Reject out-of-boundary paths before any S3 or presigning operation.
- Apply the validation consistently to `CreatePresignedURL`, `CreateMultipartUpload`, `CompleteMultipartUpload`, and `AbortMultipartUpload`.
- Preserve legitimate existing path normalization behavior.

## Tests

Added regression coverage for:
- valid paths;
- internal non-escaping `..`;
- leading/redundant separators;
- same-volume traversal;
- cross-workspace traversal;
- bucket-root escape;
- prefix-collision cases;
- shared and dedicated storage roots;
- service-level rejection before S3 presigning/upload operations.

Validation performed locally:

- `gofmt -d ...` — PASS
- `go test -v ./pkg/abstractions/volume/...` — PASS
- `go test -race ./pkg/abstractions/volume/...` — PASS
- `go vet ./pkg/abstractions/volume/...` — PASS

`go test ./pkg/...` was not run locally because full-repository compilation caused excessive CPU/RAM pressure in the WSL environment.

## Compatibility

Legitimate existing paths retain their previous normalized behavior. Inputs that attempt to escape the authenticated volume boundary are now rejected.

## Related Work

PR #1548 addresses the same general S3 object-key construction area but uses raw string concatenation and does not enforce volume-root containment. This PR takes a different approach by enforcing the actual volume-boundary invariant and adding regression coverage for traversal.

---
## Summary by cubic
Prevents S3 path traversal by enforcing that volume paths stay inside the volume root. Previously, `..` sequences could resolve to object keys outside the authenticated volume, enabling access to other objects when the service's S3 credentials permit it.

**Key changes**
- Replace `filepath` with POSIX `path` for S3 key building.
- Enforce containment: key must equal root prefix or start with root prefix + "/".
- Apply validation to presigning and all multipart upload methods.
- Add regression tests covering valid paths, traversal attempts, and prefix collisions.

Legitimate paths keep their normalized behavior; only escapes from the volume boundary are rejected.

<sup>Written for commit bf2b7748e5df6d27e53ebcaf0fdfd0d7325007f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/beam-cloud/beta9/pull/1895?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/abstractions/volume/multipart.go
+++ b/pkg/abstractions/volume/multipart.go
@@ -5,7 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"path/filepath"
+	"path"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -42,12 +43,13 @@ func (s *GlobalVolumeService) getS3Client() *s3.Client {
 	})
 }
 
-func joinCleanPath(parts ...string) string {
-	for i, part := range parts {
-		parts[i] = filepath.Clean(part)
+func buildVolumeObjectKey(rootPrefix, volumePath string) (string, error) {
+	key := path.Join(rootPrefix, volumePath)
+	if key != rootPrefix && !strings.HasPrefix(key, rootPrefix+"/") {
+		return "", errors.New("invalid volume path: outside volume root")
 	}
 
-	return filepath.Join(parts...)
+	return key, nil
 }
 
 func (s *GlobalVolumeService) GetFileServiceInfo(ctx context.Context, in *pb.GetFileServiceInfoRequest) (*pb.GetFileServiceInfoResponse, error) {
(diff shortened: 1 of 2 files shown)


luke-lombardi:
@jhapriyavart10 closed as dupe. this was pulled into another - thanks for the contribution!

jhapriyavart10:
Thanks for the clarification! Glad the investigation and fix were useful.

PR #1902, opened 2026-09-15
worker: never delete a stub-code cache another pod published

luke-lombardi:
## What broke

A `.map` burst on 2026-09-15 19:04Z failed 12 containers on `worker-default-ovh-7922b36e` with

```
failed to setup container mounts: lstat /var/lib/beta9/cache/stub-code/7f1f400c…/<file>: no such file or directory
runc run failed: … failed to fulfil mount request: open /data/objects/f714b7/54bcaca7-…: no such file or directory
```

`worker-default-ovh-1d91dba9` on the same node (`ns1024324`) was extracting the same object at the same moment. The stub-code cache root is per node, the singleflight is per process, so both pods were leaders and the second's `os.RemoveAll(cachePath)` deleted the directory the first's containers were copying from.

The `/data/objects` error is a symptom: `prepareContainerMounts` logged the setup failure and continued, so runc bound the unresolved object path.

## Fix

- `publishStubCodeCache`: rename into place; if a ready cache appeared meanwhile, adopt it and discard our extraction. Only a torn directory (no marker, no readers) is replaced.
- `setupUserCodeMount`: if copying from the cache fails for any reason, extract the object straight into the workspace. The cache is a shortcut; a container never fails because of it.
- `prepareContainerMounts` returns the setup error (now only a genuine object download/extract failure), so the container fails with the cause.
- `SetupContainerMounts` no longer aborts on a bucket that cannot be mounted (it was already tolerated downstream); later mounts still get their workspace-storage paths.

## Validation

`go test ./pkg/worker/` passes. New tests: `TestPublishStubCodeCacheAdoptsACacheAnotherPodPublished`, `TestSetupContainerMountsRecoversFromAnUnusableStubCodeCache`, `TestSetupContainerMountsSkipsFailedBucketAndResolvesLaterMounts`.

Found via VictoriaLogs; not related to #1897 (no volume validation rejections in 24h of gateway logs).

---
## Summary by cubic
Fixes a race where two worker pods on the same node could delete a stub-code cache another pod's containers were still copying from, causing container startup failures.

**Bug Fixes**
- Publishes a finished extraction by renaming it into place and adopts an already-ready cache if one appears; only a torn directory without the `.beta9-cache-ready` marker is replaced.
- Falls back to extracting the object straight into the workspace when the cache can't be copied, so a torn cache never fails a container.
- Returns mount setup errors so the container fails with the real cause instead of logging and continuing into a misleading runc bind error.
- Continues resolving later mount paths when a bucket can't be mounted; the skipped bucket remains left out.

**Validation**
- Adds tests for adopting a concurrently published cache, recovering from an unusable cache, and skipping failed buckets while resolving later mounts.

<sup>Written for commit 6012c20637c0d050afa6cc4f2fb8a97613c46285. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/beam-cloud/beta9/pull/1902?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luke-lombardi:
Folded into #1901.

PR #1916, opened 2026-09-21
Restore pre-cgroupfs gVisor checkpoints; load signing key for image credentials

luke-lombardi:
Two worker-facing fixes.

## 1. gVisor checkpoints taken before the cgroupfs mount fail to restore

#1910 added a `/sys/fs/cgroup` mount to `base_runsc_config.json`. runsc validates that the restore spec's mounts match the checkpointed spec (`runsc/specutils/restore.go`, `validateMounts`: destination set, type and options must match), so every gVisor checkpoint taken before the 0.1.759 deploy now fails to restore:

```
"Mounts" does not match across checkpoint restore for container: "__no_name_0"
  checkpoint: 27 mounts, no /sys/fs/cgroup
  restore:    ... + {Destination:"/sys/fs/cgroup", Type:"cgroup"}
```

The worker then falls back to a cold start and re-checkpoints. Confirmed on `38.65.239.14` (`skynet-4090-gvisor`, 124 checkpoints on the node): the saved spec in checkpoint `c69186af` (Sep 20) has no `/sys/fs/cgroup`; the one in `138928bc` (Sep 21 14:44, created by the fallback) has it. The only mount difference between the two lists is the cgroup mount.

**Fix.** Before `runsc restore`, `Runsc.Restore` reads the specs runsc saved in `checkpoint.img` (8-byte `gVisorSF` magic, 8-byte big-endian metadata length, JSON string map with a `container_specs` entry) and makes the bundle's `/sys/fs/cgroup` mount match the checkpoint: dropped when the checkpoint predates it, copied in when the checkpoint has it. gVisor's cgroupfs is virtual and the restored sandbox gets the checkpoint's filesystem state either way, so old checkpoints restore exactly as they did before #1910 and new ones keep the mount. No other mount is touched, so real mismatches (changed volumes, etc.) still fail and fall back as before. If the metadata can't be read, the bundle is left alone and runsc reports the definitive error.

The worker's existing fallback (`prepareRestoreFallback`) rewrites the original config before a cold start, so a container that fails restore still starts with the full base spec.

Left alone: #1910's mount and `exposeSandboxMemoryLimit`. A sandbox restored from a pre-#1910 checkpoint has no cgroupfs, so the memory-limit write logs its existing "no cgroup memory limit file" error and the workload sees what it saw before; that resolves when the checkpoint is refreshed.

## 2. Image credential decryption dereferenced a nil signing key

The scheduler passes `GetImageCredentials` the workspace copied onto a container request, which carries no `SigningKey`, so decrypting registry secrets dereferenced nil. `GetImageCredentials` now loads the key by external id when it is missing (on a copy, without mutating the caller's workspace), and `GetSecretsByNameDecrypted` returns an error for a workspace that still has none instead of panicking.

## Validation

- `go test ./pkg/runtime ./pkg/worker ./pkg/repository`
- Runtime: drop / add / no-op / non-cgroup mounts untouched / unreadable image leaves bundle alone; `Restore` end-to-end with a fake runsc asserting the aligned bundle is what the restore command sees. Ran `alignRestoreSpecCgroupMount` against the first 64 KiB of both production `checkpoint.img` files above with the current 54-mount gVisor base spec: Sep 20 checkpoint → mount dropped (53 mounts); Sep 21 checkpoint → unchanged.
- Repository: sqlmock tests for the signing-key load path (credentials decrypt, caller's copy stays keyless) and the keyless rejection.


luke-lombardi:
## Live runsc validation on a production gVisor node

Ran an isolated checkpoint/restore round trip on `38.65.239.14` (`skynet-4090-gvisor`) with the node's own runsc (`3d5aea8eefb1`, the build prod runs), its own `--root`, `--network none`, a CPU-only busybox sandbox, and the real `alignRestoreSpecCgroupMount` cross-compiled from this branch. No production containers were touched.

Each direction: run a counter loop, checkpoint, attempt restore with the mismatched spec, align the bundle, restore again with `restore --background --direct` + state polling + `wait --restore` (what `Runsc.Restore` does), then check the counter and `/sys/fs/cgroup`.

| Checkpoint spec | Restore spec | Unaligned restore | Aligned restore |
|---|---|---|---|
| no cgroup mount (pre-#1910 checkpoint) | cgroup mount (0.1.760) | `"Mounts" does not match across checkpoint restore` | `wait --restore` exit 0, state `running`, counter 8 → 14 (resumed), no cgroupfs |
| cgroup mount (0.1.760 checkpoint) | no cgroup mount (0.1.759 rollback, current prod) | `"Mounts" does not match across checkpoint restore` | exit 0, `running`, counter 8 → 15, `/sys/fs/cgroup/{cpu,cpuacct,cpuset,memory,...}` present |
| cgroup mount | cgroup mount | n/a | alignment is a no-op, exit 0, `running`, counter 8 → 15 |

Context from the node: `0.1.760` (built 14:25 UTC) ran ~14:25–14:51 and failed to restore the Sep 20 checkpoint `c69186af` (row 1); after the rollback to `0.1.759` at 14:51, the worker failed to restore `138928bc`, which `0.1.760` had created with the mount (row 2). Saved-spec diffs between `c69186af`, `138928bc` and `0bf26424` (created by the `0.1.759` fallback) confirm `/sys/fs/cgroup` is the only mount that differs in either direction.

Still not exercised: a full worker E2E (GPU checkpoint through the worker's restore path). The runtime code around the align step is unchanged, so a staging or single-node canary before promoting is the remaining check.

luke-lombardi:
## Restore of the actual failing production checkpoint (`c69186af`) with the aligned spec

Ran inside the production worker container on `38.65.239.14` (`docker exec` into `beam-agent-0ef65f4a`, worker `0.1.759` untouched), so the same runsc binary, same mount namespace, same nvproxy setup, but with a private `runsc --root`, a private netns with no host connectivity, an idle GPU (7), and scratch dirs in place of the customer's volume/output mounts. The checkpoint files were copied, not used in place. No production container was touched.

Bundle: the spec runsc saved in `checkpoint.img` (78 mounts, nvproxy annotation, 68 env vars) with the `/sys/fs/cgroup` mount inserted where `0.1.760`'s base config puts it, `root.path` on a fresh overlay over the stub's still-mounted image `c722401d56ca9f71` with the checkpoint's `filesystem/` as upper (same overlay options the worker uses), and the stub's cached code at `/mnt/code`.

| Step | Result |
|---|---|
| `runsc --nvproxy=true restore --background --direct` with the cgroup mount present | exit 128, `"Mounts" does not match across checkpoint restore for container: "__no_name_0"` — the production error, reproduced against the real checkpoint |
| `alignRestoreSpecCgroupMount` (real function, cross-compiled from this branch) | removes exactly the `/sys/fs/cgroup` mount; rest of the spec byte-identical (78 mounts) |
| same restore command with the aligned bundle | state `running`, `wait --restore` exit 0 in ~1 s |
| `ps` inside | `python3.12 -m beta9.runner.endpoint` PIDs 1 and 2 with ELAPSED `1-01:54` — the checkpointed process resumed, not a fresh start |
| `/proc/mounts` inside | every gofer attachment present as 9p: `/`, `/criu`, `/models`, `/outputs`, `/mnt/code`, `/tmp/.beta9`, `/etc/hosts`, `/etc/resolv.conf`, `/volumes/*`, SDK and binary bind mounts, `cuda-checkpoint`; no cgroup mount |
| nvproxy | `nvidia-smi` inside the sandbox lists the remapped GPU (`GPU-6023d4e1…`) |
| TCP 8001 inside | open (gunicorn master's restored listening socket) |
| 26 s later | still `running`; gunicorn had replaced the checkpointed worker and the new one loaded the real handler from `/mnt/code` and ran the app's `OnStart` loader |

Two harness-only artifacts hit along the way, documented so nobody repeats them: the saved spec already carries runsc's own `overlayfs_stale_read` option (runsc adds it at create, not beta9), so a bundle rebuilt from the saved spec must strip it or runsc doubles it and validation fails; and `ip netns add` leaves `lo` down, which the worker's slots do not.

Side observation, unrelated to this PR: after restore, gunicorn's master killed the checkpointed worker with `WORKER TIMEOUT` (stale heartbeat across the wall-clock jump) and booted a fresh one, which re-ran the app loader. If that also happens in production restores, the checkpoint saves the interpreter and imports but not the worker's warm state. Worth a separate look.

